PR #1, opened 2014-09-07
Add task and fix formatting.

nitrocaster:
It's much easier to read such docs when all lines fit in the textbox.


ghost:
Thanks for the suggestion. I'm new to github entirely.

PR #12, opened 2014-09-30
Update Proposed engine changes.txt

LoNer1:
Added some stuff. Please look into the viablility of the changes i proposed and if they're doable on the long run by the people we have within the project.


ghost:
Radiosity may not be the way to go, but improving the GI system in general sounds like a good idea. Crysis 3 has a pretty good GI system in place, I believe they use Light Propagation Volumes: http://www.crytek.com/cryengine/presentations/real-time-diffuse-global-illumination-in-cryengine-3

PR #22, opened 2015-04-23
Various fixes in xrD3D9-Null

ghost:
_First, I just wanted to format the code according to the repo coding conventions but I was... kinda 
discouraged by the amount of work needed._
So for now, I just fixed a few things in this module.The commits messages are pretty self-explanatory.

Build passed without any warns/errors on both debug and release configs. I didn't took the time to test it but I didn't make any breaking changes so it should work without any problems.


nitrocaster:
Looks good. Thanks.

Issue #24, opened 2015-06-18
Fix bug with more load on process than it should be


ViHtarb:
Plz fix a bug in xrRender/ParticleEffects.cpp in line 546 and xrParticles/particle_actions_collection.cpp in line 1779:
change from if (p_cnt < nWorkers \* 20) on if (p_cnt < nWorkers \* 64)

Xottab-DUTY:
The original code was
```
if (p_cnt < (nWorkers * 64))
    nWorkers = 1;
```
Currently, with that fix, code looks like this:
```
if (p_cnt < (nWorkers * 20))
    nWorkers = 1;
```
I found in r8384 (leaked COP svn revision ~March 2012) this code:
```
if (p_cnt < nWorkers)
   nWorkers = p_cnt;
```
Which code is better and what does it do?

ViHtarb:
@Xottab-DUTY
Hi. I don\`t know but on my old computer with Core 2 Duo E7300 code "if (p_cnt < (nWorkers * 20))" provides 100% load and particles bugs(it`s look like rainbows instead anomalies) instead "if (p_cnt < (nWorkers * 64))"

Code "if (p_cnt < (nWorkers * 64))" i found in sources of release engine version of COP

Code "if (p_cnt < nWorkers) = p_cnt;" i did not see and don`t tried to use

ViHtarb:
Now i noticed that this change is not in the dev and another branches of project. But why?

Xottab-DUTY:
@nitrocaster fixed this.. Maybe this commit was occasionally dropped.
So, we need to find, what solution is better
```
if (p_cnt < (nWorkers * 64))
    nWorkers = 1;
```
or
```
if (p_cnt < nWorkers)
   nWorkers = p_cnt;
```

revolucas:
Uneven distribution of workload on cores is not and never will be, a bad thing. It is actually a good thing.

revolucas:
CPU usage is 90% on my 6-core with:

if (p_cnt < nWorkers)
   nWorkers = p_cnt;

-------------------------
if (p_cnt < (nWorkers * 64))
    nWorkers = 1;

Is how it is in Shadow of Chernobyl and Clear Sky source and does seem to run better then * 20. Only 20% CPU usage.

I downloaded CoP revision 8384. It seems to be much older then whatever Nitro used as the base for OpenXray.

Xottab-DUTY:
Oh, thank you!
[...]

PR #35, opened 2015-11-03
Assigned tasks for Swartz27 in task_list.txt

ghost:


nitrocaster:
Some minor suggestions:
1] Add some brief description to task #24
2] Regarding pre-compiled shaders issue - it should not be completely disabled because in that case shader compilation takes noticeable amount of time and will occur every time user loads the game

ghost:
1) Will do
2) This only occurs for the very first time a shader has been changed/added. Once the shaders_cache is built when the game starts it never needs to be built again and will always load the game instantly. I've been using this for almost a year so I know this to be the case, it does not affect load times unless a new or changed shader has since been added.

nitrocaster:
Sure. I mean, if you disable shader cache, the game will compile shaders every time, and we don't want this to happen.

ghost:
I don't think I'm explaining this very well, let me try again (as I've been using this system for a long time now):

Step 1: A new or modified shader is added to the project
Step 2: The user deletes their shaders_cache folder
Step 3: The initial time, and the initial time only, the shaders_cache is re-built
Step 4: On all subsequent playthroughs the shaders_cache does not get re-built. It utilizes the shaders_cache that was built that initial time from now on. It does not compile shaders every time, it only compiles shaders that one single time.

I hope that explains it better.

Without doing this new or modified shaders cannot be added to the game. I am not the only modder that has done this.

nitrocaster:
Exactly. That's how it should work, except for one thing: step 2 must be detected by the engine automatically so user wouldn't care about deleting shader cache each time some shader changes.

ghost:
That I can completely agree with you on. I'm not quite sure how to go about that but I'll see if I can figure it out :)

PR #76, opened 2015-11-11
Close sleep dialog by ESC key – Done

Xottab-DUTY:


nitrocaster:
We don't use text task list anymore, only issues here on GitHub.

Xottab-DUTY:
Ok, no problem :)

PR #77, opened 2015-11-13
Fixed include and dependency

Xottab-DUTY:
Fixed include (wrong path to xrCore in xrD3D9-Null) and dependency (You need build xrLCUtil before building xrLC_Light)


src/xrD3D9-Null/xrD3D9-Null.h
```diff
@@ -3,7 +3,7 @@
 #include <d3d9caps.h>
 #include "IDirect3D9.h"
 
-#include "xrCore/Platform.h"
+#include "../xrCore/Platform.h"
```
Kaffeine: Don't we have "src" in include path list?
I think we should use "module/path_in_the_module/header_file" scheme instead of relative paths.
I'd like to see unified includes instead of things like "../../../xrCore/something.h", which're really error-prone and would break on move.

nitrocaster: xray-16/src is one of default include search paths.
upd: It should be, but xrD3D9-Null doesn't define it. Project settings need to be adjusted.

Xottab-DUTY:
Oops...

Issue #78, opened 2015-11-14
Remove ".dll" suffix from LoadModule calls.


Kaffeine:
Remove ".dll" suffix from LoadModule calls.

The final goal is to make LoadModule calls platform-independent (we have ".dll" on Windows and ".so" on Linux/Mac)

https://github.com/OpenXRay/xray-16/blob/ai_cleanup/src/xrEngine/EngineAPI.cpp#L135 and so on.

Note: we need to verify that the game works with the changes.

nitrocaster:
> we need to verify that the game works with the changes.

It does work.

Kaffeine:
> It does work.

Well, I meant that after ".dll" suffix removal we need to verify that the game works. You confirmed my suspicious at

> https://github.com/Xottab-DUTY/xray-16/commit/e63a186d02586ed03c1c986e3cb3baad72c20b33#commitcomment-14401586
> 
> These names are used by hxGrid library. Need to check if it accepts file names without extensions (I suppose it doesn't).

Kaffeine:
Thanks.

Closed by commit 9b789e69890afee2944bf6b58a2173fe340b3aff.

PR #79, opened 2015-11-14
Various fixes

Xottab-DUTY:


CrossVR:
Is there any reason to keep the Mixed configuration? I think it's better to get rid of it.

Xottab-DUTY:
Almost all suffixes removed. #78

I didn't remove only windows-specific ".dll", such as "d3d9.dll" and other

src/xrCore/xrCore.h
```diff
@@ -102,7 +102,7 @@
 #endif
 
 #include <time.h>
-#define MODULE_NAME "xrCore.dll"
+#define MODULE_NAME "xrCore"
```
Kaffeine: @nitrocaster: is this safe?

nitrocaster: @Kaffeine It is. You could just read the [specification on MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684175%28v=vs.85%29.aspx).

Kaffeine: I already read about LoadModule(), as well as GetModuleHandle() and some other functions, but I dont know all MODULE_NAME usages. May be it's not just those methods argument. I have no IDE, nor the source code ATM on this pc.

nitrocaster: The only place it's used is xrCore.cpp:49:

``` cpp
GetModuleFileName(GetModuleHandle(MODULE_NAME), fn, sizeof(fn));
```

So actually `MODULE_NAME` should be either defined in xrCore.cpp or inlined on the line above.

Kaffeine:
@nitrocaster 

> Need to check if it (hxGrid library) accepts file names without extensions

Would you test it? Or write a small instruction (for me or @Xottab-DUTY  or someone else) how to ensure that this works/doesn't work?
It seems to be used in Light module. Probably it would be enough to test light with the most capable Render (R4) in, say, local multiplayer game (just like in Armadas videos)?

Xottab-DUTY:
I'll split this PR

PR #88, opened 2015-11-20
Move all dependencies to the Externals folder.

CrossVR:
This PR renames the 3rd party folder to Externals and assumes all external dependencies are placed there.

This is in preparation of submoduling all dependencies and building them within the same solution. This will significantly reduce the difficulty of building the solution initially.


CrossVR:
Forgot to commit the libraries to the Externals folder, one moment

PR #91, opened 2015-11-24
Configuration fixes

Xottab-DUTY:


CrossVR:
Since no one maintains the Mixed configuration, I'll just blind merge that.

For the dependencies, there is indeed a lack of those dependencies. Though we should do a more thorough evaluation of that sometime.

LGTM :+1:

PR #103, opened 2015-12-12
Add early version of OpenGL renderer.

CrossVR:
The new OpenGL renderer compiles and so it should be added to the code base now to prevent conflicts.

Currently development on the OpenGL renderer is stalled until I can work around a problem with the NSight debugger.


Kaffeine:
:+1:

alexgdi:
nice

CrossVR:
Remember, this is just a very early version, it's completely bugged and almost none of the shaders have been ported yet. So don't expect to actually get in-game.

However, once I get the debugger working I should be able to make fairly quick progress since I have the 1.5 versions of the shaders to use as inspiration.

PR #114, opened 2016-01-24
Fix blurred text on r4

fedan:
Fixed


nitrocaster:
It actually works. Thank you!

PR #119, opened 2016-01-29
changes/fixes

fedan:
Rename hw, m_hwnd to hWnd parameter function. Moved CHWCaps from OGL,DX10 and DX11 defines to all.
Use in R4 IDXGIAdapter from CreateD3D() step instead of default adapter.
Small fix R3/R4 compile shader name calculation. Not tested.


nitrocaster:
33c0571: 'Small changes' is considered bad commit message. Please give a short description of the changes you did.

[dx10HW.cpp:345](https://github.com/fedan/xray-16/commit/cfa908fc4190d099f29dcc3b7411ce47171afd41#diff-fa35b77e619fb7a36717386469da47daR345): [see common conventions (search for `Debug::Backend`)](https://github.com/OpenXRay/xray-16/wiki/Cpp-Coding-Conventions#common-things)

Regarding all renamings (including XXX comments like `Rename m_pAdapter -> pAdapter, m_pSwapChain -> pSwapChain`) see [naming conventions (search _hungarian notation_)](https://github.com/OpenXRay/xray-16/wiki/Cpp-Coding-Conventions#naming-conventions). When you do rename refactoring, make sure new names comply with the conventions.

121bfb8: was there any problem?

fedan:
> 121bfb8: was there any problem?

No, but this is not true. The name is created based on the bit set or parameter. If a player option to some change, I do not know what will be, will create a new shader or this will be used.
for example:
shaders_cache\r4\accum_indirect_msaa_3.ps\204811110011000000011010000011133211001140000

in the case of the version HLSL, it is not very noticeable :)

> m_pAdapter -> pAdapter

pointer to Adapter
hWnd: handle to Wnd.

What is m_ ?

nitrocaster:
I suppose there may be some shader cache issues when switching between 
DX10 and DX10.1. Could you check if your fix yields any result?

> pointer to Adapter
> hWnd: handle to Wnd.
> That's hungarian notation which you should avoid in this particular case.
> 
> What is m_ ?
> It's a prefix that denotes member variable (and it's hungarian notation 
> too).

On 1/29/2016 5:08 AM, fedan wrote:

> ```
> 121bfb8
> <https://github.com/OpenXRay/xray-16/commit/121bfb854d5f0af62764ed79bf2021ce33e7b576>:
> was there any problem?
> ```
> 
> No, but this is not true. The name is created based on the bit set or
> parameter. If a player option to some change, I do not know what will
> be, will create a new shader or this will be used.
> for example:
> shaders_cache\r4\accum_indirect_msaa_3.ps\204811110011000000011010000011133211001140000
> 
> in the case of the version HLSL, it is not very noticeable :)
> 
> ```
> m_pAdapter -> pAdapter
> ```
> 
> pointer to Adapter
> hWnd: handle to Wnd.
> 
> What is m_ ?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OpenXRay/xray-16/pull/119#issuecomment-176524496.

## 

Best regards, Pavel Kovalenko.

fedan:
I can not check, my graphics card dx10 shader model 4.0.

nitrocaster:
Ok, I'll check it by myself. Meanwhile, please fix the issues mentioned in my first comment.

fedan:
Ok.

PR #120, opened 2016-01-29
Fix R3/R4 compile shader name calculation.

fedan:


nitrocaster:
Although DX10.1 is only available through console (`r3_use_dx10_1`), this `0*char(o.dx10_sm4_1)` thing actually ignores that option once the shader cache has been built.

PR #125, opened 2016-03-08
Replaced "cp" with "copy"

ShokerStlk:
For some reason "cp" command not exist in Windows 7 cmd.exe

### DIFF
--- a/xrbinprep.cmd
+++ b/xrbinprep.cmd
@@ -31,17 +31,17 @@ for %%b in (%xbin%) do (
   if not exist %%b\dedicated (
     mkdir %%b\dedicated
   )
-  cp bin\dbghelp.dll %%b\dbghelp.dll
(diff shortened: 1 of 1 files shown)


nitrocaster:
If you test your changes, you'll see that this script propmts to confirm overwrite. You may want to use `copy /Y`.

I used 'cp' because it doesn't print anything if completed successfully unless I ask to do so (`cp -v`). Windows command processor doesn't understand `cp` because is a Unix utility. Since we use git and git for Windows usually comes with Unix utilities (so one doesn't have to download and install them separately), I prefer `cp` over `copy`.

nitrocaster:
I think we could solve it in other way: use `copy` if `cp` doesn't exist.

nitrocaster:
Done: be04304fadc97a6b554d8d05c7957da649a3f777

PR #128, opened 2016-04-02
add Boost to .gitignore

alexgdi:
add Boost to .gitignore


nitrocaster:
We don't have boost dependency, do we?

alexgdi:
What is the status xrWheatherEditor? Its needs Boost.

PR #134, opened 2016-04-04
Decouple lzo

alexgdi:
lzo linked as static lib.


nitrocaster:
But why DLL? It's tiny library, you can go well with static library. Another thing is that LZO must be replaced to some other library, since its licence doesn't allow to use it in non-GPL projects without author's permission.

alexgdi:
Done!

PR #136, opened 2016-04-06
Add BugTrap as submodule

alexgdi:
Add BugTrap as submodule


tamlin-mike:
+1, but doc (.md) needs updating too - just so it's not forgotten.

PR #138, opened 2016-05-20
Few fixes

Kaffeine:


src/xrServerEntities/alife_space.h
```diff
@@ -178,7 +178,7 @@ namespace ALife {
 #endif
 	}
 #ifndef	_EDITOR
-xr_token							hit_types_token							[ ];
+        extern xr_token							hit_types_token							[ ];
```
Kaffeine, line 180: Please pay extra attention on this change. I could break something here.

tamlin-mike:
Considering the amount of code using the "smart_cast", I think we should remove it from xrServerEntities and put it in a "common" place.
I have myself added loads of includes to this very file in my refactoring, and I'm getting more and more ticked off at the placement.

Kaffeine:
> I think we should remove it from xrServerEntities and put it in a "common" place.

This can be a further refactor and doesn't interfere with my changes.

src/xrServerEntities/smart_cast.h
```diff
@@ -15,11 +15,12 @@
 
 #define  TL_FAST_COMPILATION
 #undef   STATIC_CHECK
-#include <loki/typelist.h>
```
nitrocaster, line 18: TL_FAST_COMPILATION and STATIC_CHECK must have been moved with typelist include. Otherwise compilation fails because STATIC_CHECK is not defined.

PR #149, opened 2016-11-21
Revert "Merge pull request #103 from Armada651/opengl"

CrossVR:
This reverts commit 74b862f66a44cd896f913e4d9e2484c44d02b376, reversing
changes made to 7efe143728e86ef9ee72cf0fe5cd5e2e8b37033c.

This effectively removes the incomplete OpenGL backend.

Development on the OpenGL backend has been abandoned in favor of creating a Vulkan backend. The Vulkan backend will become the preferred backend for all platforms, therefore I will call it the R5 renderer. This will also solve the renderer preference problem that the OpenGL backend had.

The Vulkan backend will make use of the HLSL frontend for glslang, so the GLSL shaders are no longer required.


Kaffeine:
O_o

CrossVR:
@Kaffeine Vulkan has some notable advantages here, it's lower level so less need to work around mundane differences in the APIs, you can compile HLSL shaders directly to SPIRV and it can be used as a base for performance improvements for both platforms.

So yeah, by now it has become apparent that the advantages outweigh the added complexity. Another bonus point for me personally is that I'm now interested in learning Vulkan, so there is more motivation for me to work on this backend.

ghost:
@Armada651 If you pull this off you'll be such a hero :D God speed!

CrossVR:
@Swartz27 Well, there's the naive way and the hard way. In the naive way I'd try to implement the DX11 state manager on top of Vulkan. Which would basically make the Vulkan backend slower than the DX11 backend, but I wouldn't really need to change the structure of the rendering backend.

Or I can do it the hard way and really rethink the structure of the rendering backend to fit the Vulkan pipeline well. If I do it the hard way we should get better performance than the DX11 backend, but it will need to be well thought out so it will take longer before a working build than the naive way.

The hard way is also much more interesting. 👍

ghost:
@Armada651 That's awesome. I'd offer my help learning Vulkan, but as I am a novice I'd probably fuck things up. Maybe when I get a little better I'll try and contribute.
What I really want to do is rewrite DirectX11. The implementation is piss-poor. I also feel the rendering in the engine in-general would be better with Deffered Lighting or Forward+, as Stalker's deffered shading is slow as fuck and doesn't really support multiple BRDF's, which is something that would fit well into GSC's material system.

CrossVR:
@Swartz27 We should coordinate any restructuring efforts, so we get a design that fits Vulkan well. It would make sense to restructure DX11 before implementing the Vulkan backend, as we can do that iteratively and constantly confirm proper rendering.

About rewriting the rendering system completely, that's also interesting and I will definitely look into Forward+ as I've only heard about that technique before and I'm not yet familiar with it. However I think just updating the Deferred Shading during the restructuring to use a more modern implementation would be a better.

I want to keep the scope as small as possible to make sure we can actually complete the Vulkan backend in a timely manner.

ghost:
@Armada651 I completely agree. Vulkan is pretty important, especially since one of the goals is to make Open Xray cross-platform. Is it ok if I email you any changes I make to the rendering structure? I suck at making commits, I still don't totally get github :P Basically Forward+ and Deffered Lighting solve three problems that Deffered Shading have: 1) They can render opaque/transparent geometry without resorting to a second, forward renderer to do it, 2) It is far more efficient in the way it handles lighting, and BRDF's. My guess with either of these methods is that people would finally be able to max out r2_gi without performance issues really. 3) It can handle MSAA properly (deffered shading sucks at MSAA).

I don't know if you're familiar with MJP, but his Sample Framework that comes with his tech demo's might be a nice place to start. You can find his site here: https://mynameismjp.wordpress.com/

CrossVR:
@Swartz27 Rather than e-mailing the changes I'd appreciate it if you could discuss them either on IRC or in an issue report before you make them. We need to have some sort of designs to follow as a goal and a roadmap.

Restructuring the backend for Vulkan support isn't just making a bunch of convenient changes, you really have to think very hard about every change to the structure you make.

Also, you should take time to learn how to use Git, I'd say it's about as important as learning C++ for a programmer.

ghost:
@Armada651 You make good points, ok, will do. 
Happy new years btw.

PR #155, opened 2017-01-13
Added clang-format config

invokr:
Config taken from the dolphin emulator project. See this file as an example of how formated source code would look like: https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Core/ActionReplay.cpp

I don't mind any specific style as long as it is consistent so I hope someone else will chip in and modify the sections they need. Important points might be such things as column with (cur 100) or ident width (cur 2).

I would run this over all source files and do another pull requests if there is a consensus.

### DIFF
--- a/src/.clang-format
+++ b/src/.clang-format
@@ -1,17 +1,17 @@
 ---
 Language:        Cpp
 AccessModifierOffset: -2
-AlignAfterOpenBracket: Align
+AlignAfterOpenBracket: DontAlign
 AlignConsecutiveAssignments: false
 AlignConsecutiveDeclarations: false
 AlignEscapedNewlinesLeft: false
 AlignOperands:   true
 AlignTrailingComments: true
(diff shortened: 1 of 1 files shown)


nitrocaster:
Just in case if you haven't yet seen the conventions we use: https://github.com/OpenXRay/xray-16/wiki/Cpp-Coding-Conventions

invokr:
I didn't actually know about that, I only saw the old guidelines in text form. I'll update the format config to incorporate as many of the guidelines as possible

CrossVR:
Are we sure we want to base it on the Dolphin code style? Shouldn't we use the LLVM style as a base? `clang-format -style=llvm -dump-config > .clang-format`

CrossVR:
Meh, using dolphin as the base is fine by me, let's just run this thing and get it over with.

PR #156, opened 2017-01-15
Reformat souce code

invokr:
Running clang-format over everything but the editors and externals.


CrossVR:
Looks like we missed one coding convention: `SpacesBeforeTrailingComments` is set to 2, while the coding convention says we should have one space. Should we update the clang-format config and run it again or just update the coding convention?

CrossVR:
There also seem to be some other differences, like whether to put the pointer next to the identifier (`void* ptr` vs `void *ptr`). But mostly I'm in favor of updating the coding style document to match the clang-format used here rather than the other way around.

invokr:
The two things mentioned are rather small in terms of readability, so I don't mind either way. Let me know what you decided and I'll run regenerate the files in case we change the config.

CrossVR:
@invokr I talked with @nitrocaster , he'd like the `SpacesBeforeTrailingComments` changed to 1. Can you fix that? After that I think we can merge this PR.

nitrocaster:
Everything seems to be good now except for spaces around binary operators. Is is possible to make clang-format ignore that?

CrossVR:
@nitrocaster I can't find any option for that and I think having spaces around binary operators makes them more readable anyway. Do you have any objections against merging the PR as-is?

CrossVR:
@invokr I can't seem to compile your PR? Did you try compiling it before submitting your PR?

invokr:
@ Armada651 Ah that sucks, wtf.

I compiled it beforehand but I figured it's just formatting, there shouldn't be any changes that change behavior. I don't mind if you reverse the merge and I'll check it out in more detail. I'll have some time to work on the engine this weekend.

CrossVR:
@invokr I'll try and fix the problem, but on PRs this big you should never assume everything will be purely cosmetic.

CrossVR:
I already found the problem: This PR re-orders the include directives alphabetically which means `stdafx.h`, which should always be first, is ordered near the bottom.

CrossVR:
Ofcourse, because X-Ray depends on includes in includes, we can't really re-order now anyway, so I'll disable that entirely.

PR #160, opened 2017-01-27
Improve build system

CrossVR:
This is a PR to improve the build process:

- [x] Automatically build luajit
- [x] Don't build the unmaintained editor projects
- [x] Add every external project as a submodule

The goal is to be able to do a simple `git clone --recursive` to clone XRay 1.6 and all dependencies to allow the developer to simply hit Build Solution without errors.


jakebriggs:
I still had to compile luajit from the command line before everything compiled - but once I did that, it compiled fine.

andrew-boyarshin:
@Armada651 why have you decided to switch to static build of Crypto++? FIPS conformance is not required by OpenXRay engine (we do not handle money or sensitive user data), plus both before and after this commit we cannot claim to be FIPS conformant.

CrossVR:
@andrew-boyarshin That's exactly the reason why I switched to a static build. Crypto++ automatically does some FIPS validation tests if it's a DLL build, most notably it checks the hash of the DLL which fails because I didn't include the hash in the output DLL.

Also it is only used in a very small part of the engine, namely checking if certain resource files are "pure" during online gameplay.

andrew-boyarshin:
@Armada651 oh, I see. I was not aware of runtime validation tests, and their website was not informative enough to tell me that fact.

andrew-boyarshin:
Tried out and confirmed to be working. Looks Good To Me.

andrew-boyarshin:
What has changed so that you've force-pushed? Just rebasing on latest commit?

P.S. When it is going to be reviewed?

Xottab-DUTY:
@andrew-boyarshin he just rebased his branch to the latest commits after merging my PRs. 
P.S. So, the branch isn't updated to the latest merged precompiled shaders PR.

CrossVR:
That doesn't matter as long as there are no merge conflicts.

andrew-boyarshin (changes requested):
Debug:
Works fine, no flaws, although I also had to apply xrSound CDB library linking fix from another PR for Debug code to compile.

Release:
I had some troubles. zlib in Release mode requires contrib asm linked in. After building these asm files (manually, does not quite match the task of the PR) and successfully linking it with the rest of the zlib I've encountered issue with BugTrap not being able to link to it due to SAFESEH enabled ("no pure asm" requirement).
This should be resolved in order to be able to merge it. Of course removing ASMV and ASMINF preprocessor flags helped to resolve the issue. And taking into consideration asm code is considered experimental in zlib - I think that's the way to go.

CrossVR:
@andrew-boyarshin Issues are fixed, I opted to disable the assembler code, because it doesn't seem like it wasn't used before either.

andrew-boyarshin:
@Armada651 will check this out and update my review.

CrossVR:
Since no one seems to disagree with these changes following the review I'm going to merge them tomorrow. Let me know if anyone has time to do a full review.

andrew-boyarshin:
In fact, I've reviewed your modifications and they look good to me. I just had little time to do the thorough testing in the game.

andrew-boyarshin (approved):

CrossVR:
The only real risk to this PR is the fact that the change to Crypto++ is mostly untested (only some sanity tests were done). The only consequence of that would be that it would break "pure" multiplayer. I don't think multiplayer is a priority anyway, so if it turns out we broke that we can fix it later.

PR #166, opened 2017-02-19
New splash

Xottab-DUTY:
Replaced splash and added .psd for everyone who wants to edit it.
There's 2 .psd files: the HD one (2667x1200) and splash sized (500x268)
The reason for 2 .psd's is that you can't just resize the HD one to splash size: you need to additionally edit it, if you want a fancy image. So I decided to make it easier =)


ShokerStlk:
Nice looking. It's probably better to move it into gamedata folder, instead of hardcoding to .exe.

Xottab-DUTY:
It's definitely better to move it to gamedata folder. But now splash loading it's almost first what .exe do, where gamedata scanning and loading resources from there done much later. In this case, this may be a problem.

nitrocaster (approved):

PR #172, opened 2017-03-08
Stretched skyboxes and PVS-Studio bugfixes

Xottab-DUTY:


Xottab-DUTY:
Blin, I saw that mistake only now. I need to rename East-North to East-West.

src/xrEngine/Environment.cpp
```diff
@@ -550,6 +550,9 @@ void CEnvironment::calculate_dynamic_sun_dir()
 
     clamp(cosAZ, -1.0f, 1.0f);
     float AZ = acosf(cosAZ);
+    extern u32 ps_r_sun_direction;
```
CrossVR: This should be changed into a member of `CEnvironment` called `m_sun_direction`. Then you can set it by making a command class that sets `g_pGamePersistent->Environment().m_sun_direction` in `Execute()`.

See https://github.com/OpenXRay/xray-16/blob/dev/src/xrEngine/xr_ioc_cmd.cpp#L463 as an example. Also you could just allow arbitrary sun directions if you make it a float in degrees and convert it to radians when setting `m_sun_direction`.

src/xrGame/space_restriction.cpp
```diff
@@ -152,7 +152,7 @@ CSpaceRestriction::CBaseRestrictionPtr CSpaceRestriction::merge(
     RESTRICTIONS::const_iterator I = temp_restrictions.begin();
     RESTRICTIONS::const_iterator E = temp_restrictions.end();
     for (; I != E; ++I)
-        temp = strconcat(sizeof(S), S, *temp, ",", *(*I)->name());
+        temp = strconcat(strlen(S), S, *temp, ",", *(*I)->name());
```
CrossVR: `strlen` is incorrect here, because the string `S` is empty, so it would always fail. Instead `sizeof(S)` should be replaced with `acc_length`, which is the variable used as the size to allocate `S`.

src/xrGame/Actor_Network.cpp
```diff
@@ -685,7 +685,7 @@ BOOL CActor::net_Spawn(CSE_Abstract* DC)
     inventory().SetPrevActiveSlot(NO_ACTIVE_SLOT);
 
     //-------------------------------------
-    m_States.empty();
+    m_States.clear(); // Xottab_DUTY: check if replace is correct
```
CrossVR, line 688: Yeah, this looks correct, but should probably be tested if it doesn't have unintended consequences.

CrossVR (changes requested):

src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -713,6 +728,7 @@ void CCC_Register()
 
     // Render device states
     CMD4(CCC_Integer, "r__supersample", &ps_r__Supersample, 1, 4);
+    CMD2(CCC_SunDirection, "r__sun_direction", &ps_sun_direction);
```
CrossVR: I think this should be an `rs_` command, `r__supersample` seems to be an outlier here.

src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -459,6 +460,20 @@ class CCC_SND_Restart : public IConsole_Command
...
+float ps_sun_direction = 0.0f;
+class CCC_SunDirection : public CCC_Float
+{
+public:
+    CCC_SunDirection(LPCSTR N, float* V) : CCC_Float(N, V, 0.0f, 360.0f) {}
+    virtual void Execute(LPCSTR args)
+    {
+        CCC_Float::Execute(args);
+        if (g_pGamePersistent)
+            g_pGamePersistent->Environment().m_sun_azimuth = ps_sun_direction * (PI / 180.0f);
```
CrossVR: Can you test what happens when `r__sun_direction 180` is in the user.ltx file? Will the sun direction still be corrected?

src/xrEngine/xr_ioc_cmd.cpp
Xottab-DUTY: rs_sun_direction seems more outlined with such commands as rs_stats, rs_fullscreen, rs_gamma and others than r__sun_direction with r__ rendering commands. Maybe the placement isn't good, but name is pretty matches to the other commands

src/xrCore/xrDebug.cpp
```diff
@@ -97,7 +97,7 @@ size_t xrDebug::BuildStackTrace(char* buffer, size_t capacity, size_t lineCapaci
 
 size_t xrDebug::BuildStackTrace(EXCEPTION_POINTERS* exPtrs, char* buffer, size_t capacity, size_t lineCapacity)
 {
-    memset(buffer, capacity * lineCapacity, 0);
+    memset(buffer, 0, capacity*lineCapacity);
```
Xottab-DUTY, line 100: May be with this fix stack trace will be more appropriate, which is not in some cases(for example, just gives nothing)

src/xrEngine/xr_ioc_cmd.cpp
Xottab-DUTY: No, it doesn't. ~By the way... Something went wrong... r__sun_direction 0 and sun born from West, but should born from South.~ All right with sun, except only it should get params from user.ltx

Xottab-DUTY:
@andrew-boyarshin It's not only rendering, but includes other various fixes. (I should pull these in other request, but what's done is done)

andrew-boyarshin:
@Xottab-DUTY I guess the term "anomalies" was used from original PVS-Studio article. It was some kind of reference to gameplay of S.T.A.L.K.E.R., so it was ok. Here it was unexpected and led to misunderstanding.

CrossVR:
@Xottab-DUTY You can still put those in another PR so we can at least merge the PVS fixes now.

Simply use these commands:

```
git branch render-fixes
git reset --hard c328536
git push -f origin pvs-fixes
git push origin render-fixes
```

Then you can submit a new PR for your new `render-fixes` branch.

CrossVR:
Actually you should also remove the PVS commit from the `render-fixes` branch, but we can fix that in a simple rebase after this PR is merged

Xottab-DUTY:
Detached sun direction from this PR, it's not ready. You can find it in `sundirection` of my fork.

CrossVR (approved):

PR #177, opened 2017-03-15
gamedata: Add pre-compiled R2 shaders.

CrossVR:
Includes a python script to compile them from scratch. The script searches shaders for macros and compiles all permutations of those macros in parallel. Afterwards it prefixes the data with a CRC32 checksum which is required by the engine. The engine itself is already equipped to load these shaders from gamedata.

The advantage of these pre-compiled shaders are faster load times, but it also has significant advantages for the cross-platform port as there are significantly more cross-compilers available for DX bytecode than for HLSL shaders.

The script is a bit ugly, as it was written just to get the job done. But when we're going to do significant work on the HLSL shaders, then the script can be cleaned up and expanded to cover more renderers.

*Note:* As can be seen in the script, there are currently two assumptions:
- Every supported GPU supports `USE_HWSMAP` so this option is force-enabled, this means we don't support GPUs that don't support pixel shader model 3.0 (DX9c)
- The `SKIN_COLOR` command-line parameter is never used so this option is disabled, it looks like this option never worked in the first place.


andrew-boyarshin:
@Armada651 full review will be as I've said no earlier than tomorrow, yet a Python issue is clearly seen: non-PEP8 conformant code. AFAIK, PEP8 also doesn't specify that you shouldn't wrap the only ```if```` block condition in parentheses (compared to most other languages), although it is always considered bad practice to do so. Should be easy to fix without need to recompile all shaders and push them again. Also, maybe split commit in two? One with script only and one with all the recompiled shaders in order to not to cause GitHub slowdowns. Also easy to improve.
UPD: Why not to specify supported Python versions in the first lines of the script? In order not to confuse users.

CrossVR:
@andrew-boyarshin There's more wrong with the script than just the codestyle. It was written to get these shaders compiled as quick as possible. I could keep the script on a personal branch until it's up to the quality we would need it to be.

andrew-boyarshin:
@Armada651 why? Keep it here, but in separate commit with codestyle fixed at least partially (as it is already working). Further improvements for further PRs.

CrossVR:
Sure, just let me know which lines you'd like to see changed. I'm not familiar at all with Python code styles.

andrew-boyarshin:
[PEP8](https://www.python.org/dev/peps/pep-0008/)
[Reasonably good online PEP8 checker](http://pep8online.com/)

CrossVR:
Done, ready for review.

res/compile.py
```diff
@@ -0,0 +1,193 @@
+import os
+import subprocess
+import re
+import binascii
+import struct
+from multiprocessing.pool import ThreadPool
+
+# Set the path to the fxc compiler
+fxc = 'C:\\Program Files (x86)\\Microsoft DirectX SDK (June 2010)\\Utilities\\bin\\x86\\fxc.exe'
```
Xottab-DUTY: Why not to use DXSDK_DIR system variable?
Added few seconds later: oh, those backslashes, ~is~ that's why?

Xottab-DUTY (commented):
First commit says that it includes .py script, while it isn't =)

res/compile.py
andrew-boyarshin: @Xottab-DUTY do you suggest using raw string literals? ```"""text"""```? That's arguable. It won't make the string shorter.

Xottab-DUTY: No no, I'm just suggest to use DXSDK_DIR instead of absolute path, if we can use it of course.

Sorry, when I said:
> those backslashes **is** why?

I mean
> those backslashes, **that's** why _we can't use DXSDK_DIR_?

CrossVR: The reason was simply that I didn't take the time to look up how to query system variables. I've updated the script.

res/compile.py
```diff
@@ -0,0 +1,193 @@
+import os
+import subprocess
+import re
+import binascii
+import struct
+from multiprocessing.pool import ThreadPool
+
+# Set the path to the fxc compiler
+fxc = os.environ['DXSDK_DIR'] + 'Utilities\\bin\\x86\\fxc.exe'
```
andrew-boyarshin, line 9: Of course, it would be better to use [pathlib](https://docs.python.org/3/library/pathlib.html), but it is not necessary for now and doesn't prevent from merging this PR.
And there should be environment variable existence check (non-critical).

res/compile.py
```diff
@@ -0,0 +1,193 @@
...
+def addCRC(filename):
+    with open(filename) as file:
+        file = open(filename,'r+b')
+        buf = file.read()
+        crc = (binascii.crc32(buf) & 0xFFFFFFFF)
+        file.seek(0)
+        file.write(struct.pack("<I", crc))
+        file.write(buf)
+
+def findOptions(shader):
```
andrew-boyarshin, line 87: It is better to have 2 new lines between methods instead of 1. Non-critical, doesn't affect mergeability.

res/compile.py
```diff
@@ -0,0 +1,193 @@
...
+    list = []
+    with open(shader) as f:
+        buffer = f.read()
+        buffer = re.sub(commex, "", buffer) # remove commented lines
+        for option in options:
+            if option in buffer:
+                list.append(option)
+        includes = inclex.findall(buffer)  # find all include files
+        for include in includes:
+            list += findOptions(os.path.join(os.path.dirname(shader), include))
```
andrew-boyarshin, line 97: Throughout the code there are lots of ```os.path.join``` usages. Importing like ```from os.path import join as pathjoin``` and using this alias would look better. Non-critical.

res/compile.py
```diff
@@ -0,0 +1,193 @@
...
+    profile = '%s_%s' % (stage, featureLevel)
+    if len(entry) > len('main'):
+        profile = entry[5:]
+    
+    object = os.path.join(objects, '%s.%s' % (name, stage))
+    if os.path.isfile(object):
+        print('Already compiled, skipping object %s' % flags)
+        return
+    
+    args = [fxc, '/nologo', '/LD', '/Zpr', '/E%s' % entry, '/T%s' % profile]
```
andrew-boyarshin, line 155: No optimization?

andrew-boyarshin (approved):
No critical notices. Can be merged with no troubles and later improved.

res/compile.py
CrossVR: /O1 is the default optimization level and any level above that is reserved for future use. The reason for not specifying the optimization level is because some shaders use old 1.x shader models which don't support any optimization level.

PR #179, opened 2017-03-15
Fix xrWeatherEditor compilation and one bug

Xottab-DUTY:


CrossVR:
Ah yes, I remember this, I thought I fixed it before committing, but apparently I forgot.

Thanks for fixing it.

CrossVR (approved):

PR #180, opened 2017-03-25
Fix xrSound wont compile on Debug configuration

Xottab-DUTY:
Unexpected one! Sorry for this.
Plus small include path fixes

### DIFF
--- a/src/xrSound/SoundRender_Core.h
+++ b/src/xrSound/SoundRender_Core.h
@@ -5,7 +5,6 @@
 #include "SoundRender.h"
 #include "SoundRender_Environment.h"
 #include "SoundRender_Cache.h"
-#include "soundrender_environment.h"
 
 class CSoundRender_Core : public CSound_manager_interface
 {
--- a/src/xrSound/stdafx.h
+++ b/src/xrSound/stdafx.h
@@ -28,14 +28,14 @@
 #include <vorbis/vorbisfile.h>
 
 #include "xrCDB/xrCDB.h"
-#include "sound.h"
+#include "Sound.h"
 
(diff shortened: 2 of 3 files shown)


src/xrSound/stdafx.cpp
```diff
@@ -15,4 +15,6 @@
 #pragma comment(lib, "OpenAL32B.lib")
 #pragma comment(lib, "dsoundb.lib")
 //#pragma comment(lib, "xrAPI.lib")
+#elif DEBUG
+#pragma comment(lib, "xrCDB.lib") // Xottab_DUTY: For some reason it wont compile without it.
```
CrossVR: Why would you put this back into the pre-compiled header when in the last PR you moved library inputs to the project files?

CrossVR (changes requested):

src/xrSound/stdafx.cpp
Xottab-DUTY: Without direct point to xrCDB.lib xrSound wont compile. I can put this in project file, but I think we should investigate in future into this (Why in debug just only project reference wont work, just like in release). So, it's a reminder in the future.

P.S. I feel like I can do some mistakes in the sentence.

CrossVR: I don't agree with that reasoning, pulling in extra library inputs for the linker is not a problem at all. You don't even need to limit it to the Debug configuration. Only objects from the library that are actually used will be linked to.

Please move this to the project file, then we can merge this PR.

CrossVR (changes requested):

CrossVR (approved):

PR #181, opened 2017-03-29
Various PVS warning fixes

Xottab-DUTY:


CrossVR:
I don't think it's necessary to change postfix to prefix. A sane compiler wouldn't actually generate less efficient code using postfix. It's an unnecessary micro-optimization.

andrew-boyarshin:
@Armada651 wouldn't hurt either. Just makes reviewing process a bit longer.
UPD: Oh, ok. Never mind.

Xottab-DUTY:
@andrew-boyarshin Well.. I already dropped that commit :D

CrossVR (approved):

PR #182, opened 2017-04-05
Fix incorrect usage of #pragma warning

Xottab-DUTY:


Xottab-DUTY:
Tested for a while. Works without any problems.

PR #183, opened 2017-04-11
Replaced STATIC_CHECK and mk_pair

Xottab-DUTY:


eotaoza:
Is it possible to get from source codes more fps?

Xottab-DUTY:
It's possible. The question is how to do that. Refactoring, optimizations, etc..

Xottab-DUTY:
Tested for a while. Works without any problems.

PR #188, opened 2017-08-06
delete duplicate

alexgdi:
duplicate and unused files deleted. original stored on src/xrCore/Compression


Xottab-DUTY:
You could use `git reset` and `git push -f` to remove occasionally commit that you have pushed here

alexgdi:
This is what I did. And something happened that is not right.

Xottab-DUTY:
That's a bad thing :D

PR #189, opened 2017-08-07
deleted duplicate 2

alexgdi:
duplicate and unused files deleted. original stored on src/xrCore/Compression


Xottab-DUTY:
@alexgdi, files are included into xrCore.vcxproj. Are you sure, that these files are unused? If they are unused you should delete it from project file too.

PR #194, opened 2018-03-14
Reimplements the onboard functionality of the stack tracing

NeoAnomaly:


src/xrCore/xrDebug.cpp
```diff
@@ -70,57 +77,204 @@ xrDebug::CrashHandler xrDebug::OnCrash = nullptr;
...
-    while (traceDump)
-    {
-        lstrcpy(buffer + frameCount * lineCapacity, traceDump);
-        frameCount++;
-        traceDump = GetNextStackTraceString(flags, exPtrs);
-    }
-    return frameCount;
+	SStringVec stackTrace = BuildStackTrace();
+	Msg("%s", header);
+	for each (const auto& frame in stackTrace)
```
nitrocaster: Use standard range-based for instead of non-standard MS-specific 'for each'.

src/xrCore/xrDebug.h
```diff
@@ -1,16 +1,13 @@
...
-{
-    static const size_t Capacity = 100;
-    static const size_t LineCapacity = 256;
-    char Frames[Capacity * LineCapacity];
-    size_t Count;
-
-    char* operator[](size_t i) { return Frames + i * LineCapacity; }
-};
+#pragma warning(push)
+#pragma warning(disable : 4091)	/// 'typedef ': ignored on left of '' when no variable is declared
```
nitrocaster: Disable warning locally, not globally.

src/xrCore/xrDebug.h
```diff
@@ -78,13 +74,27 @@ class XRCORE_API xrDebug
...
     static void WINAPI PreErrorHandler(INT_PTR);
     static void SaveMiniDump(EXCEPTION_POINTERS* exPtrs);
+
+	///
+	/// Next members relates to stack tracing
+	///
+	static bool m_SymEngineInitialized;
+
+	static xr_vector<xr_string> BuildStackTrace(PCONTEXT threadCtx, u16 maxFramesCount);
+	static bool GetNextStackFrameSring(LPSTACKFRAME stackFrame, PCONTEXT threadCtx, xr_string& frameStr);
```
nitrocaster: GetNextStackFrameS**t**ring?

src/xrCore/xrDebug.h
```diff
@@ -78,13 +74,27 @@ class XRCORE_API xrDebug
         const char* arg1 = nullptr, const char* arg2 = nullptr);
     static void DoExit(const std::string& message);
 
+	///
+	/// Note: DbgHelp is singlethreaded, so you must synchronize calls to these functions
```
nitrocaster: Why not wrap the content of LogStackTrace into a critical section?

nitrocaster:
Additionally, please adhere to project conventions (tabs, variable names, spacing, etc). It took a great effort to fix the code style.

src/xrCore/xrDebug.cpp
NeoAnomaly: Accept. I'll fix it

src/xrCore/xrDebug.h
NeoAnomaly: Ok

src/xrCore/xrDebug.h
NeoAnomaly: I didn't find a RAII compatible implementation in the current codebase. So, for minimizing size of the changes at now, I left the sync question for the caller

Xottab-DUTY: Lock isn't RAII compatible? :(

NeoAnomaly:
```
struct LockImpl
{
    std::recursive_mutex mutex;
};

Lock::~Lock()
{
    delete impl;    /// impl is LockImpl
}
```

> The behavior of a program is undefined if a recursive_mutex is destroyed while still owned by some thread.

Xottab-DUTY:
@nitrocaster, so, are we okay to merge?

ForserX:
@nitrocaster,какие люди. День добрый.

NeoAnomaly:
I've been afraid @nitrocaster with the coding covention 😁

src/xrCore/xrDebug.cpp
```diff
@@ -682,7 +681,7 @@ static void invalid_parameter_handler(
         line = __LINE__;
         xr_strcpy(mbFile, __FILE__);
     }
-    xrDebug::Fail(ignoreAlways, {mbFile, int(line), mbFunction}, mbExpression, "invalid parameter");
+    xrDebug::Fail(ignoreAlways, { mbFile, int(line), mbFunction }, mbExpression, "invalid parameter");
```
nitrocaster: Why?

src/xrCore/xrDebug.cpp
```diff
@@ -81,200 +80,200 @@ bool xrDebug::m_SymEngineInitialized = false;
...
 {
-	SStringVec stackTrace = BuildStackTrace();
-	Msg("%s", header);
-	for each (const auto& frame in stackTrace)
-	{
-		Msg("%s", frame.c_str());
-	}
+    SStringVec stackTrace = BuildStackTrace();
+    Msg("%s", header);
+    for(const auto& frame : stackTrace)
```
nitrocaster: Put one space after keywords;
Put pointer and reference next to the identifier;
https://github.com/OpenXRay/xray-16/wiki/Cpp-Coding-Conventions

src/xrCore/xrDebug.cpp
```diff
@@ -81,200 +80,200 @@ bool xrDebug::m_SymEngineInitialized = false;
...
+    {
```
nitrocaster: if/for/while statement that takes more than one line should always have brackets same thing goes when the then/loop part is more than one line.

nitrocaster:
Oh. So we still have broken style everywhere because of [this](https://github.com/OpenXRay/xray-16/commit/c96c2e4cc9fadf29c023464ba1885e7351d68943#diff-5601eafd1e0b83a71e7853cab7963713).
No wonder why you format pointers/references wrong.
After so much time was spent on making everything consistent we came to nothing.
Disregard my comments on pointers, but please fix other issues to make your code consistent with the rest. In particular:

- Private fields are written in `camelCase`
- No Hungarian notation (i.e. no `m_`, `p` and other prefixes referring to type/visibility)

I appreciate your work and hope that you understand why it is important to follow coding conventions.

nitrocaster (changes requested):

src/xrCore/xrDebug.cpp
```diff
@@ -70,57 +77,204 @@ xrDebug::CrashHandler xrDebug::OnCrash = nullptr;
...
+{
+	if (m_SymEngineInitialized)
+	{
+		SymCleanup(GetCurrentProcess());
+
+		m_SymEngineInitialized = false;
+	}
+}
+
+xr_vector<xr_string> xrDebug::BuildStackTrace(PCONTEXT threadCtx, u16 maxFramesCount)
```
nitrocaster: Can we not copy the vector with stack trace here?

src/xrCore/xrDebug.h
```diff
@@ -78,13 +74,27 @@ class XRCORE_API xrDebug
...
+	///
+	static void LogStackTrace(const char* header);
+	static xr_vector<xr_string> BuildStackTrace(u16 maxFramesCount = MaxFramesCountDefault);
```
nitrocaster: Apparently, `MaxFramesCountDefault` exists only to allow you to not write `512` here, which is not the case when you want to define constants. When you use `BuildStackTrace` anywhere else, you want to know its signature and default arguments, but instead of `512` you'll see `MaxFramesCountDefault` requires you open the header and see the actual value.

src/xrCore/xrDebug.h
```diff
@@ -78,13 +74,27 @@ class XRCORE_API xrDebug
...
 private:
     static void FormatLastError(char* buffer, const size_t& bufferSize);
-    static size_t BuildStackTrace(EXCEPTION_POINTERS* exPtrs, char* buffer, size_t capacity, size_t lineCapacity);
     static void SetupExceptionHandler(const bool& dedicated);
     static LONG WINAPI UnhandledFilter(EXCEPTION_POINTERS* exPtrs);
     static void WINAPI PreErrorHandler(INT_PTR);
     static void SaveMiniDump(EXCEPTION_POINTERS* exPtrs);
+
+	///
```
nitrocaster: One can easily understand that just by looking at the following declarations. Avoid obvious comments - they waste space.

src/xrCore/xrDebug.cpp
NeoAnomaly: And if is it one line? What's wrong here?

src/xrCore/xrDebug.cpp
NeoAnomaly: You find this no manually, didn't you? If you've been having some linter, why doesn't documentation have link to it? This would make life easier for everyone.

nitrocaster (commented):
That's also supposed to mean that if/for/while statement which takes exactly one line should not have braces around the statement. Reason: adding braces duplicates the vertical size, which is essentially bad when you have multiple precondition checks.

src/xrCore/xrDebug.cpp
Xottab-DUTY: For now we only have a Clang-Format config that's ready for use

nitrocaster: I look at diff and every single change shows up right there. Every change is made with a reason behind it, right? I guess there's no reason here because this change was made by VS which automatically reformatted pasted code. That is why you should check the diff before committing.

nitrocaster: Coding conventions page in wiki doesn't explicitly require that, but there's an example marked as `GOOD` that shows initialization formatting:

` Size size = {1024, 16};`

src/xrCore/xrDebug.cpp
nitrocaster: https://github.com/OpenXRay/xray-16/pull/194#pullrequestreview-104101807

src/xrCore/xrDebug.cpp
NeoAnomaly: As far as I know compiler applies the move optimization always then it's possible here. However, stack tracing is a debugging feature and should not be used very often.

src/xrCore/xrDebug.cpp
NeoAnomaly: >I look at diff and every single change shows up right there. Every change is made with a reason behind it, right?

I like your approach

src/xrCore/xrDebug.cpp
nitrocaster: Agreed.

NeoAnomaly:
>That's also supposed to mean that if/for/while statement which takes exactly one line should not have braces around the statement

Well. I consider the documentation should have remarks about that, it's not obviously.

Xottab-DUTY:
@nitrocaster is there major code style issues in this PR? I see only minor of these. I want to fix it myself so that we don't spend time on this.

nitrocaster (approved):

nitrocaster:
Merged. Thanks for your contribution, @NeoAnomaly

PR #199, opened 2018-05-06
Try to fix "SMAP_size" crash in OpenGL render on AMD video cards part #1

SkyLoaderr:


Xottab-DUTY:
Thanks for your contribution!

Issue #219, opened 2018-06-29
Debug Assertion Failed! Expression: (mask&~(_MCW_DN|_MCW_EM|_MCW_RC))==0


TarasHots:
Compiled solution successfully, and this error occurs when trying to launch engine in debugging mode.  Intro image is shown only, and it crashes after that. <br>Happens in xrCore.Detect() line 250 on `_control87(_PC_64, MCW_PC);` 

I can provide more info, just ask what exactly. Only thing I changed is target Windows SDK from 8.1 to 10.0.17134.0

OS: Windows 10
IDE: Visual Studio 2017 15.7.4

Here is the screenshots how it looks like:

<a href="https://imgur.com/k333Eho"><img src="https://i.imgur.com/k333Eho.png" title="source: imgur.com" /></a>

<a href="https://imgur.com/H0VUtNE"><img src="https://i.imgur.com/H0VUtNE.png" title="source: imgur.com" /></a>

UPDATE 1: Seems like related to this https://svn.boost.org/trac10/ticket/4964

TarasHots:
Seems like I found how to fix it, but need your approval, guys. I'd also mention that I'm building x64 project

> "On the x64 architecture, changing the floating point precision is not supported. If the precision control mask is used on that platform, an assertion and the invalid parameter handler is invoked, as described in Parameter Validation."

If I'll add control dependong on system architecture, would it be correct fix? (for me it seems to have sense). What do you think? I want to make a PR, but it would be good to make sure am I on the right way.

PR #221, opened 2018-06-30
Corrected precision control for x64 architecture on Windows

TarasHots:
Related to #219

### DIFF
--- a/src/xrCore/_math.cpp
+++ b/src/xrCore/_math.cpp
@@ -72,7 +72,9 @@ namespace FPU
 XRCORE_API void m24()
 {
 #if defined(WINDOWS)
+#ifndef XR_X64
     _controlfp(_PC_24, MCW_PC);
+#endif
     _controlfp(_RC_CHOP, MCW_RC);
 #elif defined(LINUX)
     fpu_control_t fpu_cw;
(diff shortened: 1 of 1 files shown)


Xottab-DUTY:
One minor issue with this PR: we are using our own macros `XR_X64` and `XR_X86`. I think it would be better to use `XR_X64` instead of `_WIN64`. Please, correct me if I'm wrong.

Xottab-DUTY:
@intorr can you say something about this?

Xottab-DUTY:
@intorr can you say something about this?

intorr:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/control87-controlfp-control87-2

_MCW_PC (Precision control) not supported on ARM or x64 platforms.
_MCW_RC (Rounding control) supported on all platforms.

Xottab-DUTY:
According to what @intorr said, we should only disable precion but not rounding:
```cpp
#ifndef XR_X64
     _controlfp(_PC_24, MCW_PC);
#endif
```

TarasHots:
Thanks for quick feedback! I’ll make all changes as soon as possible

Xottab-DUTY:
@TarasHots, also, try to do this with rebase or commit squash, so that this will be one-commit-PR

TarasHots:
@Xottab-DUTY fixed notices and made it as one-commit PR

PR #223, opened 2018-07-06
Detail objects shadows.

RainbowZerg:


Xottab-DUTY:
Much thanks for this and other your contributions!

PR #248, opened 2018-10-07
 Port src/Common and other stuff on Linux

q4a:


src/Common/Common.hpp
```diff
@@ -6,10 +6,12 @@
 #include "Common/FSMacros.hpp"
 #include "Include/xrAPI/xrAPI.h"
 
-#if __has_include(<SDL.h>)
-#include <SDL.h>
+#if defined(WINDOWS)
+#if __has_include("SDL.h")
+#include "SDL.h"
```
Xottab-DUTY, line 11: Why `<>` is replaced with `""`???

Xottab-DUTY: Ok, I've got the answer.

PR #251, opened 2018-10-10
Replaced D3DPT_TRIANGLEFAN with D3DPT_TRIANGLELIST in Portal class

SkyLoaderr:


src/Layers/xrRender/r__sector.cpp
```diff
@@ -33,40 +33,47 @@ void CPortal::OnRender()
...
-        RCache.dbg_Draw(D3DPT_LINESTRIP, &*(V.begin() + 1), V.size() - 2);
+        RCache.set_Shader	(RImplementation.m_WireShader);
+        RCache.dbg_Draw		(D3DPT_LINESTRIP,&*V.begin(),V.size()-1);
         if (bDebug)
         {
             RImplementation.rmNormal();
-        }
-        else
-        {
+        } else {
```
Xottab-DUTY, line 76: Should be
```cpp
}
else
{
```
But we can just remove braces here.

src/Layers/xrRender/r__sector.cpp
```diff
@@ -33,40 +33,47 @@ void CPortal::OnRender()
...
             Device.SetNearer(TRUE);
         }
-        RCache.set_Shader(RImplementation.m_WireShader);
-        RCache.dbg_Draw(D3DPT_LINESTRIP, &*(V.begin() + 1), V.size() - 2);
+        RCache.set_Shader	(RImplementation.m_WireShader);
+        RCache.dbg_Draw		(D3DPT_LINESTRIP,&*V.begin(),V.size()-1);
```
Xottab-DUTY, line 72:
```cpp
RCache.set_Shader(RImplementation.m_WireShader);
RCache.dbg_Draw(D3DPT_LINESTRIP, &*V.begin(), V.size() - 1);
```

src/Layers/xrRender/r__sector.cpp
```diff
@@ -33,40 +33,47 @@ void CPortal::OnRender()
...
+			V[k].set(poly[k], uPortalColor);
+		V.back().set(poly[0], uPortalColor);
+        
         if (bDebug)
         {
             RImplementation.rmNear();
-        }
-        else
-        {
+        } else {
```
Xottab-DUTY, line 68: Should be
```cpp
}
else
{
```
But we can just remove braces here.

src/Layers/xrRender/r__sector.cpp
```diff
@@ -33,40 +33,47 @@ void CPortal::OnRender()
...
-        Fvector C = {0, 0, 0};
-        for (u32 k = 0; k < poly.size(); k++)
-        {
-            C.add(poly[k]);
-            V[k + 1].set(poly[k], 0x800000FF);
-        }
-        V.back().set(poly[0], 0x800000FF);
-        C.div((float)poly.size());
-        V[0].set(C, 0x800000FF);
+        xr_vector<FVF::L> V; V.resize(poly.size()*3);
```
Xottab-DUTY, line 36: Please, move resize() to the next string:
```cpp
xr_vector<FVF::L> V;
V.resize(poly.size()*3);
```

Xottab-DUTY (changes requested):
Thank you for the effort! Could you fix the codestyle? I described what needs to be corrected.

PR #253, opened 2018-10-12
Update build-linux.txt

drug007:
From memory, inaccuracy possible

### DIFF
--- a/doc/howto/build-linux.txt
+++ b/doc/howto/build-linux.txt
@@ -19,16 +19,24 @@ JPEG
 Ubuntu 18.04 build dependencies:
 sudo apt install git cmake libglew-dev libfreeimage-dev liblockfile-dev libopenal-dev libtbb-dev libcrypto++-dev libpugixml-dev libogg-dev libtheora-dev libvorbis-dev libsdl2-dev liblzo2-dev libjpeg-dev libncurses5-dev libreadline-dev
 
+Clone command:
+```
+git clone https://github.com/OpenXRay/xray-16.git --recurse-submodules
+```
+
 Build commands:
 ------
+```
 cd xray-16
 git submodule update --init --recursive
-mkdir bin && cd bin
+mkdir build
+cd build
 cmake ..
-make
+make -jX # where X is the number of cores in your system
+```
 
 #To enable debugging:
-cmake .. -DCMAKE_BUILD_TYPE=Debug
+cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo # Building with Debug flag isn't supported now
 #To use clang:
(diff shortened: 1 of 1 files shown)


doc/howto/build-linux.txt
```diff
@@ -63,13 +63,21 @@ All mentioned libraries are already in repository as submodules!
 
 Setup:
 ------
-- Install S.T.A.L.K.E.R Call of Pripyat
+- Install S.T.A.L.K.E.R Call of Pripyat using wine. For example if you have a cd-rom you can type `wine setup.exe` to start installation. It's recommended do not use path with spaces, for instance you can use "c:\cop" as a installation path (it will be `~/.wine/drive_c/cop` path on your host linux machine.
```
Zegeri: Thank you for the contribution.
You should add some line breaks here, so it looks better on github.  There's a missing parentheses ')' at the end of the line.

doc/howto/build-linux.txt
```diff
@@ -63,13 +63,21 @@ All mentioned libraries are already in repository as submodules!
...
+Command to clone, build and run the engine:
+```
+git clone https://github.com/OpenXRay/xray-16.git --recurse-submodules
+cd xray-16
+git submodule update --init --recursive # just in case
+mkdir build
+cd build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
+mkdir ~/.wine/drive_c/cop/bin-linux-dbg # here will installed binaries with dbg symbols
+make -jX # instead of X should count of your cores*processor count
```
Zegeri: This sentence has some mistakes. I would suggest either:
```
make -jX # where X is the number of cores in your system
```
or just:
```
make -j$(nproc)
```

q4a:
Thanks. just 1 fix:
After
git clone https://github.com/OpenXRay/xray-16.git --recurse-submodules
We don't need to do:
git submodule update --init --recursive

drug007:
@q4a I left it just in case, but I wasn't sure it was useful, so it's fixed now.

Offtopic: what about using markdown instead of plain text?

drug007:
@q4a, @Zegeri, and the others: please do not merge this PR before I rebase it to current head. If you think PR looks good just write `LGTM` (looks good to me) for example or make other sign.

Frankly we shouldn't merge any PR until it is rebased to the head. This easy rule makes history much more pleasant to work with.

Xottab-DUTY:
@drug007 no need for this. There's no conflicts, so the rebase is unnecessary.

drug007:
@Xottab-DUTY there is _big_ need for this. Our repository history is awful due to this fact. Conflicts aren't the only reason to make rebase.

drug007:
@Xottab-DUTY up

PR #259, opened 2018-10-21
Fix corrupted game saves

FreeZoneMods:
After refactoring the registry container the registries are saving and loading in incorrect (reverse) order. It leads to loading game tasks registry before map locations registry. But game tasks registry requires data from the map locations registry (map spots, for example) and the game crashes when it can't find the required data.


Xottab-DUTY:
Thank you for the effort!

PR #264, opened 2018-10-30
Fix debug build

drug007:


src/xrGame/CMakeLists.txt
```diff
@@ -72,7 +72,6 @@ list(REMOVE_ITEM ${PROJECT_NAME}__SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/./best_sc
 list(REMOVE_ITEM ${PROJECT_NAME}__SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/./stats_submitter.cpp")
 list(REMOVE_ITEM ${PROJECT_NAME}__SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/ai/monsters/rats/ai_rat_fsm.cpp")
 list(REMOVE_ITEM ${PROJECT_NAME}__SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/ui/CExtraContentFilter.cpp")
-list(REMOVE_ITEM ${PROJECT_NAME}__SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/ui/UIDebugFonts.cpp")
```
Xottab-DUTY: Are you sure about this?

src/xrCDB/ISpatial.cpp
```diff
@@ -272,6 +272,8 @@ void ISpatial_DB::_insert(ISpatial_NODE* N, Fvector& n_C, float n_R)
 void ISpatial_DB::insert(ISpatial* S)
 {
     pcs->Enter();
+#if 0 // disabled because linking issue due to IGameObject belongs to xrEngine and is not
```
Xottab-DUTY: You just disabled it even on Windows. Please, don't do this.

src/Layers/xrRenderPC_GL/rgl.cpp
```diff
@@ -449,7 +449,7 @@ void CRender::OnFrame()
 void CRender::OnFrame()
 {
     Models->DeleteQueue();
-    if (ps_r2_ls_flags.test(R2FLAG_EXP_MT_CALC))
+    if (ps_r2_ls_flags.test(R2FLAG_EXP_MT_CALC) && Details)
```
Xottab-DUTY: You just disabled HOM. =(

Please, just remove this, this should be fixed in other way.
`Details` can't be nullptr here. If it is, then error is in another place.

Xottab-DUTY (changes requested):

src/xrGame/CMakeLists.txt
drug007: Well, I'm not sure because it was once. Probable it was local problem...

PR #265, opened 2018-11-01
CAI_Space improvement

FreeZoneMods:
- Get rid of raw pointers in CAI_Space, use smart pointers instead
- Make instance of CAI_Space static object (instead of maintaining global raw pointer to it)
- Add CEventNotifier class to xrCore
- Add events for script engine restarting in CAI_Space
- Get rid of hack in CALifeSimulator constructor
- Add wrapper classes xr_unique_ptr and xr_shared_ptr (will be useful in case of returning xr_new)
- Replace usage of std::unique_ptr to xr_unique_ptr

### DIFF
--- a/res/gamedata/shaders/gl/accum_emissivel.ps
+++ b/res/gamedata/shaders/gl/accum_emissivel.ps
@@ -1,0 +1,8 @@
+#include "common.h"
+#include "iostructs\p_dumb.h"
+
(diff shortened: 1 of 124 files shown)


Xottab-DUTY (changes requested):
Seems good and cool! But xrSE_Factory is broken, please fix it.

PR #268, opened 2018-11-04
Linux port

drug007:
Contains the last changes from @eagleivg and @Zegeri that belong to different branches in different repos and wasn't merged into `xd_dev`. After merging it into `xd_dev` we will be able to use simple and clean way to work with the repo branching only from OpenXRay/xd_dev and not any other repos and branches.


Xottab-DUTY:
Merged #269 instead.

PR #269, opened 2018-11-06
Latest Linux fixes

eagleivg:
Im reopen this PR instead @q4a


Externals/ode/include/ode/objects.h
```diff
@@ -26,6 +26,7 @@
 #include <ode/common.h>
 #include <ode/mass.h>
 #include <ode/contact.h>
+#include <ode/objects.h>
```
Xottab-DUTY: Won't this a recursive include in some cases?

src/Layers/xrRender/HWCaps.cpp
```diff
@@ -238,7 +238,7 @@ void CHWCaps::Update()
     // ***************** PIXEL processing
     raster_major = 4;
     raster_minor = 0;
-    raster.dwStages = 16;
+    raster.dwStages = 15;
```
Xottab-DUTY:
```suggestion
    // XXX: review this
    raster.dwStages = 15; // Previuos value is 16, but it's out of bounds
```

src/xrEngine/xr_object_list.cpp
```diff
@@ -492,66 +487,62 @@ IGameObject* CObjectList::Create(LPCSTR name)
...
-            Objects::const_iterator i = m_crows[1].begin();
-            Objects::const_iterator const e = m_crows[1].end();
-            for (u32 j = 0; i != e; ++i, ++j)
-                Msg("%d %s", j, (*i)->cName().c_str());
-            VERIFY(Device.Paused() || m_crows[1].empty());
-            m_crows[1].clear();
+            Msg("assertion !m_other_crows.empty() failed: %d", m_secondary_crows.size());
+
+            u32 j = 0;
+            for (auto iter: m_secondary_crows)
```
Xottab-DUTY: When replacing iterators with range-based for consider to use references. Otherwise `auto` will copy value and we don't wan't that, because this can lead to incorrect behavior in worst case.
```suggestion
            for (auto& iter : m_secondary_crows)
```

src/xrGame/MainMenu.cpp
```diff
@@ -895,7 +897,7 @@ LPCSTR CMainMenu::GetPlayerName()
 
 LPCSTR CMainMenu::GetCDKeyFromRegistry()
 {
-    string512 key;
+    string512 key = { 0 };
```
Xottab-DUTY, line 900: Do we need this?

src/xrGame/player_hud.cpp
```diff
@@ -664,6 +664,7 @@ void player_hud::update_inertion(Fmatrix& trans)
 
         // clamp by PI_DIV_2
         Fvector last;
+        last.set(0,0,0);
```
Xottab-DUTY: This is unnecessary, since we're assigning `last` to `st_last_dir` below (in `normalize_safe` method)

src/xrSound/OpenALDeviceList.h
```diff
@@ -9,7 +9,7 @@ constexpr pcstr AL_GENERIC_SOFTWARE = "Generic Software";
 
 struct ALDeviceDesc
 {
-    string256 name;
+    string256 name = { 0 };
```
Xottab-DUTY, line 12: Do we need this?

src/xrGame/ui/UIMainIngameWnd.cpp
```diff
@@ -235,11 +235,10 @@ void CUIMainIngameWnd::Init()
...
         AttachChild(m_quick_slots_icons.back());
-        string32 path;
-        xr_sprintf(path, "quick_slot%d", i);
-        CUIXmlInit::InitStatic(uiXml, path, 0, m_quick_slots_icons.back());
-        xr_sprintf(path, "%s:counter", path);
-        UIHelper::CreateStatic(uiXml, path, m_quick_slots_icons.back());
+        std::string path = "quick_slot" + std::to_string(i);
+        CUIXmlInit::InitStatic(uiXml, path.c_str(), 0, m_quick_slots_icons.back());
+        path += ":counter";
+        UIHelper::CreateStatic(uiXml, path.c_str(), m_quick_slots_icons.back());
```
Xottab-DUTY: Make it simplier, just use `xr_strcat` 😉 
```cpp
string32 path;
xr_sprintf(path, "quick_slot%d", i);
CUIXmlInit::InitStatic(uiXml, path, 0, m_quick_slots_icons.back());
xr_strcat(path, ":counter");
UIHelper::CreateStatic(uiXml, path, m_quick_slots_icons.back());
```

src/xrGame/xrGame.vcxproj.filters
```diff
@@ -9703,10 +9703,10 @@
     <ClCompile Include="IKLimbsController.cpp">
       <Filter>Physics\IK</Filter>
     </ClCompile>
-    <ClCompile Include="Ik\jtlimits.cxx">
+    <ClCompile Include="Ik\jtlimits.cpp">
       <Filter>Physics\IK</Filter>
     </ClCompile>
-    <ClCompile Include="Ik\limb.cxx">
+    <ClCompile Include="Ik\limb.cpp">
```
Xottab-DUTY: Since ab75afec2acf3c6be371c56ad3dde7af4e38b964 you can rename the files back.

Xottab-DUTY (changes requested):

src/Layers/xrRender/HWCaps.cpp
drug007: I believe it should be `0`, not `15`.

Externals/ode/include/ode/objects.h
eagleivg: Tested on Release and Mixed, problem do not cause.

src/xrSound/OpenALDeviceList.h
eagleivg: Yes, memcheck warn about branch dependent about this buffer content. Initialize it will be more safely

src/xrGame/MainMenu.cpp
eagleivg: Yes, same reason about  condition dependent from this unitialized buffer content.

src/Layers/xrRender/HWCaps.cpp
eagleivg: It is number of anti-aliasing stages, from 0 to 15, an i believe it just numeration mistake.

src/xrEngine/xr_object_list.cpp
drug007: Here `iter` is just a pointer to `IGameObject`, no need to use reference for pointer. Value copying is intended here. No incorrect behavior at all and it is more efficient.

src/Layers/xrRender/HWCaps.cpp
Xottab-DUTY: This is a number of anisotropic filtering (not anti-aliasing) and this value is used only for for DX9 renderer, for DX10+ it doesn't even used.

For DX10+ is 1 hardcoded stage.

src/xrEngine/xr_object_list.cpp
Xottab-DUTY: Hmm...

src/Layers/xrRender/HWCaps.cpp
drug007: `16` is equal to `0` here. So to preserve the behavior it should be `0`. Changin it to `15` may change the logic.

Xottab-DUTY: This won't change the logic since
> `for DX10+ it doesn't even used.`

This part of the code is DX10+ specific.

Externals/ode/include/ode/objects.h
```diff
@@ -34,13 +35,13 @@ extern "C" {
 /**
  * @defgroup world World
  *
- * The world object is a container for rigid bodies and joints. Objects in 
- * different worlds can not interact, for example rigid bodies from two 
+ * The world object is a container for rigid bodies and joints. Objects in
```
FreeZoneMods: Let's won't make and commit absolutely unnecessary changes. It clutters both review request and the history, makies it harder to explore the real changes and bugs and makes the 'blame' utility unusable. Revert 'style' changes from this review, please.

Xottab-DUTY: Style changes includes replacing tabs with spaces, which is necessary in the current OpenXRay codestyle.

I agree with you and disagree with you at the same time.

FreeZoneMods: Coding style is not a dogma, but just a recomendation. Now 'styling' all the files makes the review unreviewable. Please move the coding style changes to another PR.

Externals/ode/ode/src/joint.cpp
```diff
@@ -49,16 +49,16 @@ const dReal stop_early_reaction			=					M_PI*0.00f;
...
-	{
-		param-=min_err;
-		if(param<REAL(0.))param=REAL(0.);
-	}
-	else
-	{
-		param+=min_err;
-		if(param>REAL(0.))param=REAL(0.);
-	}
+    if(param>REAL(0.))
```
FreeZoneMods: And what about the coding style and spaces?)

Externals/ode/ode/src/joint.cpp
```diff
@@ -474,9 +474,9 @@ int dxJointLimitMotor::testRotationalLimit (dReal angle)
     limit = 1;
 #ifdef USE_STOPS_MIN_ERR
     limit_err = angle - lostop+min_stop_err;
-	if(limit_err>REAL(0.))limit_err=REAL(0.);
+    if(limit_err>REAL(0.))limit_err=REAL(0.);
```
FreeZoneMods: Keep the coding style, please

src/Common/PlatformLinux.inl
```diff
@@ -68,68 +68,57 @@ inline char* _strupr_l(char* str, locale_t loc)
...
-        const char* path,  // Path Input
-        char* drive,       // Drive     : Output
-        char* dir,         // Directory : Output
-        char* fname,       // Filename  : Output
-        char* ext          // Extension : Output
-){
-    char tmp[PATH_MAX] = {0};
-    if(!realpath(path, tmp))
-        strcpy(tmp, path);
+inline void _splitpath(const char* path, // Path Input
```
FreeZoneMods, line 80: The function seems to be unsafe - there is no length checking

src/Common/PlatformLinux.inl
```diff
@@ -287,15 +276,127 @@ typedef dirent DirEntryType;
 #define lstrcpy strcpy
 #define stricmp strcasecmp
 #define strupr SDL_strupr
-inline bool strncpy_s(char * dest, size_t, const char * source, size_t num) {
-    return NULL == strncpy(dest, source, num);
+
+inline int strcpy_s(char *dest, size_t num, const char *source)
+{
+    if(!num)
+        return 22;
```
FreeZoneMods: What is '22'? Please get rid of such magic values. And follow coding style, please.

src/Common/PlatformLinux.inl
```diff
@@ -333,7 +434,12 @@ inline int _filelength(int fd)
 #define _read read
 #define _set_new_handler std::set_new_handler
 #define _finite isfinite
-#define _mkdir(dir) mkdir(dir, S_IRWXU)
+inline int _mkdir(const char *dir)
+{
+    while (char* sep = strchr((char *)dir, '\\')) *sep = '/';
```
FreeZoneMods: Use convert_path_separators

src/Layers/xrRender/Shader.cpp
```diff
@@ -117,12 +117,12 @@ BOOL Shader::equal(Shader* S, int index)
...
         return FALSE;
 
-    return (E[index] == S->E[index]);
+    return E[index]->equal(*S->E[index]);
 }
 
 BOOL Shader::equal(Shader* S)
 {
-    for (int i = 0; i < 4; i++)
+    for (int i = 0; i < 5; i++)
```
FreeZoneMods, line 125: What is '5'? Get rid of magic value

src/Layers/xrRenderGL/glSH_Texture.cpp
```diff
@@ -27,8 +27,8 @@ void resptrcode_texture::create(LPCSTR _name)
 //////////////////////////////////////////////////////////////////////
 CTexture::CTexture()
 {
-    pSurface = NULL;
-    pBuffer = NULL;
+    pSurface = 0;
```
FreeZoneMods, line 30: nullptr? Or it is not a pointer? So it's a strange naming. And I suggest you to use initializer list (or even initialize such members directly in the class definition)

src/Layers/xrRenderGL/glSH_Texture.cpp
```diff
@@ -161,6 +161,7 @@ void CTexture::Load()
 
     flags.bUser = false;
     flags.MemoryUsage = 0;
+    if (0 == *cName) return;
```
FreeZoneMods: nullptr?

src/Layers/xrRenderPC_GL/rgl_shaders.cpp
```diff
@@ -42,9 +42,6 @@ static inline void load_includes(LPCSTR pSrcData, UINT SrcDataLen, xr_vector<cha
 #if defined(WINDOWS)
```
FreeZoneMods: Looks ugly.

src/utils/xrQSlim/src/MxPropSlim.cpp
```diff
@@ -342,7 +342,7 @@ unsigned int MxPropSlim::check_local_validity(unsigned int v, const float* vnew)
             mxv_cross3(n, f_n, d_yx); // n = ((y-x)^(v-x))^(y-x)
             mxv_unitize(n, 3);
 
-            // assert( mxv_dot(d_vx, n, 3) > -FEQ_EPS );
+            // R_ASSERT( mxv_dot(d_vx, n, 3) > -FEQ_EPS );
```
FreeZoneMods, line 345: Commented assertion? Maybe remove it at all?

src/xrCore/LocatorAPI.cpp
```diff
@@ -621,6 +618,7 @@ bool ignore_name(const char* _name)
 
 bool ignore_path(const char* _path)
 {
+    convert_path_separators((char *)_path);
```
FreeZoneMods: WTF??? You MUST NEVER do such ugly things like discarding consts! Even if you are going to restore original string there is a chance to get a static string in read-only memory as a parameter. Get rid of this shit, please.

src/xrCore/ModuleLookup.cpp
```diff
@@ -37,7 +37,7 @@ void* ModuleHandle::Open(pcstr moduleName)
 #ifdef WINDOWS
```
FreeZoneMods, line 37: Get rid of this ugly ifdef, please - create and call platform-dependent functions, please.

src/xrCore/xrCore.cpp
```diff
@@ -237,6 +241,20 @@ void xrCore::Initialize(pcstr _ApplicationName, pcstr commandLine, LogCallback c
 
         DWORD sz_comp = sizeof(CompName);
         GetComputerName(CompName, &sz_comp);
+#elif defined(LINUX)
```
FreeZoneMods, line 244: The whole file consists from ifdefs only...

src/xrGame/ui/UIMainIngameWnd.cpp
FreeZoneMods: xr_strcat is a C-style way, not C++. Also, using xr_strcat is less safe way. But you really need to correct the code - use xr_string instead of std::string.

src/utils/xrQSlim/src/MxPropSlim.cpp
eagleivg: No, they mark potential problem places and can be useful when debugging.

src/xrGame/player_hud.cpp
eagleivg: This set is necessary due some compiler optimizations - member of Fvector really stay unititialized

Externals/ode/ode/src/joint.cpp
Xottab-DUTY: This is ODE with it's own codestyle..

src/Common/PlatformLinux.inl
Xottab-DUTY: Some special return code?

src/xrCore/ModuleLookup.cpp
Xottab-DUTY: We have decided that it's better to split cpp to Windows and Linux versions (`ModuleLookupWindows.cpp` and `ModuleLookupLinux.cpp`) but that's not related to this PR and should be done later.

src/xrGame/Level.cpp
```diff
@@ -440,7 +440,10 @@ void CLevel::OnFrame()
     if (!GEnv.isDedicatedServer)
     {
         if (g_mt_config.test(mtMap))
+        {
+            R_ASSERT(m_map_manager);
```
Xottab-DUTY, line 444: We don't need to check for `m_map_manager`, but we should fix it's initialization on Linux 😉

src/xrGame/Level.cpp
```diff
@@ -549,6 +552,7 @@ void CLevel::OnFrame()
     {
         if (g_mt_config.test(mtLevelSounds))
         {
+            R_ASSERT(m_level_sound_manager);
```
Xottab-DUTY, line 555: We don't need to check for `m_level_sound_manager`, but we should fix it's initialization on Linux 😉

src/xrCore/FS.cpp
```diff
@@ -568,7 +569,10 @@ CVirtualFileReader::CVirtualFileReader(const char* cFileName)
 
     data = (char*)MapViewOfFile(hSrcMap, FILE_MAP_READ, 0, 0, 0);
 #elif defined(LINUX)
-    hSrcFile = ::open(cFileName, O_RDONLY | O_CREAT, S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH); //за такое использование указателя нужно убивать, но пока пусть будет
+    pstr conv_path = xr_strdup(cFileName);
+    convert_path_separators(conv_path);
+    hSrcFile = ::open(conv_path, O_RDONLY | O_CREAT, S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH); //за такое использование указателя нужно убивать, но пока пусть будет
```
FreeZoneMods, line 575: Let's won't duplicate such great pieces of code and move it to separate function. And write comments in English, please :)

src/Layers/xrRenderGL/glSH_Texture.cpp
Xottab-DUTY: pSurface is a pointer for DirectX renderer and GLuint for OpenGL renderer. We can't assign nullptr to GLuint; Also `NULL` can't be used for GLuint if we are compiling with GCC. So, we decided to use straight `0`.

src/Layers/xrRenderGL/glSH_Texture.cpp
Xottab-DUTY:
```suggestion
    if (nullptr == *cName)
        return;
```

Externals/cximage/ximaenc.cpp
```diff
@@ -996,7 +996,7 @@ bool CxImage::CheckFormat(CxFile * hFile, DWORD imagetype)
 ////////////////////////////////////////////////////////////////////////////////
 bool CxImage::CheckFormat(BYTE * buffer, DWORD size, DWORD imagetype)
 {
-	if (buffer==NULL || size==NULL){
+	if (buffer==NULL || size==0){
```
FreeZoneMods, line 999: nullptr, please

src/Layers/xrRenderPC_GL/rgl_shaders.cpp
Xottab-DUTY: Looks strange.. Why?

src/Common/PlatformLinux.inl
```diff
@@ -287,15 +276,127 @@ typedef dirent DirEntryType;
 #define lstrcpy strcpy
 #define stricmp strcasecmp
 #define strupr SDL_strupr
-inline bool strncpy_s(char * dest, size_t, const char * source, size_t num) {
-    return NULL == strncpy(dest, source, num);
+// error code numbers from original MS strcpy_s return value
+inline int strcpy_s(char *dest, size_t num, const char *source)
+{
+    if(!num)
+        return EINVAL;
```
FreeZoneMods: Maybe ERANGE? And looks like you should check dest and src for NULL first (to return EINVAL).

src/Common/PlatformLinux.inl
```diff
@@ -287,15 +276,127 @@ typedef dirent DirEntryType;
...
+            return 0;
+    }
+    dest[0] = '\0';
+    return ERANGE;
+}
+
+template <std::size_t num>
+inline int strcpy_s(char (&dest)[num], const char *source) { return strcpy_s(dest, num, source); }
+
+inline int strncpy_s(char * dest, size_t dst_size, const char * source, size_t num)
```
FreeZoneMods, line 313: And what about returning ERANGE?

src/Common/PlatformLinux.inl
FreeZoneMods: And DON'T modify the CONST argument!

src/Common/PlatformLinux.inl
```diff
@@ -977,3 +1083,16 @@ typedef void *HIC;
...
+inline void convert_path_separators(char * path)
+{
+    while (char* sep = strchr(path, '\\')) *sep = '/';
+}
+
+/** For backward compability of FS, for real filesystem delimiter set to back
+ * @brief restore_path_separators
+ * @param path
+ */
+inline void restore_path_separators(char * path)
```
FreeZoneMods, line 1095: I think there is no real necessity in this function if you won't modify 'const' strings

src/xrCore/fastdelegate.h
```diff
@@ -52,6 +52,7 @@
 #endif // _MSC_VER > 1000
 
 #include <memory.h> // to allow <,> comparisons
+#include <cassert>
```
FreeZoneMods: Please remove it from here

src/xrGame/player_hud.cpp
FreeZoneMods: Please write a note about potential compiler optimization to the comment - otherwise, the line could be unintentionally removed later

Xottab-DUTY: It's better to delete this and use another solution
```suggestion
```
Initialization happens one line below:
```cpp
last.normalize_safe(st_last_dir);
```
But GCC removes this line because `st_last_dir` is static:
```cpp
static Fvector st_last_dir = {0, 0, 0};
```

Probably we can make it volatile and GCC won't remove it:
```cpp
volatile Fvector st_last_dir = {0, 0, 0};
```
`st_last_dir` is used three times in the function.

What do you think guys?

Externals/cximage/ximaenc.cpp
Xottab-DUTY: `size` is not a pointer.

And changing `buffer==NULL` to `buffer==nullptr` is unneeded, this is a third-party library.

Xottab-DUTY: Everything is correct here, because `size` is not a pointer.

And changing `buffer==NULL` to `buffer==nullptr` is unneeded, this is a third-party library.

src/Layers/xrRenderGL/glSH_Texture.cpp
drug007: It would be better to define in renders a correspondence constant and use it to init. But considering quality of the engine I wouldn't touch it at all and let it be as it is.

src/xrCore/FS.cpp
Xottab-DUTY: Okay, we can do this later.

src/Layers/xrRenderGL/glSH_Texture.cpp
Xottab-DUTY: Yes, we'll leave it as TODO.

src/Common/PlatformLinux.inl
eagleivg: Look in code, it need for convert real unix pathes to engine-used

src/xrGame/player_hud.cpp
eagleivg: I think its bad idea:
player_hud.cpp:663:42: error: binding reference of type 'const Self& {aka const _vector3<float>&}' to 'volatile Fvector {aka volatile _vector3<float>}' discards qualifiers
         diff_dir.sub(xform.k, st_last_dir);
                                          ^

PR #277, opened 2018-11-15
Ingame localization switch

qweasdd136963:
Change language in the game.
Fixed two bugs.
Cleanup UIScriptWnd_script.


src/xrGame/xrgame_dll_detach.cpp
```diff
@@ -61,8 +61,7 @@ extern CUIXml* pWpnScopeXml;
 
 void clean_game_globals()
 {
-    // destroy object factory
-    xr_delete(g_object_factory);
+    ai().DestroyScriptEngine();
```
Xottab-DUTY: @FreeZoneMods, so, if we are deleting `g_object_factory`, then renderer scripts (`*.s` files) are trying to call destroyed factory...

src/xrGame/Level_Bullet_Manager.cpp
```diff
@@ -210,7 +212,7 @@ void CBulletManager::AddBullet(const Fvector& position, const Fvector& direction
 void CBulletManager::UpdateWorkload()
 {
 #ifdef DEBUG
-    if (g_mt_config.test(mtBullets))
+    if (!g_mt_config.test(mtBullets))
```
FreeZoneMods: Suggest you to move the conditions under the VERIFY macro too.

src/xrGame/string_table.cpp
```diff
@@ -8,6 +8,9 @@ CStringTable& StringTable() { return *((CStringTable*)gStringTable); }
 STRING_TABLE_DATA* CStringTable::pData = NULL;
 BOOL CStringTable::m_bWriteErrorsToLog = FALSE;
 
+u32 gLanguage = -1;
```
FreeZoneMods: Please don't assign negative values to unsigned values. It could lead to unexpected behaviour. Consider using std::numeric_limits traits instead.

src/xrGame/string_table.cpp
```diff
@@ -8,6 +8,9 @@ CStringTable& StringTable() { return *((CStringTable*)gStringTable); }
...
+xr_vector<xr_token> gLanguagesToken;
```
FreeZoneMods: Maybe g_LanguagesToken? Globla variables are usually uses such naming. And do we really need to do this object global? Maybe use at least 'static' with getter?

src/xrGame/string_table.cpp
```diff
@@ -30,8 +33,41 @@ void CStringTable::Init()
...
+            {
+                gLanguage = it.id;
+                found = true;
+                break;
+            }
+        }
+
+        if (!found)
+            R_ASSERT2(false, "Check localization.ltx");
+    }
```
FreeZoneMods, line 96: I think that lines 36-70 are definitely look like a reason for creating a separate class for incapsulation the logic of managing locales.

src/xrGame/string_table.h
```diff
@@ -38,4 +40,7 @@ class CStringTable : public IStringTable
     static STRING_TABLE_DATA* pData;
 };
 
+extern u32 gLanguage;
+extern xr_vector<xr_token> gLanguagesToken;
```
FreeZoneMods: 'Extern' is not a good pattern. I suggest you to consider another ways, please.

src/xrGame/string_table.cpp
Xottab-DUTY: Be simplier, just global variable is enough ;)

Xottab-DUTY: With the current code style it should be just `LanguageToken`.

Xottab-DUTY: And `LanguageID`

FreeZoneMods: It's amazing but the current 'coding style' does NOT correlate with naming which is actually used in the engine... Do you prefer to edit just some lines in the 'coding style' or make a 'global' commit which changes naming in a lot of files to less convenient one? Or maybe you prefer forget about renaming the old variables at all (and create hodgepodge in the code)?

FreeZoneMods: "Be simplier, just global variable is enough ;)" - Always code as if the guy who ends up maintaining your code will be a violent psychopath who knows where you live.

Xottab-DUTY: I prefer use current code style for the new code and leave old GSC code as is and rename it step-by-step, time to time. A giant rename commit is a bad idea. We have forks that would be forced to syncronize with this commit (and if they have a custom code, they'll have conflicts)

Xottab-DUTY: > Always code as if the guy who ends up maintaining your code will be a violent psychopath who knows where you live.

trollsad

src/xrGame/Level_Bullet_Manager.cpp
Xottab-DUTY: There's no reason for this. The code will look like that then:
```cpp
VERIFY(!g_mt_config.test(mtBullets) && m_thread_id == GetCurrentThreadId() || g_mt_config.test(mtBullets));
```
With the current code it tests if bullets are threaded and if true then just skips this VERIFY. (one check)
With your suggestion it always will do two checks.

Xottab-DUTY: It's better to leave as is, or use this variant:
```cpp
bool threaded = g_mt_config.test(mtBullets);
VERIFY(threaded || m_thread_id == GetCurrentThreadId());
```

src/xrGame/Level_Bullet_Manager.cpp
```diff
@@ -210,8 +212,8 @@ void CBulletManager::AddBullet(const Fvector& position, const Fvector& direction
 void CBulletManager::UpdateWorkload()
 {
 #ifdef DEBUG
-    if (g_mt_config.test(mtBullets))
-        VERIFY(m_thread_id == GetCurrentThreadId());
+    bool threaded = g_mt_config.test(mtBullets);
```
FreeZoneMods: Why do you need IFDEF here? VERIFY seems to be the macro which will be removed at all in release configuration. Suggest removing IFDEF and removing this local variable.

src/xrGame/MainMenu.cpp
```diff
@@ -89,6 +89,7 @@ CMainMenu::CMainMenu()
     g_btnHint = NULL;
     g_statHint = NULL;
     m_deactivated_frame = 0;
+    languageChanged = false;
```
FreeZoneMods: Use initializer list please (or even set the default value even in the definition of the class - it improves the readability)

src/xrGame/MainMenu.h
```diff
@@ -72,6 +72,8 @@ class CMainMenu : public IMainMenu,
 
     xr_vector<CUIWindow*> m_pp_draw_wnds;
 
+    bool languageChanged;
```
FreeZoneMods, line 75: Class members' names usually start with 'm_' prefix.

src/xrGame/console_commands.cpp
```diff
@@ -202,6 +202,40 @@ class CCC_GameDifficulty : public CCC_Token
     virtual void Info(TInfo& I) { xr_strcpy(I, "game difficulty"); }
 };
 
+class CCC_GameLanguage : public CCC_Token
+{
+public:
+    CCC_GameLanguage(LPCSTR N) : CCC_Token(N, (u32*)&CStringTable::LanguageID, NULL){};
+
+    virtual void Execute(LPCSTR args)
```
FreeZoneMods: override instead of virtual?

src/xrGame/console_commands.cpp
```diff
@@ -202,6 +202,40 @@ class CCC_GameDifficulty : public CCC_Token
     virtual void Info(TInfo& I) { xr_strcpy(I, "game difficulty"); }
 };
 
+class CCC_GameLanguage : public CCC_Token
+{
+public:
+    CCC_GameLanguage(LPCSTR N) : CCC_Token(N, (u32*)&CStringTable::LanguageID, NULL){};
```
FreeZoneMods: nullptr instead of NULL;

src/xrGame/inventory_item.cpp
```diff
@@ -123,6 +123,13 @@ void CInventoryItem::Load(LPCSTR section)
     m_icon_name = READ_IF_EXISTS(pSettings, r_string, section, "icon_name", NULL);
 }
 
+void CInventoryItem::ReloadNames()
+{
+    m_name = StringTable().translate(pSettings->r_string(m_object->cNameSect(), "inv_name"));
```
FreeZoneMods: maybe use string constants instead of such 'magic strings'?

src/xrGame/inventory_item.h
```diff
@@ -95,6 +95,7 @@ class CInventoryItem : public CAttachableItem,
 
 public:
     virtual void Load(LPCSTR section);
+    virtual void ReloadNames();
```
FreeZoneMods: Why virtual? Do you really need it?

src/xrGame/string_table.cpp
```diff
@@ -5,35 +5,36 @@
...
     Init();
 }
 
 void CStringTable::Init()
 {
-    if (NULL != pData)
+    if (pData != nullptr)
         return;
 
     pData = new STRING_TABLE_DATA();
```
FreeZoneMods: xr_unique_ptr seems to be a better solution

src/xrGame/string_table.cpp
```diff
@@ -5,35 +5,36 @@
...
+    FillLanguageToken();
 }
 
 CStringTable::~CStringTable() { Destroy(); }
 void CStringTable::Destroy() { xr_delete(pData); }
 void CStringTable::rescan()
 {
-    if (NULL != pData)
+    if (pData != nullptr)
         return;
```
FreeZoneMods: What about getting rid of the return? And looks like you may write the condition shorter: if (pData)

src/xrGame/string_table.h
```diff
@@ -29,13 +29,21 @@ class CStringTable : public IStringTable
...
+    xr_token* GetLanguagesToken() const;
+    static u32 LanguageID;
+
 private:
     void Load(LPCSTR xml_file);
+    void FillLanguageToken();
+    void SetLanguage();
     static STRING_VALUE ParseLine(LPCSTR str, LPCSTR key, bool bFirst);
     static STRING_TABLE_DATA* pData;
+    static xr_vector<xr_token> languagesToken;
```
FreeZoneMods, line 46: m_languagesToken?

src/xrGame/string_table.cpp
```diff
@@ -52,10 +53,55 @@ void CStringTable::Init()
...
+            pSettings->r_line("Languages", i, &lineName, &lineVal);
+            languagesToken.emplace_back(lineName, i);
+        }
+        languagesToken.emplace_back(nullptr, -1);
+    }
+}
+
+void CStringTable::SetLanguage()
+{
+    if (LanguageID != (u32)-1)
```
FreeZoneMods: Consider using std::numeric_limits, please

src/xrGame/string_table.cpp
```diff
@@ -52,10 +53,55 @@ void CStringTable::Init()
...
+        {
+            if (it.name && it.name == pData->m_sLanguage)
+            {
+                LanguageID = it.id;
+                found = true;
+                break;
+            }
+        }
+
+        if (!found)
```
FreeZoneMods: You don't need 'if' here, place condition to R_ASSERT directly.

src/xrGame/string_table.cpp
```diff
@@ -5,35 +5,36 @@
 
 CStringTable& StringTable() { return *((CStringTable*)gStringTable); }
 
-STRING_TABLE_DATA* CStringTable::pData = NULL;
+STRING_TABLE_DATA* CStringTable::pData = nullptr;
```
FreeZoneMods: xr_unique_ptr?

src/xrGame/string_table.cpp
```diff
@@ -52,10 +53,55 @@ void CStringTable::Init()
...
+    else
+    {
+        pData->m_sLanguage = pSettings->r_string("string_table", "language");
+        bool found = false;
+        for (const auto& it : languagesToken)
+        {
+            if (it.name && it.name == pData->m_sLanguage)
+            {
+                LanguageID = it.id;
+                found = true;
```
FreeZoneMods: Maybe use std::find here?

src/xrGame/string_table.h
```diff
@@ -29,13 +29,21 @@ class CStringTable : public IStringTable
     STRING_VALUE translate(const STRING_ID& str_id) const;
     void rescan();
 
+    void ReloadLanguage();
+
     static BOOL m_bWriteErrorsToLog;
     static void ReparseKeyBindings();
 
+    xr_token* GetLanguagesToken() const;
+    static u32 LanguageID;
```
FreeZoneMods, line 38: m_LanguageID

Xottab-DUTY: No `m_` in the current code style

Xottab-DUTY: image

src/xrGame/MainMenu.h
Xottab-DUTY: It corresponds with current code style.

src/xrGame/string_table.h
qweasdd136963: According to codestyle it should be `LanguageID`.

src/xrGame/string_table.h
Xottab-DUTY: No, it corresponds with the current code style.

src/xrGame/console_commands.cpp
Xottab-DUTY:
```suggestion
    CCC_GameLanguage(pcstr N) : CCC_Token(N, (u32*)&CStringTable::LanguageID, nullptr) {}
```

src/xrGame/console_commands.cpp
Xottab-DUTY:
```suggestion
    void Execute(pcstr args) override
```

src/xrGame/MainMenu.cpp
Xottab-DUTY: Initializer list for what? It's contructor!

Xottab-DUTY (approved):

Xottab-DUTY:
Oops, pressed `Approve` by accident. But anyway, I agree with this accident.

src/xrGame/Level_Bullet_Manager.cpp
```diff
@@ -184,8 +184,10 @@ void CBulletManager::AddBullet(const Fvector& position, const Fvector& direction
     const CCartridge& cartridge, float const air_resistance_factor, bool SendHit, bool AimBullet)
 {
 #ifdef DEBUG
```
FreeZoneMods: Unnecessary #ifdef

src/xrGame/MainMenu.h
FreeZoneMods, line 75: Let's won't consider a coding style as a dogma. Think a little - in the most places of the engine members are started with m_ prefix. Why? Because it makes simpler to read such code! When you look into the methods you won't need to think if your languageChanged is local variable or class member value. You don't need to search declaration of the class. 
Moreover, if you won't follow this naming it could easily lead to problems. When the most part of code uses the m_ prefix, but you don't follow this practice, your code may _mislead_ other programmers and force them to create bugs. So, using prefixes for members it's not just a beautifying the code.
Consider using m_languageChanged as a name please.

src/xrGame/inventory_item.cpp
```diff
@@ -29,6 +29,10 @@
 
 #define ITEM_REMOVE_TIME 30000
 
+constexpr pcstr invNameKey = "inv_name";
```
FreeZoneMods: 1) Constants are usually named using CAPITAL letters (remember about misleading)
2) I'd prefer to use constexpr const pcstr here.

src/xrGame/string_table.cpp
```diff
@@ -5,35 +5,41 @@
...
 }
 
 void CStringTable::Init()
 {
-    if (NULL != pData)
+    if (pData)
         return;
 
-    pData = new STRING_TABLE_DATA();
+    pData.reset(new STRING_TABLE_DATA());
```
FreeZoneMods: Use xr_make_unique, please

src/xrGame/string_table.cpp
```diff
@@ -52,10 +58,47 @@ void CStringTable::Init()
...
+    {
+        pData->m_sLanguage = pSettings->r_string("string_table", "language");
+        auto it = std::find_if(languagesToken.begin(), languagesToken.end(), [](const xr_token& token) {
+            return token.name && token.name == pData->m_sLanguage;
+        });
+
+        if (it != languagesToken.end())
+            LanguageID = (*it).id;
+        else
+            xrDebug::Fatal(DEBUG_INFO, "Check localization.ltx! Current language: %s", pData->m_sLanguage.c_str());
```
FreeZoneMods: Why not just to use R_ASSERT(it != languagesToken.end(), ...) on line 92?

src/xrGame/string_table.h
FreeZoneMods: It makes the code easy to read and support. Consider changing the code style.

src/xrGame/MainMenu.cpp
FreeZoneMods: https://en.cppreference.com/w/cpp/language/initializer_list

src/xrGame/MainMenu.h
Xottab-DUTY: We can change it, but while it isn't changed we should follow the current style.
Visual Studio (or ReSharper?) has a different colours for local, global and member variables. If there's no colours separation – come on, looking for declaration is less than minute!

The only thing I'm strongly against of is mixing `camelCase` and `lower_case`. (yes, i believe that this is a mixing)
Variables should be named like `m_language_changed` either `mLanguageChanged` but not `m_languageChanged`!

Changing the code style should be discussed with all team members. @OpenXRay/xray-contributors

Xottab-DUTY: I think, we should merge this PR with the current code style and create a separate issue so that this PR does not become a burden. It is related to ingame language change feature, not the code style.

src/xrGame/inventory_item.cpp
Xottab-DUTY: 2. We can create special type alias for this case: `using cpcstr = const char* const;`. There's many places where it can be used.

src/xrGame/MainMenu.h
Xottab-DUTY: The issue is created: #278.
Code style discussion should be continued there.
This PR will be merged with the current style.

src/xrGame/inventory_item.cpp
qweasdd136963: @FreeZoneMods 
2. `pcstr` is `const char*`. It will be `constexpr const const char*`. Does that make sense?

Zegeri: 2. `constexpr` implies `const`. I don't see the need for a change here.

PR #283, opened 2018-11-23
Fix OpenGL in Mesa drivers

Zegeri:


src/xrEngine/Device_Initialize.cpp
```diff
@@ -50,8 +50,16 @@ void CRenderDevice::Initialize()
         const Uint32 flags = SDL_WINDOW_BORDERLESS | SDL_WINDOW_HIDDEN |
             SDL_WINDOW_RESIZABLE | SDL_WINDOW_OPENGL;
 
-        m_sdlWnd = SDL_CreateWindow("S.T.A.L.K.E.R.: Call of Pripyat", 0, 0, 640, 480, flags);
+#ifdef USE_OGL
```
Xottab-DUTY: `USE_OGL` is defined and should only be used in `xrRender_GL` library.
In `Device`, it's better to check against against `psDeviceFlags`:

```cpp
if (psDeviceFlags.test(rsRGL))
{
	SDL_GL_SetAttribute(...);
	...
}
```

Xottab-DUTY (changes requested):
Much thanks for the effort!

Xottab-DUTY (approved):

SkyLoaderr (approved):

src/xrEngine/Device_Initialize.cpp
```diff
@@ -50,8 +50,17 @@ void CRenderDevice::Initialize()
         const Uint32 flags = SDL_WINDOW_BORDERLESS | SDL_WINDOW_HIDDEN |
             SDL_WINDOW_RESIZABLE | SDL_WINDOW_OPENGL;
 
-        m_sdlWnd = SDL_CreateWindow("S.T.A.L.K.E.R.: Call of Pripyat", 0, 0, 640, 480, flags);
+        if (psDeviceFlags.test(rsRGL))
```
Zegeri, line 53: Seems like at this point `psDeviceFlags` hasn't yet been initialized, so this test will fail always.

Xottab-DUTY: Oh.. True.. I'll think on it..

Xottab-DUTY: We can try to move initialization stages in `xrEngine/main.cpp`

PR #300, opened 2018-12-16
README

uralets:
Documentation rehash.
What's there now is hopelessly convoluted and outdated.

Mostly build / setup instructions for Linux, but extensible to other platforms.

Notably instructions are now split into different files.
Also does away with mentioning wine (tho it probably still needed if installing from retail release - section about which is unfinished)

### DIFF
--- a/README
+++ b/README
@@ -1,0 +1,43 @@
+WELCOME TO THE XRAY-16 PROJECT!
+===============================
+This README file is for the xray-16 computer game engine.
+
+xray-16 is a development of the engine used in STALKER: Call of Pripyat game
+originally developed by GSC Game World and released in 2009.
+
+It's an effort to get the code into a better shape by fixing bugs and
+refactoring the code.
+It is collectively developed by the OpenXRay community (github.com/OpenXRay).
+It supports GNU/Linux and MS Windows™ on x86/x64 platforms.
+
+LICENSING NOTES
+===============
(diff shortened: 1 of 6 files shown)


README
```diff
@@ -0,0 +1,43 @@
+WELCOME TO THE XRAY-16 PROJECT!
+===============================
+This README file is for the xray-16 computer game engine.
+
+xray-16 is a development of the engine used in STALKER: Call of Pripyat game
```
FreeZoneMods, line 5: Maybe "xray-16 is an evolution" sounds better?

collectioner: Then "OpenXRay" is an evolution.. Since xray-16 is already an CoP's engine.

uralets: I wouldn't mind either way - it's just a 'courtesy' intro after all.

Development and evolution seem to me as pretty much synonymous in this case.

I was thinking of OpenXRay as more of a 'community' and of xray-16 as one of its projects - thus the phrasing.

collectioner: @Xottab-DUTY we need another version of xray-16. Maybe, OpenXRay 1.6 can be.. Aw, forget it, it's another question

uralets:
Committed an updated version.

Clarifications, amends and clean-ups.

Also, note that the README file I've put together IS NOT intended to replace the github's README.md (IMO a somewhat unfortunate choice by GH - clashes with projects' traditional doc filename).
Tho by all means could be used (along with the other files from this PR) to compile one.

PR #309, opened 2018-12-30
Port r2_R_sun and r3_R_rain to Linux

eagleivg:
List of changes:

1. Port r2_R_sun and r3_R_rain to Linux (Use engine internal types and implement matrix conversion func - maybe slower than glm or another library).
2. Fixed buffer overflow on linux due mismath size of types (LPDWORD casting)
3. Fix crash at exit on Linux - forgoten Ansel clean.
4. downgrade requirements for SDL2 (for build on ubuntu 16.04)


src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -417,41 +389,42 @@ Frustum::Frustum()
...
-    //  build a bit-field that will tell us the indices for the nearest and farthest vertices from each plane...
-    for (int i = 0; i < 6; i++)
-        nVertexLUT[i] = (planes[i].x < 0.f ? 1 : 0) | (planes[i].y < 0.f ? 2 : 0) | (planes[i].z < 0.f ? 4 : 0);
-
     for (int i = 0; i < 8; i++) // compute extrema
     {
-        const Fplane& p0 = i & 1 ? camPlanes[4] : camPlanes[5];
-        const Fplane& p1 = i & 2 ? camPlanes[3] : camPlanes[2];
-        const Fplane& p2 = i & 4 ? camPlanes[0] : camPlanes[1];
+        const Fvector4 p0 = i & 1 ? camPlanes[4] : camPlanes[5];
```
FreeZoneMods: const FVector4& (here and below) ?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -479,15 +452,15 @@ const float _eps = 0.000001f;
...
+    xr_vector<Fvector4> planes;
 
     BOOL clip(Fvector3& p0, Fvector3& p1) // returns TRUE if result meaningfull
     {
         float denum;
         Fvector3 D;
         for (int it = 0; it < int(planes.size()); it++)
         {
-            Fplane& P = planes [it];
+            Fvector4 P = planes [it];
```
FreeZoneMods: const FVector4& ?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -730,8 +693,8 @@ void CRender::render_sun()
     set_Recorder(nullptr);
 
     //	Prepare to interact with D3DX code
-    const Fmatrix& m_View = Device.mView;
-    const Fmatrix& m_Projection = ex_project;
+    const Fmatrix m_View = Device.mView;
```
FreeZoneMods, line 580: Why we removing reference here?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -82,16 +82,16 @@ static inline float XRVec3Dot(const Fvector3 *pv1, const Fvector3 *pv2)
     return (pv1->x) * (pv2->x) + (pv1->y) * (pv2->y) + (pv1->z) * (pv2->z);
 }
 
-static inline float XRPlaneDotCoord(const Fplane *pp, const Fvector3 *pv)
+static inline float XRPlaneDotCoord(const Fvector4 *pp, const Fvector3 *pv)
```
FreeZoneMods: What is the profit from changing Fplane to FVector4? Does it improves performance?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -417,41 +389,42 @@ Frustum::Frustum()
...
     Fvector4 planes[6];
-    planes[0] = column4.sub(column1); // left
-    planes[1] = column4.add(column1); // right
-    planes[2] = column4.sub(column2); // bottom
-    planes[3] = column4.add(column2); // top
-    planes[4] = column4.sub(column3); // near
-    planes[5] = column4.add(column3); // far
+    Fvector4 tmp;
+    tmp.set(column4);
+    planes[0] = tmp.sub(column1); // left
```
FreeZoneMods, line 283: Maybe convert magic number '6' to named constant  and create cycle here?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -497,35 +470,26 @@ struct DumbClipper
...
-                    tmp.mul(cls0 / denum);
-                    tmp.invert();
-                    p0.add(tmp);
-//                    p0 += - D * cls0 / denum;
+                    D.invert();
+                    D.mul(cls0 / denum);
+                    p0.add(D);
                 }
             }
             if (cls1 > 0)
```
FreeZoneMods, line 364: A body of this condition is almost same as a body of condition above. Could we rewrite the fragment to avoid code duplication?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -497,35 +470,26 @@ struct DumbClipper
...
                 denum = XRPlaneDotNormal(&P, &D);
                 if (denum != 0)
                 {
-                    Fvector3 tmp;
-                    tmp = D;
-                    tmp.mul(cls0 / denum);
-                    tmp.invert();
-                    p0.add(tmp);
-//                    p0 += - D * cls0 / denum;
+                    D.invert();
```
FreeZoneMods, line 359: You change order of 'mul' and 'invert' here. Are you sure that it always works as expected and there is no side-effects (like problems with rounding)?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
 #define ALMOST_ZERO(F) ((FLT_AS_DW(F) & 0x7f800000L)==0)
 #define IS_SPECIAL(F) ((FLT_AS_DW(F) & 0x7f800000L)==0x7f800000L)
 
+static inline Fmatrix* XRMatrixIdentity(Fmatrix *pout)
+{
+    if ( !pout ) return NULL;
+    pout->m[0][1] = 0.0f;
```
FreeZoneMods: Do we have a method for setting all components of the vector to a certain number? I'm almost sure that it presents somewhere. If no, maybe we should create it?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
 #define ALMOST_ZERO(F) ((FLT_AS_DW(F) & 0x7f800000L)==0)
 #define IS_SPECIAL(F) ((FLT_AS_DW(F) & 0x7f800000L)==0x7f800000L)
 
+static inline Fmatrix* XRMatrixIdentity(Fmatrix *pout)
+{
+    if ( !pout ) return NULL;
```
FreeZoneMods: I think that single return-point in the end of the function looks better, so maybe inverse condition like "if (pout) { ... }; return pout;" ? 
And does the situation (here and in the next functions) when we get nullptr in parameters is normal? If not, maybe add an assertion?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+
+static inline float XRVec3Length(const Fvector3 *pv)
+{
+    if (!pv) return 0.0f;
+    return sqrtf( pv->x * pv->x + pv->y * pv->y + pv->z * pv->z );
+}
+
+Fvector3* XRVec3Normalize(Fvector3 *pout, const Fvector3 *pv)
+{
+    FLOAT norm;
```
FreeZoneMods: Why 'FLOAT'?  Maybe 'float' or even 'auto'? And don't split declaration and initialization, please - we are using not C'89 standart ;)

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+        pout->z = 0.0f;
+    }
+    else
+    {
+        pout->x = pv->x / norm;
+        pout->y = pv->y / norm;
+        pout->z = pv->z / norm;
+    }
+
+    return pout;
```
FreeZoneMods: You are forget to check arguments for nullptr in this function.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+}
+
+Fvector3* XRVec3Normalize(Fvector3 *pout, const Fvector3 *pv)
+{
+    FLOAT norm;
+
+    norm = XRVec3Length(pv);
+    if ( !norm )
+    {
+        pout->x = 0.0f;
```
FreeZoneMods: And once again we see that the method for assigning certain nubler to all components of the vector should be useful.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+    if (!pv) return 0.0f;
+    return sqrtf( pv->x * pv->x + pv->y * pv->y + pv->z * pv->z );
+}
+
+Fvector3* XRVec3Normalize(Fvector3 *pout, const Fvector3 *pv)
+{
+    FLOAT norm;
+
+    norm = XRVec3Length(pv);
+    if ( !norm )
```
FreeZoneMods: We shouldn't compare floating point numbers with exact values without epsilons.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+        pout->y = pv->y / norm;
+        pout->z = pv->z / norm;
+    }
+
+    return pout;
+}
+
+Fvector3* XRVec3TransformNormal(Fvector3 *pout, const Fvector3 *pv, const Fmatrix *pm)
+{
+    const Fvector3 v = *pv;
```
FreeZoneMods: Why do you need to copy it instead of making const reference? Does your processor is too fast and you want to slower it? And check the arguments first, please.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+    pout->x = pm->m[0][0] * v.x + pm->m[1][0] * v.y + pm->m[2][0] * v.z;
+    pout->y = pm->m[0][1] * v.x + pm->m[1][1] * v.y + pm->m[2][1] * v.z;
+    pout->z = pm->m[0][2] * v.x + pm->m[1][2] * v.y + pm->m[2][2] * v.z;
+    return pout;
+
+}
+
+Fmatrix* XRMatrixTranslation(Fmatrix *pout, float x, float y, float z)
+{
+    XRMatrixIdentity(pout);
```
FreeZoneMods: Check an argument, please.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+    pout->m[3][0] = x;
+    pout->m[3][1] = y;
+    pout->m[3][2] = z;
+    return pout;
+}
+
+Fmatrix* XRMatrixOrthoOffCenterLH(Fmatrix *pout, float l, float r, float b, float t, float zn, float zf)
+{
+    XRMatrixIdentity(pout);
+    pout->m[0][0] = 2.0f / (r - l);
```
FreeZoneMods, line 76: Check pout, please. And do we need to check denominators for zeros before performing divisions?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+    pout->m[1][1] = 2.0f / (t - b);
+    pout->m[2][2] = 1.0f / (zf -zn);
```
FreeZoneMods, line 78: Check spaces before variables, please (here and below).

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+    pout->m[3][0] = -1.0f -2.0f *l / (r - l);
+    pout->m[3][1] = 1.0f + 2.0f * t / (b - t);
+    pout->m[3][2] = zn / (zn -zf);
+    return pout;
+}
+
+Fmatrix* XRMatrixInverse(Fmatrix *pout, float *pdeterminant, const Fmatrix *pm)
```
FreeZoneMods: Check arguments, please

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+
+    t[0] = pm->m[1][0] * pm->m[2][1] - pm->m[2][0] * pm->m[1][1];
+    t[1] = pm->m[1][0] * pm->m[3][1] - pm->m[3][0] * pm->m[1][1];
+    t[2] = pm->m[2][0] * pm->m[3][1] - pm->m[3][0] * pm->m[2][1];
+    v[8] = pm->m[3][3] * t[0] - pm->m[2][3] * t[1] + pm->m[1][3] * t[2];
+    v[12] = -pm->m[3][2] * t[0] + pm->m[2][2] * t[1] - pm->m[1][2] * t[2];
+
+    det = pm->m[0][0] * v[0] + pm->m[0][1] * v[4] +
+        pm->m[0][2] * v[8] + pm->m[0][3] * v[12];
+    if (det == 0.0f)
```
FreeZoneMods, line 103: Don't compare floats with an exact value. And nullptr instead of NULL, please.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+
+    v[15] = pm->m[0][0] * (pm->m[1][1] * pm->m[2][2] - pm->m[1][2] * pm->m[2][1]) -
+        pm->m[1][0] * (pm->m[0][1] * pm->m[2][2] - pm->m[0][2] * pm->m[2][1]) +
+        pm->m[2][0] * (pm->m[0][1] * pm->m[1][2] - pm->m[0][2] * pm->m[1][1]);
+
+    det = 1.0f / det;
+
+    for (i = 0; i < 4; i++)
+        for (j = 0; j < 4; j++)
+            pout->m[i][j] = v[4 * i + j] * det;
```
FreeZoneMods, line 150: Maybe change a type of 'v' to a matrix? It should allow us to use standard method for scalar multiplying for matrices here instead of duplicating a code.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+
+    return pout;
+}
+
+Fmatrix* XRMatrixMultiply(Fmatrix *pout, const Fmatrix *pm1, const Fmatrix *pm2)
+{
+    Fmatrix out;
+    int i,j;
```
FreeZoneMods: I suggest to declare this variables directly in the 'for' expressions.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+
+    for (i=0; i<4; i++)
+    {
+        for (j=0; j<4; j++)
+        {
+            out.m[i][j] = pm1->m[i][0] * pm2->m[0][j] + pm1->m[i][1] * pm2->m[1][j] + pm1->m[i][2] * pm2->m[2][j] + pm1->m[i][3] * pm2->m[3][j];
+        }
+    }
+
+    *pout = out;
```
FreeZoneMods: Why do you need this extra copying (here and in the functions below)? Is it an optimization for the case when an output is stored in the heap? I think we need to check if it really improves a performance. And check input pointers before. please.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+    return pout;
+}
+
+Fvector3* XRVec3TransformCoord(Fvector3 *pout, const Fvector3 *pv, const Fmatrix *pm)
+{
+    Fvector3 out;
+    FLOAT norm;
```
FreeZoneMods: Maybe 'float' instead of 'FLOAT'? And initialize in the single line with a declaration, please.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+
+    norm = pm->m[0][3] * pv->x + pm->m[1][3] * pv->y + pm->m[2][3] *pv->z + pm->m[3][3];
+
+    out.x = (pm->m[0][0] * pv->x + pm->m[1][0] * pv->y + pm->m[2][0] * pv->z + pm->m[3][0]) / norm;
+    out.y = (pm->m[0][1] * pv->x + pm->m[1][1] * pv->y + pm->m[2][1] * pv->z + pm->m[3][1]) / norm;
+    out.z = (pm->m[0][2] * pv->x + pm->m[1][2] * pv->y + pm->m[2][2] * pv->z + pm->m[3][2]) / norm;
+
+    *pout = out;
```
FreeZoneMods: Check an arguments, please.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+
+    return pout;
+}
+
+Fvector3* XRVec3TransformCoordArray(Fvector3* out, uint outstride, const Fvector3* in, uint instride, const Fmatrix* matrix, uint elements)
+{
+    uint i;
```
FreeZoneMods: Check arguments and declare counter in the 'for' expression, please.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
+
 //////////////////////////////////////////////////////////////////////////
 struct Frustum
 {
     Frustum();
-    Frustum(const D3DXMATRIX* matrix);
+    Frustum(const Fmatrix* matrix);
 
-    D3DXPLANE camPlanes [6];
+    Fplane camPlanes [6];
```
FreeZoneMods: Maybe get rid ot these magic constants?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
 {
-    D3DXVECTOR3 minPt;
-    D3DXVECTOR3 maxPt;
+    Fvector3 minPt;
+    Fvector3 maxPt;
 
-    BoundingBox(): minPt(1e33f, 1e33f, 1e33f), maxPt(-1e33f, -1e33f, -1e33f) { }
+    BoundingBox()
+    {
+        minPt.set(1e33f, 1e33f, 1e33f);
```
FreeZoneMods, line 171: Why you are refusing to use initializer list in the constructors (here and below)?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -40,51 +36,300 @@ static int facetable[6][4] = {
...
         {
             Merge(&(*boxes)[i].maxPt);
             Merge(&(*boxes)[i].minPt);
         }
     }
 
-    void Centroid(D3DXVECTOR3* vec) const { *vec = 0.5f * (minPt + maxPt); }
+    void Centroid(Fvector3* vec) const
+    {
+        Fvector3 tmp = *vec;
```
FreeZoneMods: Looks like you are writing the same code using much more lines and additional copying. Why? Does it improves a performance?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -94,14 +339,15 @@ struct BoundingBox
         maxPt.z = _max(maxPt.z, vec->z);
     }
 
-    D3DXVECTOR3 Point(int i) const
+    Fvector3 Point(int i) const
     {
-        return D3DXVECTOR3(i & 1 ? minPt.x : maxPt.x, i & 2 ? minPt.y : maxPt.y, i & 4 ? minPt.z : maxPt.z);
+        return {i & 1 ? minPt.x : maxPt.x, i & 2 ? minPt.y : maxPt.y, i & 4 ? minPt.z : maxPt.z};
```
FreeZoneMods, line 223: Please don't return this monster, it's quite hard to read. It's better to make a temoporary variable here.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -94,14 +339,15 @@ struct BoundingBox
...
     }
 };
 
+
```
FreeZoneMods: Unnecessary change, please revert.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -94,14 +339,15 @@ struct BoundingBox
...
 ///////////////////////////////////////////////////////////////////////////
-BOOL LineIntersection2D(D3DXVECTOR2* result, const D3DXVECTOR2* lineA, const D3DXVECTOR2* lineB)
+BOOL LineIntersection2D(Fvector2* result, const Fvector2* lineA, const Fvector2* lineB)
```
FreeZoneMods: Check arguments?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -118,64 +364,71 @@ BOOL LineIntersection2D(D3DXVECTOR2* result, const D3DXVECTOR2* lineA, const D3D
...
+    result = n1_n2;
+    result.add(n2_n0);
+    result.add(n0_n1);
+    result.mul(secTheta);
+    result.invert ();
+    *intersectPt = result;
     return TRUE;
 }
 
+
```
FreeZoneMods, line 264: Extra line

src/xrGame/GamePersistent.cpp
```diff
@@ -238,7 +238,9 @@ void CGamePersistent::OnAppEnd()
 
     GMLib.Unload();
 
+#ifdef WINDOWS
```
FreeZoneMods, line 241: Looks suspicious.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -94,111 +335,96 @@ struct BoundingBox
...
 {
-    D3DXVECTOR3 n0(p0->a, p0->b, p0->c);
-    D3DXVECTOR3 n1(p1->a, p1->b, p1->c);
-    D3DXVECTOR3 n2(p2->a, p2->b, p2->c);
-
-    D3DXVECTOR3 n1_n2, n2_n0, n0_n1;
+    Fvector3 n0 = {p0->x, p0->y, p0->z};
+    Fvector3 n1 = {p1->x, p1->y, p1->z};
+    Fvector3 n2 = {p2->x, p2->y, p2->z};
+
```
FreeZoneMods, line 237: Extra tab

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
eagleivg: and lost comments for side of each plane?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
eagleivg: Because D3DXPLANE - 4-d vector, unlike the Fplane (content normale 3-d vector and d-coefficent)

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
eagleivg: no, this is mirror difference, first case D=p1-p0, second D=p0-p1 and all subsequent operations same. This be look ugly and more complicated.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
eagleivg: Yes, im sure, because this operation (vector addition and inverse) don`t change result in different order of execution.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
eagleivg: replaced on internal identify() method of Fmatrix

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
eagleivg: fixed

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
eagleivg: This if standart six planes for frustrum culling, im not sure than need write this...

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
eagleivg: Because internal struct Fvector type don`t have constructor, only set() method

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
eagleivg: Because  internal struct Fvector type don`t have defined operators, only methods used below.

src/xrGame/GamePersistent.cpp
eagleivg: Style from code of constructor above. Ansel don`t implemented on linux, that is normal, as i think.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
FreeZoneMods: So, maybe we should implement such operators? The current code is quite hard to read.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -36,235 +36,120 @@ static int facetable[6][4] = {
...
-static inline float XRVec3Length(const Fvector3 *pv)
-{
-    if (!pv) return 0.0f;
-    return sqrtf( pv->x * pv->x + pv->y * pv->y + pv->z * pv->z );
-}
-
-Fvector3* XRVec3Normalize(Fvector3 *pout, const Fvector3 *pv)
-{
-    float norm;
+    uint i;
```
FreeZoneMods: Consider declaration of the variable directly in for-statement.

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -36,235 +36,120 @@ static int facetable[6][4] = {
...
 {
-    if ( !pv1 || !pv2 ) return 0.0f;
-    return (pv1->x) * (pv2->x) + (pv1->y) * (pv2->y) + (pv1->z) * (pv2->z);
+    pout->x = pm.m[0][0] * pv.x + pm.m[1][0] * pv.y + pm.m[2][0] * pv.z;
+    pout->y = pm.m[0][1] * pv.x + pm.m[1][1] * pv.y + pm.m[2][1] * pv.z;
+    pout->z = pm.m[0][2] * pv.x + pm.m[1][2] * pv.y + pm.m[2][2] * pv.z;
 }
 
-static inline float XRPlaneDotCoord(const Fvector4 *pp, const Fvector3 *pv)
+void XRVec3TransformCoordArray(Fvector3* out, const Fvector3* in, const Fmatrix& matrix, uint elements)
```
FreeZoneMods: Why 'in' is pointer, not reference? And maybe check arguments for null?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
```diff
@@ -345,7 +345,7 @@ struct DumbClipper
         Fvector3 D;
         for (int it = 0; it < int(planes.size()); it++)
         {
-            Fvector4 P = planes [it];
+            const Fvector4 P = planes [it];
```
FreeZoneMods, line 346: Do you really want to copy the vector? Or maybe make a reference instead of copying?

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
eagleivg: This is pointer to array of pointers to Fvector3

src/Layers/xrRenderPC_GL/r2_R_sun.cpp
FreeZoneMods: You can write something like:
```
    const int PLANES_NUMBER = 6;
    Fvector4 planes[PLANES_NUMBER];
    //...
        
        
    const std::array<std::pair<Fvector4&, float>, PLANES_NUMBER> actions = 
    {{ 
        {column1, -1.0}, //left
        {column1,  1.0}, //right
        {column2, -1.0}, //bottom        
        {column2,  1.0}, //top
        {column3, -1.0}, //near
        {column3,  1.0}, //far
    }};
    
    for (int i = 0; i < PLANES_NUMBER; ++i)
    {
        FVector4 tmp = actions[i].second * column4;
        planes[i] = tmp.add(actions[i].first);
    }
}
```

eagleivg:
Big losses in drawing speed, you need to rewrite

PR #312, opened 2019-01-01
Fixed available video modes count printed in console

xrombik:


drug007:
Commit 182db49c0ff3308c3f2128574268d5b99978b088 should be removed from PR.

PR #314, opened 2019-01-01
Using more portable features for further FreeBSD portage.

devnexen:


src/Common/PlatformLinux.inl
```diff
@@ -1087,7 +1089,7 @@ typedef void *HIC;
 #define D3DTSS_TCI_CAMERASPACEREFLECTIONVECTOR    0x30000
 #define D3DTSS_TCI_SPHEREMAP                      0x40000
 
-inline BOOL SwitchToThread() { return (0 == pthread_yield()); }
+inline BOOL SwitchToThread() { return (0 == sched_yield()); }
```
culhatsker, line 1092: Can you explain why you're changing `pthread_` function to `sched_` function?

src/xrCore/cpuid.cpp
```diff
@@ -163,7 +163,7 @@ unsigned int query_processor_info(processor_info* pinfo)
     unsigned int pa_mask_save = 0;
     cpu_set_t my_set;
     CPU_ZERO(&my_set);
-    sched_getaffinity(0, sizeof(cpu_set_t), &my_set);
+    pthread_getaffinity_np(pthread_self(), sizeof(cpu_set_t), &my_set);
```
culhatsker, line 166: And here you do the opposite: `sched_` function is replaced with `pthread_`

src/Common/PlatformLinux.inl
devnexen: sched_yield is more portable, same signature whereas pthread_yield returns void in another platform for example. This PR is to be able to port to a bit more oses/platforms.

src/xrCore/cpuid.cpp
devnexen: sched_getaffinity is a Linux-ism whereas pthread_getaffinity_np exists outside of Linux's libc.

culhatsker: Hm.. This is weird. But let it be.

uralets: sched_getaffinity is Linux-specific as per its man page.
pthread_getaffinity_np (np is for non-portable) is OS-specific as per its respective man page for eg GNU libc and FreeBSD.

drug007:
@devnexen This PR looks like incomplete one, could you elaborate goal you'd like to achieve by this PR? What OSes do you plan to support additional to Windows and Linux?

devnexen:
It s on purpose... just a first step to support freebsd afterwards... plan was first to see no regression for linux

drug007:
I see. Could you change the PR title to reflect this intention?
I don't know how different FreeBSD and Linux are but wouldn't it be better to have another conditional compilation option for FreeBSD? In current form of PR your changes are rather contradictory - `pthread` changes to `sched` and vice versa simultaneously. Probably using conditional compilation for FreeBSD resolve this problem?

devnexen:
Both are more portable as explained by @uralets even though the naming might confuse ☺. In fact the goal is also to decrease a bit the ifdef which will be unavoidable in other places. Agreed to change the title when I can.

PR #315, opened 2019-01-02
Added constructor for _flags struct that sets it to 0

xrombik:


src/xrCore/xr_shortcut.h
```diff
@@ -17,21 +17,22 @@ struct XRCORE_API xr_shortcut
         {
             u8 key;
             Flags8 ext;
-        };
+        } ext_key;
```
Xottab-DUTY, line 20: Why?

xrombik: It seems to be a unnamed union, and adding a constructor provides some compiler errors. It resolves it.

src/xrCore/xr_shortcut.h
```diff
@@ -17,21 +17,22 @@ struct XRCORE_API xr_shortcut
...
         u16 hotkey;
     };
 
-    xr_shortcut(u8 k, bool a, bool c, bool s) : key(k)
```
Xottab-DUTY, line 24: Why have you removed this?
```cpp
 : key(k)
```
It's equal to
```cpp
ext_key.key = k;
```

xrombik: This line `: key(k)` provides error to.

src/xrCore/_flags.h
```diff
@@ -13,6 +13,9 @@ struct _flags
 
     T flags;
 
+    _flags () { flags = T(0); }
+    _flags (T v) { flags = T(v); }
```
Xottab-DUTY: Please, use `static_cast`.

xrombik: Ok. But this goes to intends to add more static casts.

Xottab-DUTY: Why?

xrombik: `/xray-16/src/xrCore/_flags.h:31`
```
    SelfRef zero() noexcept
    {
        flags = T(0);
        return *this;
    }

    SelfRef one() noexcept
    {
        flags = T(~0);
        return *this;
    }
```

Xottab-DUTY: Ah, it should be OK then.

src/xrCore/_flags.h
```diff
@@ -13,17 +13,20 @@ struct _flags
...
     {
-        flags = T(0);
+        flags = static_cast<T>(0);
         return *this;
     }
 
     SelfRef one() noexcept
     {
-        flags = T(-1);
+        flags = static_cast<T>(~0);
```
Xottab-DUTY, line 29:
```suggestion
        flags = static_cast<T>(-1);
```

src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -442,11 +442,7 @@ class CCC_VidMonitor : public CCC_Token
         bEmptyArgsHandled = false;
     }
 
-    void Execute(pcstr args) override
-    {
-        CCC_Token::Execute(args);
-        FillVidModesToken(Vid_SelectedMonitor);
-    }
```
Xottab-DUTY: This is intended, don't remove this please.

xrombik: For what?

Xottab-DUTY: When you are changing monitor you need to update vid mode token. That's why.

src/xrCore/_flags.h
xrombik: 1 is inverted 0, but not -1. -1 it is a signed value.

Xottab-DUTY: Yes, this is intended too.

Xottab-DUTY: Ah.. I understand. But I'm not sure about this..

xrombik:
```
#include <stdio.h>
#include <stdlib.h>
int main()
{
    printf("%lx %x %x %x\n", ~0UL, ~0U, (unsigned short)(~0U), (unsigned char)(~0U));
}
```

xrombik: `~0` to `-1` will not change, because `-1` instead of `~0` is a "hack", so people write when they want to show how smart they are, that they say that in the additional code `-1` these are all ones in binary form. But the readability of the code at the same time falls.

PR #318, opened 2019-01-03
Fix crash and xray allocator

### LINKED ISSUE #322: Crash when NPCs try to open/close doors.
Sometimes you can get an error in the moment when several NPCs one by one pass through the doors.
image

qweasdd136963:
1. Fixed crash on starting the game.
2. Added old xray allocator. Fixed some bugs with deallocation. Also fixes #322.

### DIFF
--- a/src/xrCore/Memory/xalloc.h
+++ b/src/xrCore/Memory/xalloc.h
@@ -1,0 +1,69 @@
+#pragma once
+
+template <typename T>
+class xalloc
+{
+public:
+    using size_type = size_t;
+    using difference_type = std::ptrdiff_t;
+    using pointer = T*;
+    using const_pointer = const T*;
+    using reference = T&;
+    using const_reference = const T&;
+    using value_type = T;
+
+    template <class Other>
+    struct rebind
+    {
+        using other = xalloc<Other>;
+    };
+
+    pointer address(reference ref) const { return &ref; }
+    const_pointer address(const_reference ref) const { return &ref; }
+
+    xalloc() = default;
+    xalloc(const xalloc<T>&) = default;
+
+    template <class Other>
+    xalloc(const xalloc<Other>&)
+    {
+    }
+
+    template <class Other>
+    xalloc& operator=(const xalloc<Other>&)
+    {
+        return *this;
+    }
+
+    pointer allocate(const size_type n, const void* p = nullptr) const { return xr_alloc<T>(n); }
+    void deallocate(pointer p, const size_type /*n*/) const { xr_free(p); }
+    void deallocate(void* p, const size_type /*n*/) const { xr_free(p); }
+
+    template <class U, class... Args>
+    void construct(U* ptr, Args&&... args)
+    {
+        new (ptr) U(std::forward<Args>(args)...);
+    }
+    template <class U>
+    void destroy(U* p)
+    {
+        p->~U();
+    }
+    size_type max_size() const
+    {
+        constexpr auto count = std::numeric_limits<size_type>::max() / sizeof(T);
+        return count > 0 ? count : 1;
+    }
+};
+
+template <class T, class Other>
+bool operator==(const xalloc<T>&, const xalloc<Other>&)
+{
+    return true;
+}
+
+template <class T, class Other>
+bool operator!=(const xalloc<T>&, const xalloc<Other>&)
+{
+    return false;
+}
--- a/src/xrCore/xrMemory.h
+++ b/src/xrCore/xrMemory.h
@@ -1,5 +1,4 @@
 #pragma once
-
 #include "_types.h"
 
 /*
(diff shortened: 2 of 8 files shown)


culhatsker:
Hi, @qweasdd136963, can you explain why we need this commit: https://github.com/OpenXRay/xray-16/pull/318/commits/db0085d6ea301e6ca585a1da83914b7d8ea73fc9?

culhatsker:
Please link this PR to a GitHub issue that this PR resolves. It's better to have context on what problem it solves and why it was happening. If there's no ticket that describes this problem, feel free to create it. This shortest PR description doesn't say much. Let others understand why these changes are necessary.

Please fix Linux build as it seems to be broken by these changes (see Travis build assigned to this PR).

The classes this PR add have a different naming style than others in the engine. Most classes have an `xr` prefix, but those you added have `xray` and `x` prefixes which breaks accustomed conventions.

You created a directory `xrCore/Memory` can you just move these files to `xrCore` directory without adding new one or even move the code to `xrCore/xrMemory.h`?

Xottab-DUTY:
I think we also need to return `xr_aligned_offset_malloc` functions and original memory pools. So, new `Memory` folder will be useful for containing these files.

qweasdd136963:
@culhatsker, that commit fixes crash with log:
```
[error] Expression    : fsimilar(m_tex_rect[fmLT].height(), m_tex_rect[fmT].height())
[error] Function      : InitTextureEx
[error] File          : /home/chip/xray-16/src/xrUICore/Windows/UIFrameWindow.cpp
[error] Line          : 51
[error] Description   : ui_pda2_fr
```

These files were deleted 2 days ago in this commit: https://github.com/OpenXRay/xray-16/commit/625339a4748b5334db7be7b2b6b18ea14a54ded5
I just restored them.

src/xrCommon/xr_deque.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 #include <deque>
-#include "xrCore/xrMemory.h"
+#include "xrCore/Memory/XRayAllocator.h"
```
FreeZoneMods: Do you really want to rollback changes from 170484 here? So, maybe remove these changes directly in 170484 to keep the diffs more simple?

src/xrCore/Memory/XRayAllocator.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
```
FreeZoneMods: Do you restore the file which was deleted in the previous commit? Or do you made some changes? Anyway, I think we should avoid removing it at all in the previous commit.

src/xrCore/Memory/xalloc.h
```diff
@@ -0,0 +1,87 @@
...
+
+    template <class Other>
+    xalloc& operator=(const xalloc<Other>&)
+    {
+        return *this;
+    }
+
+    pointer allocate(const size_type n, const void* p = nullptr) const { return xr_alloc<T>(n); }
+    void deallocate(pointer p, const size_type /*n*/) const { xr_free(p); }
+    void deallocate(void* p, const size_type /*n*/) const { xr_free(p); }
```
FreeZoneMods, line 40: The prototypes of the 'deallocate' methods looks just the same. Do you really need such overloads? Won't it lead to ambiguity errors in compile time? Or is it a designed and expected behaviour?

src/xrCore/xrMemory.h
```diff
@@ -12,19 +11,18 @@
 //#include "tbb/tbbmalloc_proxy.h" // Xottab_DUTY: works bad, disabled it..
 #endif
 
+#if defined(WINDOWS)
+#define NO_TBB_MALLOC
```
FreeZoneMods, line 13: Why you enable it only for windows? Does the Linux port has some problems with it?

src/xrEngine/pure.h
```diff
@@ -98,9 +98,9 @@ class MessageRegistry
             messages[0].Object->OnPure();
         else
         {
-            for (const auto& message : messages)
-                if (message.Prio != REG_PRIORITY_INVALID)
-                    message.Object->OnPure();
+            for (int i = 0; i < messages.size(); i++)
```
FreeZoneMods, line 101: What's the reason for such 'downgrading' of the for-expression? As for me, the initial variant seems to be more cool.

src/xrCore/xrMemory.h
qweasdd136963: No. On Linux this define can be enabled using CMake.

src/xrCommon/xr_deque.h
qweasdd136963: Fixed

src/xrCore/Memory/xalloc.h
qweasdd136963: I really don't know why we need this. But this line was in the original allocator from GSC, I just did not remove it.

src/xrEngine/pure.h
qweasdd136963: Using range-based for here with gsc allocator, I get a crash: `Exception thrown: read access violation.
message.Object was 0xF.`

qweasdd136963: Во время цикла на 3 итерации, в вектор messages добавляется еще один элемент, из-за этого все и слетает. Я пока не выяснил почему так происходит.

src/xrCore/xrMemory.h
eagleivg: On windows need use Common.props

src/xrEngine/pure.h
Xottab-DUTY: When you're adding some elements to the vector all iterators become invalid. That's why the crash happens.

This fix is correct.

We have faced with this problem with @intorr, when he edited renderers code 🙂

src/xrCore/Memory/xalloc.h
Xottab-DUTY: Probably, this helps avoid some problems..

src/xrEngine/pure.h
FreeZoneMods: Yes, all iterators became invalid. However, I think it is NOT a correct fix. Modifying vectors while iterating it is a very bad solution. It could lead to a bugs which could be very hard to find. Looks like the problem needs a further investigation.

FreeZoneMods, line 101: And looks like the current implementation is probably not ready to modification of the vector size while execution of the cycle. The 'i' variable will be incremented anyway. But if we are modifying the vector, it could lead to skipping some elements or to executing method OnPure twice. Please check it.

PR #320, opened 2019-01-04
Lua: Removed spam in a console about mouse events in Mixed/Debug build

ShokerStlk:
Currently it's print a console message for **every** mouse move.


collectioner:
@ShokerStlk It'd be better to remove code, rather than just comment it, IMHO.

Xottab-DUTY:
@collectioner, in this case it's better to comment it.. It may be useful for someone..
@ShokerStlk, oh.. Thank you, this is what we needed.

PR #323, opened 2019-01-09
FixedMAP comeback

Xottab-DUTY:
Returned FixedMap usage to xrRender. This container was used before and it is proven to be faster than xr_vector and xr_unordered_map.

### DIFF
--- a/src/xrCore/Containers/FixedMap.h
+++ b/src/xrCore/Containers/FixedMap.h
@@ -1,0 +1,425 @@
+#pragma once
+
+#include "xrCommon/xr_allocator.h"
+
+template <class K, class T>
+struct xr_fixed_map_node
+{
+    K first;
+    T second;
+
+    xr_fixed_map_node<K, T> *left, *right;
+
+    xr_fixed_map_node()
+        : first(), second(),
+          left(nullptr),
+          right(nullptr) {}
+
+    ~xr_fixed_map_node()
+    {
+        left = nullptr;
+        right = nullptr;
+    }
+
+    xr_fixed_map_node(xr_fixed_map_node&& other) noexcept
+    {
+        first = std::move(other.first);
+        second = std::move(other.second);
+        left = other.left;
+        right = other.right;
+        other.left = nullptr;
+        other.right = nullptr;
+    }
+
+    xr_fixed_map_node(const xr_fixed_map_node& other) noexcept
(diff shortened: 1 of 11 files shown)


src/xrCore/Containers/FixedMap.h
```diff
@@ -0,0 +1,406 @@
...
+struct xr_fixed_map_node
+{
+    xr_fixed_map_node()
+        : key(), val(),
+          left(nullptr),
+          right(nullptr) {}
+
+    ~xr_fixed_map_node()
+    {
+        left = right = nullptr;
```
FreeZoneMods: Don't write such expressions, please.

src/xrCore/Containers/FixedMap.h
```diff
@@ -0,0 +1,406 @@
...
+        K first;
+    };
+
+    union
+    {
+        T val;
+        T second;
+    };
+
+    xr_fixed_map_node<K, T> *left, *right;
```
FreeZoneMods: Maybe move fields to the beginning of the class declaration?

src/xrCore/Containers/FixedMap.h
```diff
@@ -0,0 +1,406 @@
+#pragma once
+
+#include "xrCommon/xr_allocator.h"
+
+template <class K, class T>
+struct xr_fixed_map_node
+{
+    xr_fixed_map_node()
+        : key(), val(),
```
FreeZoneMods: Looks very strange. Should "T" be a POD type? If yes (and I think we should put static_assert somewhere to check it), standard guarantees that 'key' would be initialized and the line is redundant. If no... What is the 'union from non-pod types'? And why we initialize it twice?
And maybe use tupples instead of unions?

src/xrCore/Containers/FixedMap.h
```diff
@@ -0,0 +1,406 @@
...
+        "xr_fixed_map<K, T, allocator> allocator mismatch");
+
+private:
+    value_type* nodes;
+    size_t pool;
+    size_t limit;
+
+    void resize()
+    {
+        size_t newLimit = limit + SG_REALLOC_ADVANCE;
```
FreeZoneMods: Exponential grow is more efficient.

FreeZoneMods (approved):

PR #324, opened 2019-01-09
Refactored particles multithreading

Xottab-DUTY:
It's faster to process heavy particles with in singlethreaded mode. So, I refactored the code to do it.

### DIFF
--- a/src/Layers/xrRender/ParticleEffect.cpp
+++ b/src/Layers/xrRender/ParticleEffect.cpp
@@ -453,126 +453,136 @@ __forceinline void magnitude_sse(Fvector& vec, float& res)
     _mm_store_ss((float*)&res, tv);
 }
 
-void ParticleRenderStream(FVF::LIT* pv, u32 count, PAPI::Particle * particles, CParticleEffect * pPE)
+void ParticleRenderStream(FVF::LIT* pv, u32 count, PAPI::Particle* particles, CParticleEffect* pPE)
 {
-    float sina = 0.0f, cosa = 0.0f;
-    // Xottab_DUTY: changed angle to be float instead of DWORD
-    // But it must be 0xFFFFFFFF or otherwise some particles won't play
-    float angle = 0xFFFFFFFF;
+    auto processParticle = [&](u32 i, Particle& m)
+    {
+        float sina = 0.0f, cosa = 0.0f;
 
-    FOR_START(u32, 0, count, i)
+        // Should be max or some particles won't play
+        float angle = std::numeric_limits<float>::max();
+
+        Fvector2 lt, rb;
+        lt.set(0.f, 0.f);
+        rb.set(1.f, 1.f);
+
+        _mm_prefetch((char*)&particles[i + 1], _MM_HINT_NTA);
+
+        if (angle != m.rot.x)
         {
-            PAPI::Particle& m = particles[i];
-            Fvector2 lt, rb;
-            lt.set(0.f, 0.f);
-            rb.set(1.f, 1.f);
+            angle = m.rot.x;
+            sina = sinf(angle);
+            cosa = cosf(angle);
+        }
 
-            _mm_prefetch((char*)&particles[i + 1], _MM_HINT_NTA);
+        _mm_prefetch(64 + (char*)&particles[i + 1], _MM_HINT_NTA);
 
-            if (angle != m.rot.x)
-            {
-                angle = m.rot.x;
-                sina = sinf(angle);
-                cosa = cosf(angle);
-            }
+        if (pPE->m_Def->m_Flags.is(CPEDef::dfFramed))
+            pPE->m_Def->m_Frame.CalculateTC(iFloor(float(m.frame) / 255.f), lt, rb);
 
-            _mm_prefetch(64 + (char*)&particles[i + 1], _MM_HINT_NTA);
+        float r_x = m.size.x * 0.5f;
+        float r_y = m.size.y * 0.5f;
+        float speed = 0.f;
+        bool speed_calculated = false;
 
-            if (pPE->m_Def->m_Flags.is(CPEDef::dfFramed))
-                pPE->m_Def->m_Frame.CalculateTC(iFloor(float(m.frame) / 255.f), lt, rb);
+        if (pPE->m_Def->m_Flags.is(CPEDef::dfVelocityScale))
+        {
+            magnitude_sse(m.vel, speed);
+            speed_calculated = true;
+            r_x += speed * pPE->m_Def->m_VelocityScale.x;
+            r_y += speed * pPE->m_Def->m_VelocityScale.y;
+        }
 
-            float r_x = m.size.x * 0.5f;
-            float r_y = m.size.y * 0.5f;
-            float speed = 0.f;
-            bool speed_calculated = false;
-
-            if (pPE->m_Def->m_Flags.is(CPEDef::dfVelocityScale))
-            {
+        if (pPE->m_Def->m_Flags.is(CPEDef::dfAlignToPath))
+        {
+            if (!speed_calculated)
                 magnitude_sse(m.vel, speed);
-                speed_calculated = true;
-                r_x += speed * pPE->m_Def->m_VelocityScale.x;
-                r_y += speed * pPE->m_Def->m_VelocityScale.y;
-            }
-
-            if (pPE->m_Def->m_Flags.is(CPEDef::dfAlignToPath))
+            if ((speed < EPS_S) && pPE->m_Def->m_Flags.is(CPEDef::dfWorldAlign))
             {
-                if (!speed_calculated)
-                    magnitude_sse(m.vel, speed);
-                if ((speed < EPS_S) && pPE->m_Def->m_Flags.is(CPEDef::dfWorldAlign))
-                {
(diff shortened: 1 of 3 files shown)


src/Layers/xrRender/ParticleEffect.cpp
```diff
@@ -453,126 +453,136 @@ __forceinline void magnitude_sse(Fvector& vec, float& res)
...
+            else
+            {
+                FillSprite(pv, RDEVICE.vCameraTop, RDEVICE.vCameraRight, m.pos, lt, rb, r_x, r_y, m.color, sina, cosa);
+            }
+        }
+    };
+
+    constexpr u32 max_count = 75;
+
+    if (true && count <= max_count)
```
Xottab-DUTY: Strange, but this cycle is always faster in multithreaded mode.

Since, this is a WIP pull request, the code was pushed as is.

src/xrParticles/particle_actions_collection.cpp
```diff
@@ -1740,21 +1660,75 @@ void PATurbulence::Execute(ParticleEffect* effect, const float dt, float& tm_max
...
+
+        _vmo = _mm_shuffle_ps(_vmo, _vmo, _MM_SHUFFLE(0, 0, 0, 0)); // _vmo = scale | scale | scale | scale
+        _mvel = _mm_mul_ps(_mvel, _vmo);
+
+        _mm_store_fvector(m.vel, _mvel);
+    };
+
+    constexpr u32 max_count = 75;
+
+    if (count < max_count)
```
Xottab-DUTY: > Strange, but this cycle is always faster in multithreaded mode.

But this cycle is faster in singlethreaded mode for heavy particles...

Why?

src/Layers/xrRender/ParticleEffect.cpp
```diff
@@ -453,126 +453,136 @@ __forceinline void magnitude_sse(Fvector& vec, float& res)
...
-    float sina = 0.0f, cosa = 0.0f;
-    // Xottab_DUTY: changed angle to be float instead of DWORD
-    // But it must be 0xFFFFFFFF or otherwise some particles won't play
-    float angle = 0xFFFFFFFF;
+    auto processParticle = [&](u32 i, Particle& m)
+    {
+        float sina = 0.0f, cosa = 0.0f;
+        // Xottab_DUTY: changed angle to be float instead of DWORD
+        // But it must be 0xFFFFFFFF or otherwise some particles won't play
+        float angle = 0xFFFFFFFF;
```
FreeZoneMods: I think we should get rid of this magic value and make a constant like ANGLE_NO_PARTICLES = 0xFFFFFFFF. And assigning such value to the float variable is... strange. It doesn't set all bits of the variable to 1 (as you can think when looking through this code), it assigns the value '4.29497e+09' to the variable.

src/Layers/xrRender/ParticleEffect.cpp
```diff
@@ -453,126 +453,136 @@ __forceinline void magnitude_sse(Fvector& vec, float& res)
...
+                FillSprite(pv, RDEVICE.vCameraTop, RDEVICE.vCameraRight, p, lt, rb, r_x, r_y, m.color, sina, cosa);
+            }
+            else
+            {
+                FillSprite(pv, RDEVICE.vCameraTop, RDEVICE.vCameraRight, m.pos, lt, rb, r_x, r_y, m.color, sina, cosa);
+            }
+        }
+    };
+
+    constexpr u32 max_count = 75;
```
FreeZoneMods: Maybe we should use capital letters for naming constants?

src/Layers/xrRender/ParticleEffect.cpp
FreeZoneMods: Why you add 'true' to the condition?

src/xrParticles/particle_actions_collection.cpp
FreeZoneMods: I think it could depend from the hardware features.

src/Layers/xrRender/ParticleEffect.cpp
Xottab-DUTY: > Strange, but this cycle is always faster in multithreaded mode.
Since, this is a WIP pull request, the code was pushed as is.

src/Layers/xrRender/ParticleEffect.cpp
Xottab-DUTY: These variables was `DWORD` before, I changed the type to `float`, since they was anyway casted to floats. If we set this to `0.0f` or `1.0f` or another value, then particles stop working.
This magic value `0xFFFFFFFF` is required, if you wan't to see the particles =)
So, this value is not `ANGLE_NO_PARTICLES`, but I'd say `PARTICLES_DEFAULT_ANGLE`.

xrombik:
```
float angle = 0.0f;
*((u32*) &angle) = 0xFFFFFFFF;
```

FreeZoneMods: Apples and pineapples are different fruits. Don't mix them or you'll get problems...

OmniBlade: Nothing wrong with this per se to force a given bit pattern, but on clang and GCC you will need to disable strict aliasing or use the __may_alias__ attribute as well in the cast or you may find it optimised out.

FreeZoneMods: Nothing wrong, but the code _smells_. And necessity to disable compiler checks confirms it.

xrombik: > Nothing wrong, but the code _smells_. And necessity to disable compiler checks confirms it.

@FreeZoneMods 
Are you shure?

src/xrParticles/particle_actions_collection.cpp
xrombik: @FreeZoneMods 
What hardware and features do you mean?

src/Layers/xrRender/ParticleEffect.cpp
xrombik: > Maybe we should use capital letters for naming constants?

Oh no! We use capital letters for `#define`-constants only!

src/Layers/xrRender/ParticleEffect.cpp
OmniBlade: It falls foul of C++ standard rules relating to type aliasing where the standard is IMO poorly defined. Accessing the memory of one type through a pointer to another type falls into undefined behaviour, so clang and GCC will basically remove the assignment. However for floating point types this kind of type punning is actually incredibly useful and indeed the underlying maths libraries for floating point operations almost all certainly make use of it. In C doing this using a union is explicitly allowed and GCC at least allows it in C++ code as well.
Personally I would either define types using the GCC/clang attributes to allow it and use those or use a union to do the assignment as both will bring it into defined behaviour for the supported compilers. If you want to use the idiomatic but IMO silly workaround way of doing this for C++ you could do `memset(&angle, 0xFF, sizeof(angle));` to give it the correct value. Some compilers will ellide this to an assignment in the actual generated assembly making the gymnasics silly, but that is where we are.

FreeZoneMods: @OmniBlade, thank you for your detailed reply. I fully agree that accessing some data via pointer of another type is a 'gray' area and it's easy to violate rules and get UB (especially when the optimization is turned on). I'm not sure if accessing floats via pointers to dwords is explicitly allowed by the standard and won't lead to such UBs. However, even if this trick is fully legal, I'd not recommend to use it in the code if you can write the code without it. The engine is not designed for executing on 'hard-embedded' devices, so saving 4 bytes can be neglected. But it's much more important to keep the code easy to read and understand. I don't think that such trick makes our code more clear and more robust, but I'm sure that using it could lead to 'magic' bugs in the future (the other developers could miss this trick and try to use this 'invalid' value as 'valid').

src/Layers/xrRender/ParticleEffect.cpp
FreeZoneMods: Ok, keep using it if you think that this style is better. However, in my opinion the capital letters should be used for all constants because it helps us to make the code more readable.

src/xrParticles/particle_actions_collection.cpp
FreeZoneMods: Unfortunately, I can't tell you the exact feature. Modern CPUs are too complex to fully understand the behavior of the code. The simplest example relates to caches in the processor. When some block of the code is fully located in the cache, the program works very fast, but when you tries to optimize the program (for example via loop unrolling) you could see a terrible performance drop (instead of expected performance improvements).

src/Layers/xrRender/ParticleEffect.cpp
OmniBlade: I guess if other parts of the code are actually casting back to int and comparing against this magic value, then yeah, its probably just best to refactor it out and choose some other invalid float value to assign that can be directly compared against as a float.
As far as I know, only char and unsigned char are allowed by the standard to alias other types, but only Clang and GCC to my knowledge actually optimise based on it and provide none portable attributes to get around it if you want/need to.

Issue #232, opened 2018-08-10
Block any buttons except "Delete" for incompatible game save file


Xottab-DUTY:
I'd say this is an advanced fix for #226

PR #328, opened 2019-01-12
Disabling load invalid saves

collectioner:
Fix for #232

### DIFF
--- a/res/gamedata/scripts/ui_load_dialog.script
+++ b/res/gamedata/scripts/ui_load_dialog.script
@@ -1,0 +1,354 @@
+saved_game_extension	= ".scop"
+
+class "load_item" (CUIListBoxItem)
+
+function load_item:__init(height) super(height)
+	printf("%f",height)
+	self.file_name		= "filename"
+
+		self:SetTextColor(GetARGB(255, 170, 170, 170))
+
+		self.fn = self:GetTextItem()
+		self.fn:SetFont(GetFontLetterica18Russian())
+		self.fn:SetEllipsis(true)
+end
+
+function load_item:__finalize()
+end
+
+class "load_dialog" (CUIScriptWnd)
+
+function load_dialog:__init() super()
+	 self:InitControls()
+	 self:InitCallBacks()
+end
+
+function load_dialog:__finalize()
+end
+
+function load_dialog:FillList()
+	self.list_box:RemoveAll()
+	local f		= getFS()
+	local flist = f:file_list_open_ex("$game_saves$",bit_or(FS.FS_ListFiles,FS.FS_RootOnly),"*"..saved_game_extension)
+	local f_cnt = flist:Size()
+
+	flist:Sort(FS.FS_sort_by_modif_down)
+
+	for	it=0, f_cnt-1 	do
+		local file		=	flist:GetAt(it)
+		local file_name = string.sub(file:NameFull(), 0, (string.len(file:NameFull()) - string.len(saved_game_extension)))
+		local date_time = "[" .. file:ModifDigitOnly() .. "]"
+		--menu_item =  ..
+		self:AddItemToList(file_name, date_time)
+	end
+end
+
+function load_dialog:InitControls()
+	self:SetWndRect			(Frect():set(0,0,1024,768))
+
+	local xml				= CScriptXmlInit()
+	local ctrl
+	xml:ParseFile			("ui_mm_load_dlg.xml")
+
+	xml:InitStatic			("background",self)
+
+	ctrl					= CUIWindow()
+	xml:InitWindow			("file_item:main",0,ctrl)
+
+	self.file_item_main_sz	= vector2():set(ctrl:GetWidth(),ctrl:GetHeight())
+
+	xml:InitWindow			("file_item:fn",0,ctrl)
+	self.file_item_fn_sz	= vector2():set(ctrl:GetWidth(),ctrl:GetHeight())
+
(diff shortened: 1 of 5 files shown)


src/xrGame/saved_game_wrapper.cpp
```diff
@@ -74,6 +74,7 @@ CSavedGameWrapper::CSavedGameWrapper(LPCSTR saved_game_name)
         m_actor_health = 1.f;
         m_level_id = _LEVEL_ID(-1);
         m_level_name = "";
+        m_valid_save = false;
```
FreeZoneMods, line 77: I think it would be better to initialize 'default-invalid' values directly in the header (or in the inilialization list). It should help us to keep the state of the object consistent and to prevent some stupid errors in the future.

src/xrGame/saved_game_wrapper.cpp
```diff
@@ -161,10 +163,15 @@ CSavedGameWrapper::CSavedGameWrapper(LPCSTR saved_game_name)
         {
             CGameGraph graph(*chunk);
             m_level_id = graph.vertex(object->m_tGraphID)->level_id();
+
             if (graph.header().level_exist(m_level_id))
                 m_level_name = graph.header().level(m_level_id).name();
+
```
FreeZoneMods, line 169: Maybe it would be better to add brackets instead of adding additional line for improving readability?

res/gamedata/scripts/ui_load_dialog.script
```diff
@@ -81,9 +81,9 @@ function load_dialog:InitControls()
 	self.list_box:ShowSelectedItem	(true)
 	self:Register			(self.list_box, "list_window")
 
-	ctrl					= xml:Init3tButton("form:btn_load",	self.form)
-	self:Register			(ctrl, "button_load")
-
+	self.load_btn			= xml:Init3tButton("form:btn_load",	self.form)
+	self:Register			(self.load_btn, "button_load")
+
```
FreeZoneMods, line 86: Remove this extra tab, please

res/gamedata/scripts/ui_load_dialog.script
```diff
@@ -154,6 +154,10 @@ function file_data(fname)
 		local file		=	flist:GetAt(0)
 		local sg = CSavedGameWrapper(fname)
 
+		local isSaveValid = sg:valid_save()
+		self.load_btn:Enable(isSaveValid)
```
FreeZoneMods, line 158: I think we shouldn't allow this function to modify state of the UI. The function is designed for parsing save data, so by editing the dialog state we violate the single-responsibility principle. I suggest you to return from this function the state of parsing data (true for successful parsing, false otherwise) and process the result in load_dialog:OnListItemClicked.

src/xrGame/saved_game_wrapper.cpp
```diff
@@ -88,6 +89,7 @@ CSavedGameWrapper::CSavedGameWrapper(LPCSTR saved_game_name)
         CALifeTimeManager time_manager(alife_section);
         time_manager.load(reader);
         m_game_time = time_manager.game_time();
+        m_valid_save = true;
```
FreeZoneMods, line 92: Hmm... Does the save is really 'valid' in case of returning from lines 118, 130, 149, 160?

src/xrGame/saved_game_wrapper.h
```diff
@@ -22,6 +22,7 @@ class CSavedGameWrapper
     _LEVEL_ID m_level_id;
     shared_str m_level_name;
     float m_actor_health;
+    bool m_valid_save;
```
FreeZoneMods, line 25: Do we really need this boolean value? Looks like we can easily determine if the save is valid or not by inspecting the values of members m_level_id, m_level_name. So maybe it will be better to create and export a method like 'can_be_loaded' for determining validity instead of creating extra flags?

src/xrGame/saved_game_wrapper.cpp
collectioner: Good question. Need @Xottab-DUTY help to clarify is it true or not.

src/xrGame/saved_game_wrapper.h
collectioner: Again, it depends on previous comment. I guess that game can be loaded with empty m_level_id or m_level_name, because I don't find any validations for acceptable values of these variables. So, if it's true, I think we need another variable, like this one. If game cannot be loaded - I agree with you, we can re-write with additional method.

res/gamedata/scripts/ui_load_dialog.script
collectioner: Agree, will fix in the next commit.

res/gamedata/scripts/ui_load_dialog.script
collectioner: Agree, will fix in the next commit.

src/xrGame/saved_game_wrapper.cpp
collectioner: It's not clear for me how we can add brackets. Like this:
```
{
if (…)

}
else (…)
```
… or like this …
```
{
if (….)
else (….){}
}
```
?

src/xrGame/saved_game_wrapper.cpp
collectioner: I think this is should be next PR, like code-style.

collectioner:
Closed as origin branch was removed.

PR #330, opened 2019-01-14
Resources \ UI: Improved FOV tracker in Settings

ShokerStlk:
1) Step is set to 1.0
2) Show numeric fov value

image

### DIFF
--- a/src/xrUICore/TrackBar/UITrackBar.cpp
+++ b/src/xrUICore/TrackBar/UITrackBar.cpp
@@ -13,6 +13,12 @@ CUITrackBar::CUITrackBar()
     m_pSlider = new CUI3tButton();
     AttachChild(m_pSlider);
     m_pSlider->SetAutoDelete(true);
+
+    m_static = new CUIStatic();
+    m_static->Enable(false);
+    AttachChild(m_static);
+    m_static->SetAutoDelete(true);
+
     m_b_mouse_capturer = false;
 }
 
@@ -106,6 +112,7 @@ void CUITrackBar::Draw()
 {
     CUI_IB_FrameLineWnd::Draw();
     m_pSlider->Draw();
+    m_static->Draw();
 }
 
 void CUITrackBar::Update()
@@ -290,6 +297,20 @@ void CUITrackBar::UpdatePos()
         pos.x = free_space - pos.x;
 
     m_pSlider->SetWndPos(pos);
+
+    if (m_static->IsEnabled())
+    {
+        string256 buff;      
+        if (m_b_is_float)
+        {
+            xr_sprintf(buff, (m_static_format == nullptr ? "%.1f" : m_static_format.c_str()), m_f_val);
+        }
+        else
+        {
+            xr_sprintf(buff, (m_static_format == nullptr ? "%d" : m_static_format.c_str()), m_i_val);
+        }
+        m_static->TextItemControl()->SetTextST(buff);
+    }
 }
 
 void CUITrackBar::OnMessage(LPCSTR message)
--- a/src/xrUICore/TrackBar/UITrackBar.h
+++ b/src/xrUICore/TrackBar/UITrackBar.h
@@ -34,6 +34,9 @@ public:
     void SetOptIBounds(int imin, int imax);
     void SetOptFBounds(float fmin, float fmax);
 
+    CUIStatic* m_static;
+    shared_str m_static_format;
+
 protected:
     void UpdatePos();
     void UpdatePosRelativeToMouse();
--- a/src/xrUICore/XML/UIXmlInitBase.cpp
+++ b/src/xrUICore/XML/UIXmlInitBase.cpp
@@ -1167,6 +1167,15 @@ bool CUIXmlInitBase::InitTrackBar(CUIXml& xml_doc, LPCSTR path, int index, CUITr
         }
     }
 
+    string512 buf;
+    strconcat(sizeof(buf), buf, path, ":output_wnd");
+    if (xml_doc.NavigateToNode(buf, index))
+    {
+        InitStatic(xml_doc, buf, index, pWnd->m_static);
+        pWnd->m_static_format = xml_doc.ReadAttrib(buf, index, "format", nullptr);
+        pWnd->m_static->Enable(true);
+    }
+
     return true;
 }
 
--- a/res/gamedata/configs/ui/ui_mm_opt.xml
+++ b/res/gamedata/configs/ui/ui_mm_opt.xml
@@ -587,8 +587,11 @@
 		<cap_fov x="20" y="3" width="135" height="24">
 			<text r="170" g="170" b="170" font="letterica16" align="r" vert_align="c">ui_mm_fov</text>
 		</cap_fov>
-		<track_fov x="180" y="5" width="235" height="20" min="55.0" max="115.0">
+		<track_fov x="180" y="5" width="205" height="20" min="55.0" max="115.0" step="1.0">
 			<options_item entry="fov" group="mm_opt_gameplay" />
+			<output_wnd x="215" y="0" width="20" height="20" format="%.0f">
+				<text r="170" g="170" b="170" font="letterica16" align="r" vert_align="c">0</text>
+			</output_wnd>
 		</track_fov>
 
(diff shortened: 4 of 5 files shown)


res/gamedata/scripts/ui_mm_opt_gameplay.script
```diff
@@ -87,3 +88,11 @@ function opt_gameplay:InitControls(x, y, xml, handler)
 	btn = xml:Init3tButton	("tab_gameplay:btn_check_updates", self)
 	handler:Register		(btn, "btn_check_updates")
 end
+
+function opt_gameplay:Update(handler)
```
FreeZoneMods: Can we find a way to avoid checking it on every update? Maybe we could subscribe a callback to track fov changing?

res/gamedata/scripts/ui_mm_opt_gameplay.script
```diff
@@ -35,7 +35,8 @@ function opt_gameplay:InitControls(x, y, xml, handler)
 	
 	_st = xml:InitStatic("tab_gameplay:templ_item", nil)
 	xml:InitStatic("tab_gameplay:cap_fov", _st)
-	xml:InitTrackBar("tab_gameplay:track_fov", _st)
+	self.fov_track = xml:InitTrackBar("tab_gameplay:track_fov", _st)
+	self.fov_track_val = xml:InitStatic("tab_gameplay:cap_fov_val", _st)
```
FreeZoneMods: Hmm... maybe it will be better to implement an option for displaying current value of the track directly in the engine? It could be useful for reusability in the future.

collectioner: It looks like option just for option. If we'll need this in future, we'll make it in future. Now it'll be unused/dead code, which is not very good.

Xottab-DUTY: @FreeZoneMods is right. It will be very useful to implement such option for trackbar in the engine so that other modders can use it.

Xottab-DUTY: @ShokerStlk could you do that?

res/gamedata/scripts/ui_mm_opt_gameplay.script
ShokerStlk: Слишком много возни и кода. В движке есть только родной калбэк на изменение ползунка (ui_events.BUTTON_CLICKED - на который нужно ещё писать код подписки). Далее чтобы **при открытии** меню сразу выводилась нужная цифра в тексте - пришлось у ui_mm_opt_gameplay.script переопределять метод Show. И то это оказалась не панацея - калбэк на Show срабатывал только при смене вкладок в опциях, а при открытии \ закрытии всего окна опций он не вызывается. Пришлось ещё делать врезку в ui_mm_opt_main.script, чтобы из неё обновлять значения в ui_mm_opt_gameplay.script (options_dialog:SetCurrentValues()). 

Я посмотрел на то, что получается, и заменил более лёгким и **надёжным** кодом через Update.
 
- - - - - - - - -
It's too much work and code. In the engine there is only a native callback to change the slider (ui_events.BUTTON_CLICKED - to which you still need to write the subscription code). You also need to override the Show method in ui_mm_opt_gameplay.script + add some code to ui_mm_opt_gameplay.script (options_dialog:SetCurrentValues()). 

I looked at what was coming out and replaced it with much simple and **reliable** code via Update.

res/gamedata/scripts/ui_mm_opt_gameplay.script
ShokerStlk: UITrackBar.cpp does not support any text by default (it's not even child of the CUIStatic). I started doing similar to what GSC have in UIComboBox, but soon dropped it, and moved to Lua. To be honest i don't want to spend more time on this.

ShokerStlk: Ok, i will try now to implenet this through engine.

ShokerStlk:
Done.

src/xrUICore/XML/UIXmlInitBase.cpp
```diff
@@ -1167,6 +1167,15 @@ bool CUIXmlInitBase::InitTrackBar(CUIXml& xml_doc, LPCSTR path, int index, CUITr
         }
     }
 
+    string512 buf;
```
FreeZoneMods, line 1170: Why not xr_string?

src/xrUICore/TrackBar/UITrackBar.cpp
```diff
@@ -290,6 +297,22 @@ void CUITrackBar::UpdatePos()
...
+    {
+        string128 buff;      
+        if (m_b_is_float)
+        {
+            xr_sprintf(buff, (m_static_format == nullptr ? "%.1f" : m_static_format.c_str()), m_f_val);
+            m_static->TextItemControl()->SetTextST(buff);
+        }
+        else
+        {
+            string32 buff;
```
FreeZoneMods: The second variable named 'buff'??? And maybe use xr_string?

src/xrUICore/TrackBar/UITrackBar.cpp
```diff
@@ -290,6 +297,22 @@ void CUITrackBar::UpdatePos()
...
+            xr_sprintf(buff, (m_static_format == nullptr ? "%d" : m_static_format.c_str()), m_i_val);
+            m_static->TextItemControl()->SetTextST(buff);
```
FreeZoneMods: I think we can avoid duplication of this line.

src/xrUICore/TrackBar/UITrackBar.h
```diff
@@ -34,6 +34,9 @@ class XRUICORE_API CUITrackBar : public CUI_IB_FrameLineWnd, public CUIOptionsIt
     void SetOptIBounds(int imin, int imax);
     void SetOptFBounds(float fmin, float fmax);
 
+    CUIStatic* m_static;
```
FreeZoneMods, line 37: xr_unique_ptr?

ShokerStlk: Never worked with unique_ptr.

src/xrUICore/XML/UIXmlInitBase.cpp
ShokerStlk: 100% copy paste from GSC code.

src/xrUICore/TrackBar/UITrackBar.h
Xottab-DUTY: No need for this. S.T.A.L.K.E.R. UI is completely filled with raw pointers.

src/xrUICore/TrackBar/UITrackBar.cpp
ShokerStlk: Don't know what is better. Will fix buff.

Xottab-DUTY: It's better to use simple type which `string32` is.

src/xrUICore/XML/UIXmlInitBase.cpp
xrombik: @FreeZoneMods `xr_string` has code overheads for this point of use.

FreeZoneMods: This code executes not so frequently, so, the performance decreasing shouldn't be noticeably. Using xr_string will make the code slight safer.
I think we shouldn't follow bad practices (introduced by GSC) when writing new code and get rid of such code with the development of the project.

src/xrUICore/TrackBar/UITrackBar.h
FreeZoneMods: And that's sad.

src/xrUICore/TrackBar/UITrackBar.cpp
FreeZoneMods: std::string usually has small-string optimization technique (to avoid allocations if the length of the string is lesser than 16 bytes, for example). So, xr_string seems to be a good candidate here.

FreeZoneMods (approved):

PR #331, opened 2019-01-19
Fixed `CMake removes SDK/include/tbb #313`

xrombik:


CMakeLists.txt
```diff
@@ -107,14 +107,6 @@ add_definitions(-DdSINGLE) # for ODE
 set(LUA_INCLUDE_DIR Externals/LuaJIT/src)
 
 if (NOT WIN32)
-if(EXISTS ${CMAKE_SOURCE_DIR}/Externals/cryptopp/dsa.h)
```
eagleivg, line 110: Have you even tried to build something with such an edit?

eagleivg (changes requested):
If you want stop removind tbb includes, your need add correct includes for all subprojects.

CMakeLists.txt
xrombik: Yes.
`AppVeyor build succeeded`
https://ci.appveyor.com/project/OpenXRay/xray-16/builds/21745526

eagleivg: AppVeyor used msvc solution for build, not cmake!
https://travis-ci.org/OpenXRay/xray-16/jobs/481854470

xrombik: I have saw this error after merging with some two last PR, but then i roll back to current state. I have clone and build, everything is ok. I'm confused.

drug007: The reason of this is that the version of tbb that is specific to openxray CAN (in contrary to should) conflicts with system libs. Either we should build ALL from our sources or we should use only system libs. Of course it's impossible in fact. More practical way is to make using own version of tbb optional. This way also can be used with other libs those specific versions are used in openxray.
```
option( USING_OWN_TBB   "Using specific tbb version from OpenXRay" ON )
if( USING_OWN_TBB )
{
    include_directories(path/to/tbb);
}
```
Only drawback I can see that this demands to manually add calls of `include_directories` to every dependency.

xrombik: Well i see this error in my local build now. I suppose, that success build appeared with no `tbb` folder present. This file removal causes unpredictable problems in `build/commit/pull/push`. You have to advise more robust solution.

drug007: I do not have to, first of all. Then this is rather robust solution that allows to avoid using `tbb` without removing it. What's wrong with my proposal?

xrombik: @drug007 

```
option( USING_OWN_TBB   "Using specific tbb version from OpenXRay" ON )
if( USING_OWN_TBB )
{
    include_directories(path/to/tbb);
}
```

Where this lines should be included?

drug007: **Disclaimer: not tested at all**

I would delete these [lines](https://github.com/OpenXRay/xray-16/blob/xd_dev/CMakeLists.txt#L114) in root `CMakeLists.txt` then this [line](https://github.com/OpenXRay/xray-16/blob/xd_dev/CMakeLists.txt#L132) and then manually unroll the last line to list of all 3rd party dependencies except `tbb` like:
```
include_directories(${CMAKE_SOURCE_DIR})
include_directories({CMAKE_SOURCE_DIR}/Common)

include_directories(${CMAKE_SOURCE_DIR}/Externals/dependency0)
include_directories(${CMAKE_SOURCE_DIR}/Externals/dependency1)
...
include_directories(${CMAKE_SOURCE_DIR}/Externals/dependencyN)

include_directories(${CMAKE_SOURCE_DIR}/Externals/gli/external/glm)
include_directories(${CMAKE_SOURCE_DIR}/../sdk/include/loki) # hmm, haven't we got rid of loki?
```
Then add lines:
```
option( USING_OWN_TBB   "Using specific tbb version from OpenXRay" ON )
if( USING_OWN_TBB )
{
    include_directories(${CMAKE_SOURCE_DIR}/Externals/tbb);
}
```
There is some probability that `FindTBB.cmake` module will need some tweaking to support this approach in case when system version of `tbb` is installed but openxray version is used.

If you do not have system version of `tbb` it should work out of box. So testing of this patch by  @eagleivg (with USING_OWN_TBB equals to `OFF`) is mandatory to prove the patch  correctness.

eagleivg (commented):
Блин, 600 файлов! Нафига?

drug007:
@xrombik even if you need to do something with `cryptopp` this definitely should be done in another PR. This PR is about `tbb` only.

xrombik:
Just some edits in VisualStudio build repearing is left.

Xottab-DUTY:
Moving `cryptopp` to `cryptopp/cryptopp` is a nonsense. It won't be accepted.

xrombik:
> Moving `cryptopp` to `cryptopp/cryptopp` is a nonsense. It won't be accepted.

I am sure that you have enough experience to offer an even better solution.

eagleivg:
Close due 8f6af2bc195831d35d66e204f91ce249ef97acc1

PR #340, opened 2019-01-29
Double free and game intro fix

vamit611:


collectioner:
Hi, @vamit611 
We really appreciate your help and would love to see you on our official Discord server, if you aren't there yet. 

Collector.

src/xrGame/PHCommander.cpp
```diff
@@ -17,10 +17,10 @@ CPHCall::~CPHCall()
     xr_delete(m_action);
     xr_delete(m_condition);
 }
-bool CPHCall::obsolete() { return !m_action || m_action->obsolete() || !m_condition || m_condition->obsolete(); }
+bool CPHCall::obsolete() { return m_action->obsolete() || m_condition->obsolete(); }
```
drug007: Are you sure that `m_action` and `m_condition` always aren't null? The previous variant handles the situation when some of these fields is null but yours variant would segfault.

src/xrGame/PHCommander.cpp
```diff
@@ -17,10 +17,10 @@ CPHCall::~CPHCall()
...
 void CPHCall::check()
 {
-    if (m_condition && m_condition->is_true() && m_action)
+    if (m_condition->is_true() && m_action)
```
drug007: Also here you remove check for `m_condition` but the one for `m_action` still exists

src/xrGame/PHCommander.cpp
vamit611: I suppose these changes were added as an attempt to fix bugs that have occurred during the adding of tbb allocator.

vamit611:
Fix game intro. It starts now.

src/xrGame/PHCommander.cpp
FreeZoneMods: Let's won't remove this check. If you are sure these variables must always be not equal to zero - please add VERIFY or R_ASSERT statement to ensure it. It will help us to discover bugs in the future.

src/xrGame/PHCommander.cpp
```diff
@@ -17,10 +17,10 @@ CPHCall::~CPHCall()
     xr_delete(m_action);
     xr_delete(m_condition);
 }
-bool CPHCall::obsolete() { return !m_action || m_action->obsolete() || !m_condition || m_condition->obsolete(); }
+bool CPHCall::obsolete() { return m_action->obsolete() || m_condition->obsolete(); }
 void CPHCall::check()
 {
-    if (m_condition && m_condition->is_true() && m_action)
+    if (m_condition->is_true())
```
FreeZoneMods: Please, don't remove the checks blindly. If you are sure the member is always non-zero - use VERIFY macro.

src/xrGame/PHCommander.h
```diff
@@ -73,6 +73,7 @@ class CPHCommander : public IPHWorldUpdateCallbck
 {
     Lock lock;
     PHCALL_STORAGE m_calls;
+    xr_unordered_map<CPHCall*, bool> m_callsUpdateDeferred;
```
FreeZoneMods, line 76: What about using xr_unique_ptr instead of raw pointer here?

src/xrGame/PHCommander.h
```diff
@@ -82,13 +83,16 @@ class CPHCommander : public IPHWorldUpdateCallbck
         CPHCondition* condition, CPHReqComparerV* cmp_condition, CPHAction* action, CPHReqComparerV* cmp_action);
     void add_call(CPHCondition* condition, CPHAction* action);
     void add_call_threadsafety(CPHCondition* condition, CPHAction* action);
+    void AddCallDeferred(CPHCondition* condition, CPHAction* action);
```
FreeZoneMods, line 86: Is it possible to add 'const' modifier to the arguments?

src/xrUICore/Cursor/UICursor.cpp
```diff
@@ -107,19 +107,11 @@ Fvector2 CUICursor::GetCursorPositionDelta()
...
-        vPos.y = (float)pti.y * (UI_BASE_HEIGHT / (float)Device.m_rcWindowClient.h);
-    }
-    else
-    {
-        float sens = 1.0f;
-        vPos.x += _dx * sens;
-        vPos.y += _dy * sens;
-    }
+
+    float sens = 1.0f;
```
FreeZoneMods: Does it works normally when the game is running in windowed mode? Are you sure the game cursor and the system cursor won't live with own lives?

src/xrGame/PHCommander.cpp
SkyLoaderr: Btw, in some cases, people have a crash here, because m_condition or m_action is nullptr.

src/xrGame/ui/UIGameTutorial.cpp
```diff
@@ -97,21 +97,22 @@ bool CUISequenceItem::Stop(bool bForce)
 CUISequencer::CUISequencer() { m_flags.zero(); }
 void CUISequencer::Start(LPCSTR tutor_name)
 {
-    // Skip any tutorial except "game_loaded", since we need to show "st_press_any_key" hint
-    if (load_screen_renderer.IsActive() && xr_strcmp(tutor_name, "game_loaded") != 0)
+    VERIFY(m_sequencer_items.size() == 0);
+
+    CUIXml uiXml;
+    uiXml.Load(CONFIG_PATH, UI_PATH, UI_PATH_DEFAULT, "game_tutorials.xml");
```
Xottab-DUTY: I believe it should be done in other way to not touch game resources.

Tutorial works for Mixed configuration and it works in original COP, so, it seems that there's something wrong in our code which has `#ifdef` `MASTER_GOLD`, `DEBUG` or `NDEBUG` macros.

src/xrGame/PHCommander.cpp
```diff
@@ -48,6 +48,29 @@ void CPHCommander::clear()
         delete_call(it);
 
     m_calls.clear();
+    m_callsUpdateDeferred.clear();
+}
+
+void CPHCommander::UpdateDeferred()
+{
```
Xottab-DUTY, line 55: I hope that deferred update will finally fix the problem..

src/xrGame/PHCommander.h
vamit611: No, because next stage wiil be assigning of 'condition'/'action' to 'm_condition'/'m_action'. And entity of type 'const CPHCondition*' can't be assigned to an entity of type 'CPHCondition*'. Therefore, we'll have to use 'const_cast'.

src/xrUICore/Cursor/UICursor.cpp
vamit611: I tested this code and I didn't see anything strange.

src/xrGame/ui/UIGameTutorial.cpp
vamit611: @Xottab-DUTY, what code? Game intro didn't start because
`if (load_screen_renderer.IsActive() && xr_strcmp(tutor_name, "game_loaded") != 0)`
returned true.

Ps.
Game intro (prehistory of the zone, I mean) is a tutorial, and it didn't work before on any configuuration.

src/xrGame/PHCommander.h
vamit611: Hmm.. Look at [this](https://github.com/OpenXRay/xray-16/pull/340/commits/2ea57b5461e0334c61ba062fbb2e75398e45d79d#diff-53994ec8d3d0fa54a96023d820d9de14R214). As I understand, it's impossible to get unique_ptr from raw pointer.

src/xrGame/ui/UIGameTutorial.cpp
ShokerStlk: It's better to check why xr_strcmp(tutor_name, "game_loaded")  does not work, so we can keep gamedata folder as empty, as possible. Anyaway thnx for fix!

vamit611:
Done.

Xottab-DUTY:
> Game intro (prehistory of the zone, I mean) is a tutorial, and it didn't work before on any configuuration.

Maybe it's my fault. Do you know when it stopped working?

vamit611:
@Xottab-DUTY, https://github.com/OpenXRay/xray-16/commit/3e334e43c9a12bea4616e52c266bc74a644453d4#diff-94811d53ac022b7f420826ee2c721101

Xottab-DUTY:
Ah, thank you.

Xottab-DUTY:
Can you describe the bug with the cursor in windowed mode?

vamit611:
@Xottab-DUTY,
1) Launch the game in windowed mode and open options.
2) Try to move your mouse.

vamit611:
First, the cursor moves to the center of the screen, and then, after you move the mouse, it immediately appears in another location.

Xottab-DUTY:
Sorry for the late reply!

> First, the cursor moves to the center of the screen, and then, after you move the mouse, it immediately appears in another location.

Ah, I know what is it.. This bug should be fixed by moving the system mouse to the center of the screen.

I'm asking you to remove this fix so we can just merge your PR and step forward. 😃

src/xrGame/ui/UIGameTutorial.cpp
```diff
@@ -97,8 +97,9 @@ bool CUISequenceItem::Stop(bool bForce)
 CUISequencer::CUISequencer() { m_flags.zero(); }
 void CUISequencer::Start(LPCSTR tutor_name)
 {
-    // Skip any tutorial except "game_loaded", since we need to show "st_press_any_key" hint
-    if (load_screen_renderer.IsActive() && xr_strcmp(tutor_name, "game_loaded") != 0)
+    // Skip any tutorial except "game_loaded" and "intro_game", on load screen
+    if (load_screen_renderer.IsActive() && xr_strcmp(tutor_name, "game_loaded") != 0 &&
```
Xottab-DUTY, line 101: Should be OR instead of AND
```suggestion
    if (load_screen_renderer.IsActive() && xr_strcmp(tutor_name, "game_loaded") != 0 ||
```

vamit611: Why? Ok. Assume that `load_screen_renderer.IsActive() == true` and `tutor_name == "game_loaded"`.

The first case:
`if (true && false && true)` returns false and tutorial "game_laoded" will start.

In your case:
`if (true && false || true)` returns true and the execution of the function is stopped.

Xottab-DUTY: Oh! Sorry, you're right. I just mixed up `strcmp == 0` and `strcmp != 0` conditions.

Xottab-DUTY (approved):

Xottab-DUTY:

PR #346, opened 2019-02-04
Numerous improvements for command line renderer selection

vTurbine:
* Fixed a bug when renderer setting was overrided by `user.ltx`
* Fixed `rsXX` device flags updating
* `CCC_r2` renamed to `CCC_renderer`

### DIFF
--- a/src/xrEngine/Device_Initialize.cpp
+++ b/src/xrEngine/Device_Initialize.cpp
@@ -50,8 +50,7 @@ void CRenderDevice::Initialize()
 
   
 #if !defined(LINUX)
-        // xxx: need to fix getting rsRGL flag, it doesn't work now
-        if (strstr(Core.Params, "-gl") || psDeviceFlags.test(rsRGL))
+        if (psDeviceFlags.test(rsRGL))
 #endif
         {
             SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
--- a/src/xrEngine/Device_Initialize.cpp
+++ b/src/xrEngine/Device_Initialize.cpp
@@ -48,10 +48,7 @@ void CRenderDevice::Initialize()
         const Uint32 flags = SDL_WINDOW_BORDERLESS | SDL_WINDOW_HIDDEN |
             SDL_WINDOW_RESIZABLE | SDL_WINDOW_OPENGL;
 
-  
-#if !defined(LINUX)
         if (psDeviceFlags.test(rsRGL))
-#endif
         {
             SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
             SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 4);
--- a/src/xrEngine/xr_ioc_cmd.cpp
+++ b/src/xrEngine/xr_ioc_cmd.cpp
@@ -563,12 +563,7 @@ public:
              * Since the Engine doesn't support switches between renderers
              * in runtime, it's safe to disable this command until restart.
              */
-
-            // keep only actual value
-            const auto& it = std::remove_if(VidQualityToken.begin(), VidQualityToken.end(),
-                [&](const auto& renderer) { return renderer.id != renderer_value; });
-
-            VidQualityToken[1].name = '\0';
+            Msg("Renderer is overrided by command line argument");
             return;
         }
 
--- a/src/xrEngine/xr_ioc_cmd.cpp
+++ b/src/xrEngine/xr_ioc_cmd.cpp
@@ -536,21 +536,21 @@ virtual void Save (IWriter *F) {};
 
 ENGINE_API BOOL r2_sun_static = TRUE;
 ENGINE_API BOOL r2_advanced_pp = FALSE; // advanced post process and effects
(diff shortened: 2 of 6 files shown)


src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -536,18 +536,33 @@ virtual void Save (IWriter *F) {};
...
 
-u32 renderer_value = 3;
-
-class CCC_r2 : public CCC_Token
+class CCC_renderer : public CCC_Token
 {
     typedef CCC_Token inherited;
 
+    u32 renderer_value = 3;
+    static BOOL cmd_lock;
```
Xottab-DUTY: No, please, remove this lock. It breaks the ability to change the renderer more than once.

src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -536,18 +536,33 @@ virtual void Save (IWriter *F) {};
 
 ENGINE_API BOOL r2_sun_static = TRUE;
 ENGINE_API BOOL r2_advanced_pp = FALSE; // advanced post process and effects
+ENGINE_API BOOL renderer_allow_override = FALSE;
```
Xottab-DUTY: Please, use `bool` instead of `BOOL`

src/xrEngine/Device_Initialize.cpp
```diff
@@ -50,8 +50,7 @@ void CRenderDevice::Initialize()
 
   
 #if !defined(LINUX)
-        // xxx: need to fix getting rsRGL flag, it doesn't work now
-        if (strstr(Core.Params, "-gl") || psDeviceFlags.test(rsRGL))
+        if (psDeviceFlags.test(rsRGL))
 #endif
```
Xottab-DUTY: Please, remove this `#if !defined(LINUX)` block. You have fixed the problem because of which it was here =)

Xottab-DUTY (changes requested):

src/xrEngine/xr_ioc_cmd.cpp
vTurbine: Actually no. The `cmd_lock` has effect only until `renderer_allow_override` set to `FALSE` (default value). If no `-rXX` cmd line options provided the `renderer` command will behave as usually. Please refer to added line 295 in `main.cpp` and statement at line 553 in `xr_ioc_cmd.cpp`.

In other words there are two possible scenarios:
1. Application has `-rXX` set and use it as only option (further renderer changes won't have effect)
2. No renderer overridings provided = regular case. The command works as is.

Xottab-DUTY: Ah! Then it's okay.

But, then we should lock the renderer dropbox in the options, so that user will know that he can't change it because cmd key is set.

Can you try to do this? It it's impossible with the current scripting abilities, then we should open an issue for that)

src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -560,6 +563,12 @@ class CCC_renderer : public CCC_Token
              * Since the Engine doesn't support switches between renderers
              * in runtime, it's safe to disable this command until restart.
              */
+
+            // keep only actual value
+            const auto& it = std::remove_if(VidQualityToken.begin(), VidQualityToken.end(),
+                [&](const auto& renderer) { return renderer.id != renderer_value; });
+
+            VidQualityToken[1].name = '\0';
```
Xottab-DUTY: Oh, no! I mean block it in the game scripts..

src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -581,16 +581,11 @@ class CCC_renderer : public CCC_Token
...
+        cmd_lock = true;
     }
 
-    virtual void Save(IWriter* F)
+    void Save(IWriter* F) override
     {
-        if (renderer_allow_override == FALSE)
-        {   // Do not save forced value
-            return;
-        }
```
Xottab-DUTY: Ah, why it was discarded? I found it very nice!

So, the final approach: can you return it and squash the commits? (although, squashing isn't necessary, this can be done with merge-button click)

src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -581,16 +581,11 @@ class CCC_renderer : public CCC_Token
...
         r2_advanced_pp = (renderer_value >= 3);
     
-        cmd_lock = TRUE;
+        cmd_lock = true;
     }
 
-    virtual void Save(IWriter* F)
+    void Save(IWriter* F) override
     {
-        if (renderer_allow_override == FALSE)
```
Xottab-DUTY: Ah, why it was discarded? I found it very nice!

So, the final approach: can you return it?

vTurbine: Both approaches are losing but if you like I'm putting it back J

src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -1,4 +1,7 @@
 #include "stdafx.h"
+
+#include <algorithm>
```
Xottab-DUTY: Oh, sorry, the last thing: unnecessary include, please remove it.

vTurbine: Yep, my fault

Xottab-DUTY (approved):

PR #351, opened 2019-02-08
Moving forward more portability with FreeBSD.

devnexen:


src/Common/PlatformLinux.inl
```diff
@@ -16,7 +16,13 @@
 #include <algorithm> // for min max
 
 #include <string>
+#if defined(LINUX)
 #include <alloca.h>
+#else
```
drug007: I would add FREEBSD explicitly

src/Common/PlatformLinux.inl
```diff
@@ -16,7 +16,13 @@
 #include <algorithm> // for min max
 
 #include <string>
+#if defined(LINUX)
```
eagleivg: It looks bad in this file. Perhaps it is better to copy the most contents of PlatformLinux.inl into a separate PlatformBSD.inl file?

src/xrCommon/inlining_macros.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-#if defined(__GNUC__)
+#if defined(__GNUC__) || defined(__clang__)
```
drug007: isn't it compiler wise? Some people try to add support for clang so probably it would be better to add as distinct PR?

src/xrCore/_math.cpp
```diff
@@ -15,9 +15,22 @@
 //?
 #endif
 
-#elif defined(LINUX)
+#elif defined(LINUX) || defined(FREEBSD)
 #include <x86intrin.h> // __rdtsc
+#ifdef LINUX
 #include <fpu_control.h>
+#else
```
drug007:
```suggestion
#elif defined(FREEBSD)
```
It makes code cleaner

src/xrCore/_math.cpp
```diff
@@ -15,9 +15,22 @@
...
+#include <sys/sysctl.h>
```
eagleivg, line 23: better define it in PlatformBSD.inl

src/xrCommon/inlining_macros.h
devnexen: Yes surely even though it s mainly because I had no much of a choice actually gcc for FreeBSD code generation crashes at the moment ... still to be fixed.

eagleivg (changes requested):

src/xrCore/_math.cpp
devnexen: all right then.

PR #352, opened 2019-02-10
Render interface: minor code cleanups

vTurbine:
* consistency: interface method `DestroyHW()` renamed to `Destroy()`
* `D3DXRenderBase`: not related to scene graph methods moved into separate file
* cleanup: redundant `virtual` specificators removed


Xottab-DUTY (changes requested):

It's a good practice to do one commit for one action.

Could you split your PR into three commits?

It's better to sort it in this order:
1. Layers/xrRender/D3DXRenderBase.h: replace virtual with override
2. Layers/xrRender/D3DXRenderBase.h: rename DestroyHW() -> Destroy()
   For consistency with Create()
3. Layers/xrRender/r__dsgraph_build.cpp: moved D3DXRenderBase methods into separate file

PR #356, opened 2019-02-13
xrEngine: fix game crash for window mode

eagleivg:
The problem was in the old configuration file, the monitors were numbered from 1, and now from 0. Added processing of this case.

### DIFF
--- a/src/xrEngine/MonitorManager.cpp
+++ b/src/xrEngine/MonitorManager.cpp
@@ -27,8 +27,15 @@ MonitorsManager::ResolutionPair MonitorsManager::GetMinimalResolution()
 MonitorsManager::ResolutionPair MonitorsManager::GetMaximalResolution()
 {
     const ResolutionsMap& resolutions = Monitors[Vid_SelectedMonitor];
-    const auto it = resolutions.crbegin();
-    return it->first;
+    if(!resolutions.empty())
+    {
+        const auto it = resolutions.crbegin();
+        return it->first;
+    }
+    else
+    {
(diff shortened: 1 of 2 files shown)


src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -433,9 +433,10 @@ class CCC_VidMonitor : public IConsole_Command
 
     void Execute(pcstr args) override
     {
-        int id;
-        sscanf(args, "%d. *", &id);
-        Vid_SelectedMonitor = id;
+        u32 id = 0;
+
+        if(1 == sscanf(args, "%u. *", &id))
```
Xottab-DUTY:
```suggestion
        if (1 == sscanf(args, "%u. *", &id))
```

src/xrEngine/MonitorManager.cpp
```diff
@@ -27,8 +27,15 @@ MonitorsManager::ResolutionPair MonitorsManager::GetMinimalResolution()
...
-    return it->first;
+    if(!resolutions.empty())
+    {
+        const auto it = resolutions.crbegin();
+        return it->first;
+    }
+    else
+    {
+        return ResolutionPair(640,480);
+    }
```
Xottab-DUTY, line 38: Well... This was wrong decision to change the order from one to zero.

Let's return it back then?

eagleivg: Ok, let`s do it. Close this PR then?

PR #357, opened 2019-02-13
xrEngine: setting of `rsGL` flag fixed

vTurbine:
Seems the previous fix was incomplete and this one solves the issue. At last.
Also some cosmetics applied.

### DIFF
--- a/src/xrEngine/main.cpp
+++ b/src/xrEngine/main.cpp
@@ -122,10 +122,10 @@ ENGINE_API void InitInput()
     pInput = new CInput(captureInput);
 }
 
-ENGINE_API void destroyInput() { xr_delete(pInput); }
+ENGINE_API void DestroyInput() { xr_delete(pInput); }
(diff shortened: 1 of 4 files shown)


src/xrEngine/main.cpp
```diff
@@ -298,6 +297,7 @@ ENGINE_API int RunApplication()
     else
         Console->Execute("renderer renderer_r1");
 
+    Device.Initialize();
```
Xottab-DUTY, line 301: It seems that if `CheckBenchmark()` will return true, then Device will be unitialized

Xottab-DUTY: But anyway I just fixed it by moving initialization stages without changing them. Sorry if this was impolite.

Xottab-DUTY:
Since the issue was fixed in b0a2128da0ec71109ddcad798fe28e9b8641f20d you can leave this as code style changing PR if you want.

vTurbine:
This looks better, thanks. I'm closing the PR

PR #361, opened 2019-02-23
GOG do not stop the game for missing content

devnexen:


src/Layers/xrRenderGL/glSH_Texture.cpp
```diff
@@ -211,7 +211,7 @@ void CTexture::Load()
             GLenum err = glGetError();
             if (err != GL_NO_ERROR)
             {
-                FATAL_F("Invalid video stream: 0x%x", err);
+                Msg("Invalid video stream: 0x%x", err);
```
Xottab-DUTY, line 214: Is it safe?

devnexen: I could play a long session at least ... in fact when launching a new party -> no intro -> the game stops whereas here get green screen but the game started at least.

PR #368, opened 2019-03-09
LAN game launch segfaults on Linux, fix proposal.

devnexen:


Xottab-DUTY:
@devnexen, can you describe what are you tried to fix?

Now, with this _fix_, the game crashes on Windows when I'm trying to open servers list..

devnexen:
What happens if you test getparent() in the condition ?

Xottab-DUTY:
It's wrong to check against GetParent() since parent can be null

collectioner:
@devnexen Relation between LAN and UI box is not clear. Also, we hadn't any issues related to the UI with MP on Windows. 

Agree with @Xottab-DUTY : can you describe the issue? Maybe we can fix it in other way.

devnexen:
It simply segfault one accepting the LAN settings, it was just to give more infos about the cause.

PR #369, opened 2019-03-10
Display current used CPU.

devnexen:


nitrocaster:
Before we get to some other considerations, why do we even need to know CPU usage at a single moment during startup? Including it in rs_stats output would be useful (it it is not there yet), but what does a single data sample tell you? I don't get the idea.

devnexen:
There is no use at startup indeed when I did the little dev I was just pushing it as a "debug" info but forgot to remove. Was interested in the end by the console.

Xottab-DUTY (approved):
Thank you for your contribution and sorry for the long-time review!

PR #371, opened 2019-03-18
Travis CI linux build fixes

Wedmer:
Replaced R_ASSERT with R_ASSERT3 as there're 3 arguments.
Changed CS_LIBRARY_NAME macro to use auto concatenated static strings (it's the same fix as in #370 but I found that later)


Xottab-DUTY:

Issue #362, opened 2019-02-25
[Linux] Use-after-free problem


eagleivg:
So, linux build can crash, due call `DestroyCharacter()` in `CAI_Stalker::wounded()` and subsequently iterate over `m_update_objects` and reading deleted `m_body`

Pastebin sanitizers output of problem: https://pastebin.com/axqP3ZqG

Save which cause problem (need wait while stalker will wounded): https://cdn.discordapp.com/attachments/410173937238933515/548175706052362263/iorlov_-_quicksave.scop)

eagleivg:
UPD: any kill can cause this crash: https://pastebin.com/Da7pvQDD

Xottab-DUTY:
Now, testers should test if the issue was fixed with 08c82f06fc73f17918b8edcc7e6583080d06d14b

eagleivg:
08c82f0 does not fix the problem. Although the direction, I think, is correct.

eagleivg:
The problem was solved by assembling the project via gcc-8

PR #385, opened 2019-04-10
ODE - asan detects underflow access with some saves.

devnexen:


Xottab-DUTY:
Can you give us savefiles to reproduce this?

devnexen:
Unrelated issue but I was able to reproduce as well with this attached save file https://github.com/OpenXRay/xray-16/issues/362

Xottab-DUTY:
So, what about save file?

devnexen:
There is one in the link above.

ForserX:
It's just not beautiful.

eagleivg:
Thank you for drawing attention to the problem. The proposed method is not suitable, since inactive joints will be processed. Fixed in https://github.com/OpenXRay/xray-16/commit/08c82f06fc73f17918b8edcc7e6583080d06d14b

PR #388, opened 2019-04-21
xrEngine: Console: Use SDL text input mode only when console is active

### LINKED ISSUE #355: Periodically incorrect input to console
Sometimes incorrect input to console occurs.
I found algorithm to stably get that bug:
1. Open the console, manually write "demo_record 1".
2. Press Enter.
3. Press W button (or D button or another).
4. Open the console, press "UP" button and then "1".
5. You will see W button in console (button you pressed in p.3). But "1" should have been displayed.

vTurbine:
This fixes #355

### DIFF
--- a/src/xrEngine/line_edit_control.cpp
+++ b/src/xrEngine/line_edit_control.cpp
@@ -124,6 +124,7 @@ void line_edit_control::clear_states()
     m_buf3[0] = 0;
 
     m_cur_pos = 0;
+    m_inserted_pos = 0;
     m_select_start = 0;
     m_p1 = 0;
     m_p2 = 0;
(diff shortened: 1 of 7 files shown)


Xottab-DUTY:
Oh, that's my bad. Much thanks!

By the way, can SDL TextInput replace our custom text editor? (everything in `text_editor` namespace)
If no, then do we even need to handle SDL TextInput?

vTurbine:
I bet usage of TI is incorrect in current implementation.
As I can see from code the `on_key_press` action can be called only on `SDL_KEYDOWN` event and then process to peeking  `SDL_TEXTINPUT` from queue. Two issues came here:
- if no TI event received we inserting a character in accordance to the button pressed (so why the TI was introduced then?)
- delayed TI event can be processed only _after_ key pressed down and this will likely add previously entered letter into string instead of actually pushed one.

In my particular case whole console input looks like hell so I prefer to get rid of TI

Xottab-DUTY:
> I bet usage of TI is incorrect in current implementation.

> In my particular case whole console input looks like hell so I prefer to get rid of TI

Ok, just comment it out for now. I'm interesting if it's possible to remove `text_editor` and provide correct TI implementation, because... It may be better? I'm not sure, just need to test it.

vTurbine:
Well, I have an idea how to fix this. Give me a day to apply

src/xrEngine/line_edit_control.cpp
```diff
@@ -364,7 +365,7 @@ void line_edit_control::assign_callback(int const dik, key_state state, Callback
     m_actions[dik]->on_assign(prev_action);
 }
 
-void line_edit_control::insert_character(char c) { m_inserted[0] = c; }
+void line_edit_control::insert_character(char c) { m_inserted[0] = c; } // FIXME: increment index or use vector-like container instead
```
vTurbine: In case when TI returns more that one character this will be broken

Xottab-DUTY: The character will be broken? (wrong character) Or the engine will be broken? 😃

Xottab-DUTY:
Awesome! So, now it's fully functional?

vTurbine:
Console works fine for me now. But please cross check it.

Xottab-DUTY:
Can you remove last merge commit? It's better to do a rebase, so that history will be clean.

vTurbine:
It was pushed accidentally. Now I'm not sure that force-push is exact thing you wanted. If it's better I can re-open this as a new PR.

Xottab-DUTY:
Using force push for pull requests is totally normal. It's better to force push, rather than opening a new PR.

Xottab-DUTY:
Thank you, ll works fine! I'm sorry it took so long to check.

PR #390, opened 2019-04-24
Small changes in gamepad support

justtails:
I added devices initialization and now you can bind gamepad buttons into Options->Controls. But sticks still can't be used.

### DIFF
--- a/src/xrEngine/xr_input_xinput.cpp
+++ b/src/xrEngine/xr_input_xinput.cpp
@@ -138,11 +138,11 @@ _action actions[] = {
 
 _keyboard keyboards[] =
 {
-    { "mouse1",                 MOUSE_1,                         "Mouse 1" },
-    { "mouse2",                 MOUSE_2,                         "Mouse 2" },
-    { "mouse3",                 MOUSE_3,                         "Mouse 3" },
-    { "mouse4",                 MOUSE_4,                         "Mouse 4" },
-    { "mouse5",                 MOUSE_5,                         "Mouse 5" },
+    { "mouse1",                 MOUSE_1,                         "LMB" },
+    { "mouse3",                 MOUSE_3,                         "MMB" },       // This is not a mistake cause init algorithm was changed.
+    { "mouse2",                 MOUSE_2,                         "RMB" },
+    { "mouse4",                 MOUSE_4,                         "Mouse X1" },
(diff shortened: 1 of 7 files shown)


src/xrGame/Level_input.cpp
```diff
@@ -112,7 +112,7 @@ extern float g_separate_radius;
 #include "xrScriptEngine/script_engine.hpp"
 #include "ai_space.h"
 
-void CLevel::IR_OnKeyboardPress(int key)
```
Xottab-DUTY: Please, revert this `key -> dik` and another related changes. DIK stands for Direct Input Key, but we don't use DirectInput anymore. And this change just increases diff size without adding something useful.

src/xrEngine/xr_input_xinput.cpp
```diff
@@ -138,11 +138,11 @@ _action actions[] = {
 
 _keyboard keyboards[] =
 {
-    { "mouse1",                 MOUSE_1,                         "Mouse 1" },
-    { "mouse2",                 MOUSE_2,                         "Mouse 2" },
-    { "mouse3",                 MOUSE_3,                         "Mouse 3" },
-    { "mouse4",                 MOUSE_4,                         "Mouse 4" },
-    { "mouse5",                 MOUSE_5,                         "Mouse 5" },
+    { "mouse1",                 MOUSE_1,                         "LMB" },
+    { "mouse3",                 MOUSE_3,                         "MMB" },       // This is not a mistake cause init algorithm was changed.
```
Xottab-DUTY, line 142: It won't break something, yes?
image

Xottab-DUTY (changes requested):

src/xrEngine/xr_input_xinput.cpp
justtails: I just give mouse buttons correct user friendly name. And yeap, I used some "hack" because mouse buttons in SDL and in vanilla xray have different positions in "key list". If you change this, I think the game will show Middle button like Right button, and vice versa.

src/xrGame/Level_input.cpp
justtails: I resolved.

Xottab-DUTY (approved):
Seems to be Ok

PR #393, opened 2019-04-28
ODE - Disabling linux-ism for FreeBSD

devnexen:


Xottab-DUTY:
Interesting fix! So, FreeBSD doesn't use `malloc.h`?

devnexen:
Not at all !

Xottab-DUTY:
@devnexen by the way, whats the progress on porting the engine to FreeBSD?

Issue #142, opened 2016-07-13
One source of "jerkyness" may have been identified


tamlin-mike:
Repro: Start a new game, run 30-or-so m.

Starting a new game in a debug build, while running under a profiler, I found out some lua script loaded a bunch of audio files - most if not all completely unrelated to the area I entered, and definitely not somethign expected to be played within the next seconds.

This made me realize the audio subsystem needs at least one, and maybe two, new functions.
1. preload.
   This one should be obvious. What may not be obvious is that it should run in another thread (think WorkItem's).
2. play_ambience.
   If added, this one should use the same preload (or even WorkQueue) mechanism, to not lock up the main game pointlessly.

revolucas:
I suppose you mean **sound_theme.script**? It loads all the character_voices defined in **script_sound.ltx** when a NPC comes online; assuming it matches the community for **avail_communities** field. Probably, this script needs rewritten to account for story_id, so that regular stalkers aren't needlessly setting up to play quest sounds.

Xottab-DUTY:
Solved with preload on precaching stage. (#407, da654d6c5ed221ab8762ab3991408bb8afd9d8d0 and a4920069f57ab1bca59800d41a2feaba4c80eb92)

Xottab-DUTY:
~~Duplicate of #62~~ Oh, no, it's not duplicate.

Related to #62

Xottab-DUTY:
Recently I've added parallel sound data preloading during the work on SteamAudio.

Now it makes sense to rework it more:
- [ ] Create sound sources in parallel on demand instead of precaching all sound sources at once (as it was introduced in #407).
This will reduce the RAM usage and speed-up game loading.

FYI: sound data is the actual PCM data from already created sound source, so sound source is obviously a different thing.

Issue #131, opened 2016-04-02
Game stuttering issue


ghost:
It seems like we've all tried just about everything to eliminate stuttering in Stalker. I think some of us felt that luajit/luabind being updated would fix this but it doesn't unfortunately.
However, I feel I know the main cause.

Stalker does not appear to stream in assets properly. A good example of this is in a mod like Call of Chernobyl, which uses a modified Open Xray engine. After adding "-dbg" to the shortcut, and when in the game going to the menu and pressing "S" to bring up the debug spawner, try to spawn an NPC.

What will happen every single time is as soon as you enter the game, it will hang for a bit as it loads the NPC, before the NPC is fully configured. Try spawning multiple NPC's at once and you'll get a longer wait time followed by some stuttering at first as the various NPC's are brought online.

Apparently GSC does not know how to properly stream in assets, and if you think about the fact that new elements are brought online whenever the switch distance is hit and something new is coming online, which is often, that is often the same time the game begins stuttering.

As for a solution, I am a beginning programmer so this may not be best something suited to me, but I do know I would like to help where I can as I know Nitrocaster and Armada already have a ton of great work that they're doing. I think we should try and recruit some fresh talent to work on Open Xray. My best guess would be trying out people on gamedev.net

So basically, if we could properly stream assets into the game rather than loading the entire asset in all at once, I feel stuttering could be an issue of the past.

ghost:
To add to this: I'm going to do a partial release of the code for "Radium" as a branch from Avotishin's branch of the repo. I am absolutely terrible with proper coding conventions, but I feel I have a lot of good stuff that should be added to Open Xray. So, once it is up, I hope you guys feel free to add it to the main repository after cleaning it up.

I would try a merge on my own, but due to the fact that I am not currently using the main branch, it would be pretty complicated.

alexgdi:
We need to do quite a lot of work to organize a good multithreading...

ghost:
I'm not talking about multithreading, I am talking about streaming assets. They have nothing to do with each other.

CrossVR:
Stuttering was never really a problem for me, but that's probably because I installed Stalker on my SSD.

TYoung86:
You say "I am a beginning programmer", and then you assert that "streaming" and "multithreading" have nothing to do with eachother... Perhaps you are unaware that streaming an asset will require dedicating a thread context to loading the asset?

[...]

PR #406, opened 2019-05-28
Speed up sound source list by using an unordered map

Automat-GH:
This avoids iterating the entire list whenever a new sound source is created (which was contributing to the stuttering mentioned in #142 and #131).

### DIFF
--- a/src/xrSound/SoundRender_Core.cpp
+++ b/src/xrSound/SoundRender_Core.cpp
@@ -86,8 +86,10 @@ void CSoundRender_Core::_clear()
     env_unload();
 
     // remove sources
-    for (u32 sit = 0; sit < s_sources.size(); sit++)
-        xr_delete(s_sources[sit]);
+    for (auto it = s_sources.begin(); it != s_sources.end(); ++it)
+    {
+        xr_delete(it->second);
+    }
     s_sources.clear();
 
     // remove emitters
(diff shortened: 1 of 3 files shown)


src/xrSound/SoundRender_Core.h
```diff
@@ -54,7 +55,7 @@ class CSoundRender_Core : public ISoundManager
     CDB::MODEL* geom_ENV;
 
     // Containers
-    xr_vector<CSoundRender_Source*> s_sources;
+    xr_unordered_map<std::string, CSoundRender_Source*> s_sources;
```
Xottab-DUTY: Please, don't use `std::` types, since we have our own. At least we should use `xr_string`... But, can you try to use `shared_str` here?

src/xrSound/SoundRender_Core.cpp
```diff
@@ -86,8 +86,10 @@ void CSoundRender_Core::_clear()
     env_unload();
 
     // remove sources
-    for (u32 sit = 0; sit < s_sources.size(); sit++)
-        xr_delete(s_sources[sit]);
+    for (auto it = s_sources.begin(); it != s_sources.end(); ++it)
+    {
+        xr_delete(it->second);
+    }
```
Xottab-DUTY, line 92: We can use range-based for here

src/xrSound/SoundRender_Core.cpp
```diff
@@ -592,9 +594,9 @@ void CSoundRender_Core::refresh_sources()
 {
     for (u32 eit = 0; eit < s_emitters.size(); eit++)
         s_emitters[eit]->stop(false);
-    for (u32 sit = 0; sit < s_sources.size(); sit++)
+    for (auto it = s_sources.begin(); it != s_sources.end(); ++it)
```
Xottab-DUTY: We can use range-based for here

Xottab-DUTY (changes requested):

PR #407, opened 2019-05-29
Added option to create all sound sources on startup

Automat-GH:
While researching issues #142 and #131, I used xperf to profile the stuttering that happens when a new AI comes online. During the stutter, most of the time is in CSoundRender_Core::i_create_source (and child calls).
This change adds a console variable that causes all sound sources to be created on startup, instead of being created when the sound is first needed. So it frontloads the work to make the actual play experience smoother. On my machine, on a release build, this adds 2.7 seconds to the startup time, but 90%+ of the stutter is eliminated when an AI comes online. These times will probably change depending on CPU and disk hardware.

### DIFF
--- a/src/xrEngine/xr_ioc_cmd.cpp
+++ b/src/xrEngine/xr_ioc_cmd.cpp
@@ -37,7 +37,7 @@ const xr_token constant_fps_token[] =
     { nullptr, -1 }
 };
 
-const xr_token snd_create_all_sources_token[] = {{"off", 0}, {"on", 1}, {nullptr, 0}};
+const xr_token snd_precache_all_token[] = {{"off", 0}, {"on", 1}, {nullptr, 0}};
 
 void IConsole_Command::InvalidSyntax()
 {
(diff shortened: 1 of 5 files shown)


Xottab-DUTY:
That's nice thank you! I think, that it's better to name it as `snd_precache_all`: this is a shorter name, but still clear about what it does.

src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -37,6 +37,8 @@ const xr_token constant_fps_token[] =
     { nullptr, -1 }
 };
 
+const xr_token snd_precache_all_token[] = {{"off", 0}, {"on", 1}, {nullptr, 0}};
```
Xottab-DUTY, line 40:
```suggestion
constexpr xr_token snd_precache_all_token[] = {{"off", 0}, {"on", 1}, {nullptr, -1}};
```

Xottab-DUTY: Oh, `constexpr` won't work here.

PR #409, opened 2019-06-03
More cross platform accurate file path length definitions

xrombik:


src/xrCore/_types.h
```diff
@@ -55,7 +55,15 @@ constexpr double dbl_min = type_min<double>;
...
-constexpr int max_path = 260;
+#ifdef LINUX
+constexpr int max_path = PATH_MAX;
+#elif FREEBSD
+constexpr int max_path = PATH_MAX;
+#elif WINDOWS
+constexpr int max_path = MAX_PATH;
+#else
+#error Unsupported platform
+#endif
```
Xottab-DUTY, line 64: `#elif WINDOWS` won't work as you expect! You should check if it's defined.

It should be like that:
```cpp
#if defined(LINUX) || defined(FREEBSD)
constexpr size_t max_path = PATH_MAX;
#elif defined(WINDOWS)
constexpr size_t max_path = MAX_PATH;
#else
#error Add your platform here
#endif
```
(I also replaced `int` with `size_t` because it can't be negative and mostly used as array size)

Xottab-DUTY (changes requested):

Xottab-DUTY:
Thank you!

PR #411, opened 2019-06-04
More using of referencies instead of values for Frect struct

xrombik:


src/xrUICore/Windows/UIFrameWindow.cpp
```diff
@@ -318,7 +318,7 @@ void CUIFrameWindow::draw_tile_line(Frect rect, int i, bool b_horz, Fvector2 con
     }
 }
 
-void CUIFrameWindow::draw_tile_rect(Frect rect, int i, Fvector2 const& ts)
+void CUIFrameWindow::draw_tile_rect(Frect& rect, int i, Fvector2 const& ts)
```
Xottab-DUTY, line 321: ARE YOU SURE that there's no potential error?

xrombik: Yes, i'm sure. The project is build and played normally. If you need to check something extra, I will check it out.

PR #421, opened 2019-06-20
[wip][testers needed][charconv] replace atof and atoi with from_chars

ZeeWanderer:
- src
  - [x] atof replaced
  - [x] atoi replaced

Notes:
- xrUICore, xrSE_Factory, ctool and tinyxml were ignored.
- xr_ini has some error handling for from_chars due to difference from atoi in treating invalid input, the performance cost is negligible.

### DIFF
--- a/src/Layers/xrRender/SH_Texture.cpp
+++ b/src/Layers/xrRender/SH_Texture.cpp
@@ -247,7 +247,8 @@ void CTexture::Load()
                 flags.seqCycles = TRUE;
                 _fs->r_string(buffer, sizeof(buffer));
             }
-            u32 fps = atoi(buffer);
+            u32 fps;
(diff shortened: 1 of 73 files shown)


375gnu:
`std::from_chars` is not a replacement for `atoi`/`atof`:

- for integers:
> "0x" or "0X" prefixes are not recognized for base 16
> only the minus sign is recognized (not the plus sign), and only for signed integer types of value. 
- the same for floats:
> the plus sign is not recognized outside of the exponent (only the minus sign is permitted at the beginning) 
> if fmt is std::chars_format::hex, the prefix "0x" or "0X" is not permitted (the string "0x123" parses as the value "0" with unparsed remainder "x123").

PR #422, opened 2019-06-27
Game version switcher

FreeZoneMods:
A way to use multiple versions of the game using the same base resources.


FreeZoneMods:
For those who want to test the feature in-game: https://yadi.sk/d/1LM7oECEfEmzaA

PR #431, opened 2019-07-16
xrCore: if VERIFY failed, then show "Verifying failed" message instea…

qweasdd136963:
…d of "Assertion failed".

### DIFF
--- a/src/xrCore/xrDebug_macros.h
+++ b/src/xrCore/xrDebug_macros.h
@@ -68,7 +68,7 @@
     {\
         static bool ignoreAlways = false;\
         if (!ignoreAlways && !(expr))\
-            xrDebug::Fail(ignoreAlways, DEBUG_INFO, #expr);\
+            xrDebug::Fail(ignoreAlways, DEBUG_INFO, #expr, "Verifying failed");\
     } while (false)
 #define VERIFY2(expr, desc)\
     do\


nitrocaster:
This makes no sense. VERIFY and R_ASSERT macros are used in the same context and do the same thing. A more consistent way would be to rename VERIFY to D_ASSERT.

Xottab-DUTY:
I think it's better not to rename, because VERIFY is more understandable name, that says that if it's failed then it's not so critical. (In the most cases, except cases like debug checks that pointer is not nullptr)

qweasdd136963:
I completely agreee with Xottab-DUTY. VERIFY denotes noncritical errors that can be skipped (in the most cases). And the message "Verrifying failed" allows programmer distinguish it from the ASSERT errors.

src/xrCore/xrDebug_macros.h
```diff
@@ -68,7 +68,7 @@
     {\
         static bool ignoreAlways = false;\
         if (!ignoreAlways && !(expr))\
-            xrDebug::Fail(ignoreAlways, DEBUG_INFO, #expr);\
+            xrDebug::Fail(ignoreAlways, DEBUG_INFO, #expr, "Verifying failed");\
```
Xottab-DUTY, line 71: Sorry for the delay in response.
According to "Assertion failed" message, it's better to say "Verification failed".
```suggestion
            xrDebug::Fail(ignoreAlways, DEBUG_INFO, #expr, "Verification failed");\
```

nitrocaster: "Assertion" is a well known thing whereas "verification" is something that makes no sense in this context. Like I said, VERIFY and R_ASSERT are basically the same and there is no need to introduce another message that just adds more confusion. It's not like you actually need to know whether the error was caused by VERIFY or R_ASSERT macro - you will have file name and line number anyway.

Xottab-DUTY: They are almost the same, but VERIFY doesn't work for Release and it indicates non-critical errors. (in the most cases)

But... If we say `Even failed VERIFY should be threaten as an error` then it's a good point to not to change the message and leave it as `assertion failed`.

PR #438, opened 2019-07-17
Add gamepad bind menu to UI

justtails:
Looks like that, but now a little better (old screenshot).

Also now you can bind multiple actions on one key. This fixes the inability to save the game on the F5 button.

### DIFF
--- a/res/gamedata/scripts/ui_mm_opt_controls.script
+++ b/res/gamedata/scripts/ui_mm_opt_controls.script
@@ -1,0 +1,32 @@
+class "opt_controls" (CUIWindow)
+
+function opt_controls:__init() super()
+end
+
+function opt_controls:__finalize()
+
+end
+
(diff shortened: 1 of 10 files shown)


res/gamedata/scripts/ui_mm_opt_controls.script
```diff
@@ -0,0 +1,33 @@
+class "opt_controls" (CUIWindow)
```
Xottab-DUTY, line 1: Add original file with separate commit and then commit your changes to it.

Xottab-DUTY (changes requested):

res/gamedata/scripts/ui_mm_opt_controls.script
justtails: I did It.

Xottab-DUTY (approved):

PR #449, opened 2019-08-05
xr_sha1 refactoring

kacejot:
- Renamed xr_sha256 to xr_sha1 due to the algorithm it uses.
- Fixed chunk size in hash calculation. Valid chunk size for SHA1 is 512 bits, not bytes.
- Performed minor refactoring.


kacejot:
Maybe renaming `sha_checksum_t` to `hash_t` makes sense...

src/xrCore/Crypto/xr_dsa_signer.cpp
```diff
@@ -6,37 +6,28 @@ xr_dsa_signer::xr_dsa_signer(u8 const p_number[crypto::xr_dsa::public_key_length
...
-    m_sha.start_calculate(data, data_size);
-    long sha_process_counter = 0;
-    while (!m_sha.continue_calculate())
-    {
-        yielder(sha_process_counter);
-        ++sha_process_counter;
-    }
-    return m_dsa.sign(m_private_key, m_sha.pointer(), m_sha.digest_length);
+    auto hash = m_sha.calculate(data, data_size);
+    return m_dsa.sign(m_private_key, hash.data(), crypto::xr_sha1::DIGEST_SIZE);
```
wizir6: Maybe we should use upper case only for macros
@Xottab-DUTY

src/xrCore/Crypto/xr_sha.h
```diff
@@ -1,35 +1,22 @@
...
-    u32 m_data_size;
-
-    u8 m_result[digest_length];
-    CryptoPP::SHA1 m_sha_ctx;
-}; // xr_sha256
-
-} // namespace crypto
+    // m_buf is added as a field, because in this case it will be 
+    // allocated on the stack where xr_sha1 instance was created
+    sha_checksum_t m_buf;
```
wizir6: You can return buffer as out parameter

kacejot: Yeah, I thought about it.
In fact, it makes no difference - there is danger to have invalid reference anyway. But out parameter makes this more obvious, so I will apply these changes.

src/xrCore/Crypto/xr_dsa_signer.cpp
```diff
@@ -6,37 +6,28 @@ xr_dsa_signer::xr_dsa_signer(u8 const p_number[crypto::xr_dsa::public_key_length
...
 }
 
 shared_str const xr_dsa_signer::sign_mt(u8 const* data, u32 data_size, sha_process_yielder yielder)
 {
-    m_sha.start_calculate(data, data_size);
-    long sha_process_counter = 0;
-    while (!m_sha.continue_calculate())
-    {
-        yielder(sha_process_counter);
```
kacejot, line 36: I forgot to apply yielder in my refactoring. Will do.

kacejot (commented):
I have finished my edits

src/xrCore/Crypto/xr_sha.h
kacejot: Done

src/xrCore/Crypto/xr_dsa_verifyer.cpp
```diff
@@ -11,17 +11,17 @@ xr_dsa_verifyer::xr_dsa_verifyer(u8 const p_number[crypto::xr_dsa::public_key_le
 }
 
 xr_dsa_verifyer::~xr_dsa_verifyer() {}
-bool xr_dsa_verifyer::verify(u8 const* data, u32 data_size, shared_str const& dsign)
+std::pair<bool, const crypto::xr_sha1::hash_t&> xr_dsa_verifyer::verify(u8 const* data, u32 data_size, shared_str const& dsign)
```
wizir6: return reference to local variable

kacejot: Okay, hash should be copied

kacejot:
I will return from vacation and perform dev-testing, guys. Don't close this, please

Xottab-DUTY:
@kacejot, so, have you returned from vacation?))

kacejot:
@Xottab-DUTY, oh shi~.
Of course! I've completely forgotten, sry

kacejot:
@Xottab-DUTY, I have finished the changes, please review.
Could you please give some use cases, so I could check my changes.
Also, the good things is to cover this logic with tests. Have we some testing frameworks in our project?

Xottab-DUTY:
> Also, the good things is to cover this logic with tests. Have we some testing frameworks in our project?

No test performed at all. Yes, we need some framework for tests.. and tests themselves.

> Could you please give some use cases, so I could check my changes.

I don't know a good use case. But all of them are in multiplayer:
- Player config dump – unknown if it works or broken.
- Stealthy (async) screenshots – seems to be broken. (more info in #472)
- GameSpy profile data – achievements, etc.. This case should be good, because *probably* GSC GameSpy won't let you in if something is broken. *(not sure)*
Not sure because original GameSpy was closed and I don't know if the GSC server stores achievements.

src/utils/mp_configs_verifyer/configs_dump_verifyer.cpp
```diff
@@ -40,7 +40,7 @@ static char* search_info_section(u8* buffer, u32 buffer_size)
     return nullptr;
 }
 
-bool const configs_verifyer::verify_dsign(u8* data, u32 data_size, sha_checksum_t& sha_checksum)
+bool const configs_verifyer::verify_dsign(u8* data, u32 data_size, crypto::xr_sha1::hash_t& sha_checksum)
 {
     char* tmp_info_sect = search_info_section(data, data_size);
```
dr-Fade, line 45: char -> u8, for more consistency.

kacejot: Will do in configs_verifyer refactoring

src/utils/mp_configs_verifyer/configs_dump_verifyer.cpp
```diff
@@ -40,7 +40,7 @@ static char* search_info_section(u8* buffer, u32 buffer_size)
     return nullptr;
 }
 
-bool const configs_verifyer::verify_dsign(u8* data, u32 data_size, sha_checksum_t& sha_checksum)
+bool const configs_verifyer::verify_dsign(u8* data, u32 data_size, crypto::xr_sha1::hash_t& sha_checksum)
```
dr-Fade, line 43: u8* -> const u8* const data, since data does not suppose to change in the function.

src/utils/mp_configs_verifyer/configs_dump_verifyer.h
```diff
@@ -21,8 +21,7 @@ class configs_verifyer
     bool const verify(u8* data, u32 data_size, string256& diff);
 
 private:
-    typedef u8 sha_checksum_t[crypto::xr_sha256::digest_length];
-    bool const verify_dsign(u8* data, u32 data_size, sha_checksum_t& sha_checksum);
+    bool const verify_dsign(u8* data, u32 data_size, crypto::xr_sha1::hash_t& sha_checksum);
```
dr-Fade, line 24: u8* -> const u8* const data

src/xrGame/configs_dump_verifyer.h
```diff
@@ -21,8 +21,7 @@ class configs_verifyer
     bool const verify(u8* data, u32 data_size, string256& diff);
 
 private:
-    typedef u8 sha_checksum_t[crypto::xr_sha256::digest_length];
-    bool const verify_dsign(u8* data, u32 data_size, sha_checksum_t& sha_checksum);
+    bool const verify_dsign(u8* data, u32 data_size, crypto::xr_sha1::hash_t& sha_checksum);
```
dr-Fade, line 24: u8* -> const u8* const data
Const, for the god of const!

src/xrGame/screenshots_writer.cpp
```diff
@@ -33,7 +33,7 @@ void writer::set_player_cdkey_digest(shared_str const& cdkeydigest)
     m_info_data.w_string(ss_info_secion, ss_admin_name_key, admin_name.c_str());
 }*/
 
-u32 const writer::write_info(sha_process_yielder* yielder)
+u32 const writer::write_info(crypto::yielder_t* yielder)
```
dr-Fade, line 36: weak_ptr.

src/xrGame/screenshots_writer.h
```diff
@@ -33,7 +33,7 @@ class writer
     void set_player_cdkey_digest(shared_str const& cdkeydigest);
     // void	set_admin_name			(shared_str const & admin_name);
 
-    u32 const write_info(sha_process_yielder* yielder = NULL);
+    u32 const write_info(crypto::yielder_t* yielder = NULL);
```
dr-Fade, line 36: weak_ptr.

dr-Fade (approved):

src/xrGame/screenshots_writer.cpp
kacejot: No, weak_ptr requires shared_ptr to own the yielder. shared_ptr has big runtime.
It is better to be sure that yielder will never outlives its caller. As far as writer::write_info calls in the same thread as its caller, we are safe.

Maybe you would like something like this:
```suggestion
u32 const writer::write_info(const crypto::yielder_t& yielder)
```

src/xrGame/screenshots_writer.h
kacejot: The same thing

src/xrGame/configs_dump_verifyer.h
kacejot: Will do

src/utils/mp_configs_verifyer/configs_dump_verifyer.h
kacejot: Will do

src/utils/mp_configs_verifyer/configs_dump_verifyer.cpp
kacejot: Okay

src/xrCore/Crypto/xr_sha.cpp
```diff
@@ -1,24 +1,32 @@
 #include "stdafx.h"
 #include "xr_sha.h"
 
-namespace crypto
-{
+using namespace crypto;
 
-const xr_sha1::sha_checksum_t& xr_sha1::calculate(const u8* data, u32 data_size)
+const xr_sha1::sha_checksum_t& xr_sha1::calculate(const u8* data, u32 data_size, std::optional<sha_process_yielder> yielder)
```
nitrocaster: Why not take a pointer to `sha_process_yielder` here? `std::optional` is good for returning an optional value, but in this case it's rather redundant.

src/xrCore/Crypto/xr_sha.cpp
```diff
@@ -1,24 +1,32 @@
...
+            yielder.value()(chunks_processed);
+            ++chunks_processed;
+        }
     }
     m_buf.fill(0);
     sha_ctx.Final(m_buf.data());
     return { m_buf };
 }
 
+const inline xr_sha1::sha_checksum_t& xr_sha1::calculate(const u8* data, u32 data_size)
```
nitrocaster: What is the purpose of `inline` here? And why does this function exist to begin with, if one can just assign `yielder` to be `std::nullopt` or `nullptr`?

src/xrCore/Crypto/xr_sha.h
```diff
@@ -1,22 +1,29 @@
...
 #include "xrCore/xrCore.h"
 #include <cryptopp/sha.h>
+
+#include <optional>
 #include <array>
 
 namespace crypto
 {
+
+using sha_process_yielder = fastdelegate::FastDelegate1<long>;
```
nitrocaster: Yielder should probably take `size_t` as an argument.

src/xrCore/Crypto/xr_dsa_signer.cpp
nitrocaster: Upper case is indeed only for macros. This has to be fixed.

src/xrGame/configs_dump_verifyer.h
nitrocaster: What is the point of returning a *const pointer to a const value* instead of just a *pointer to a const value*?

src/xrGame/configs_dump_verifyer.cpp
```diff
@@ -197,20 +197,17 @@ bool const configs_verifyer::verify(u8* data, u32 data_size, string256& diff)
 
     m_orig_config_body.w_stringZ(add_str);
 
-    crypto::xr_sha256 tmp_sha_checksum;
-    tmp_sha_checksum.start_calculate(m_orig_config_body.pointer(), m_orig_config_body.tell());
-    while (!tmp_sha_checksum.continue_calculate())
-    {
-    };
+    crypto::xr_sha1::hash_t hash{};
```
nitrocaster: Given that 'crypto::xr_sha1::calculate()' is now a single self-contained function, you can return the hash from it without passing it by reference.

src/xrGame/configs_dump_verifyer.cpp
```diff
@@ -197,20 +197,17 @@ bool const configs_verifyer::verify(u8* data, u32 data_size, string256& diff)
...
+    crypto::xr_sha1::calculate(hash, m_orig_config_body.pointer(), m_orig_config_body.tell());
 
-    u8 tmp_checksum[crypto::xr_sha256::digest_length];
+    crypto::xr_sha1::hash_t tmp_checksum{};
```
nitrocaster: Zero-initializing of `tmp_checksum` is unnecessary here because its value is used only when `verify_dsign()` returns `true`.

src/xrGame/gsc_dsigned_ltx.cpp
```diff
@@ -75,10 +75,10 @@ bool gsc_dsigned_ltx_reader::load_and_verify(u8* buffer, u32 const size)
     *dsign_section = 0;
     xr_strcat(dsign_section, dsign_sect_size, ltx_date.c_str());
     u32 new_size = size - dsign_sect_size + ltx_date.size() + 1; // xr_strlen(buffer)
+
```
nitrocaster: You removed redundant braces around one-line block (which is fine), but then inserted two empty lines. Vertical space is still wasted, but the diff is now larger. I suggest you either undo this completely or not remove 2 new empty lines.

src/xrCore/Crypto/xr_dsa_signer.cpp
kacejot: Okay

src/xrGame/gsc_dsigned_ltx.cpp
kacejot: I will remove empty lines here

src/xrGame/configs_dump_verifyer.cpp
kacejot: hash_t is an std::array which is a class type. It will be default-initialized whether with curved brackets or not.

src/xrGame/configs_dump_verifyer.h
kacejot: @nitrocaster, `data` var should not be changed in `verify_dsign`, so `data` should be a pointer to const.

src/xrCore/Crypto/xr_sha.cpp
kacejot: The main purpose is to rid of raw pointers that can become dangling any moment.
Yes, std::optional creates a copy of yielder on its creation, but it is safer.
I can return it to a pointer if you want.

Also there is no need in 2 `calculate` functions. `yielder` could be a default parameter.

src/xrCore/Crypto/xr_sha.cpp
kacejot: Yes I have already thought about default parameter. I just missed this thing, yes, I will add it.
About `inline`: there is no need in this keyword here. I will remove it.

src/xrGame/configs_dump_verifyer.cpp
kacejot: Will do

src/xrGame/configs_dump_verifyer.h
kacejot: I decided not to change this part of code, because it is from another project.

src/xrCore/Crypto/xr_sha.h
kacejot: The whole code in Engine uses long. So this change has a large impact. I have tried to do this, but it should be done in the context of the separate task.

src/utils/mp_configs_verifyer/configs_dump_verifyer.h
kacejot: I decided not to change this part of code, because it is from another project.

src/utils/mp_configs_verifyer/configs_dump_verifyer.cpp
kacejot: I decided not to change this part of code, because it is from another project.

kacejot:
@nitrocaster 

I have done some changes due to your comments.
I didn't add an optional yielder parameter to calculate because:
- Optional parameters are bad tone. It is better to have two named functions for clearness.
- There is no need to check yielder for nullptr every calculation iteration. I pass the empty yielder that will be optimized to nothing instead of that. 

Also I have documented the code with the comments. I have checked it with doxygen.

Xottab-DUTY:
@nitrocaster any comments?

src/xrGame/gsc_dsigned_ltx.cpp
```diff
@@ -75,10 +75,8 @@ bool gsc_dsigned_ltx_reader::load_and_verify(u8* buffer, u32 const size)
     *dsign_section = 0;
     xr_strcat(dsign_section, dsign_sect_size, ltx_date.c_str());
     u32 new_size = size - dsign_sect_size + ltx_date.size() + 1; // xr_strlen(buffer)
-
```
Xottab-DUTY: ~~Please, return that empty line.~~
Ehm, no, don't return it)

*(https://github.com/OpenXRay/xray-16/pull/449#discussion_r354068105)*

src/xrCore/Crypto/xr_dsa.cpp
```diff
@@ -15,17 +15,17 @@ xr_dsa::xr_dsa(u8 const p[public_key_length], u8 const q[private_key_length], u8
...
     CryptoPP::Integer exp(priv_key.m_value, sizeof(priv_key.m_value));
     CryptoPP::DSA::PrivateKey private_key;
     private_key.Initialize(m_dsa, exp);
 
     std::string signature;
     CryptoPP::DSA::Signer signer(private_key);
-    CryptoPP::StringSource(data, data_size, true, new CryptoPP::SignerFilter(m_rng, signer,
-                                                      new CryptoPP::StringSink(signature)) // SignerFilter
-        ); // StringSource
+    CryptoPP::StringSource(data, data_size, true,\
```
Xottab-DUTY: No need for that `\`

Xottab-DUTY (changes requested):

src/xrGame/gsc_dsigned_ltx.cpp
```diff
@@ -75,10 +75,8 @@ bool gsc_dsigned_ltx_reader::load_and_verify(u8* buffer, u32 const size)
     *dsign_section = 0;
     xr_strcat(dsign_section, dsign_sect_size, ltx_date.c_str());
     u32 new_size = size - dsign_sect_size + ltx_date.size() + 1; // xr_strlen(buffer)
-    
     if (!verify(buffer, new_size, ltx_dsign))
         return false;
-
```
Xottab-DUTY: Please, return that empty line.

src/xrGame/gsc_dsigned_ltx.cpp
Xottab-DUTY: I don't see a problem if the diff will be larger for just two lines. But without them, code readability will suffer.

Xottab-DUTY: Code was somehow readable with the braces, but now, without even a single empty space, it's more unhandy to read the code.

src/xrCore/Crypto/xr_dsa.cpp
Xottab-DUTY: What's even the point of it?

src/xrGame/gsc_dsigned_ltx.cpp
```diff
@@ -76,9 +76,7 @@ bool gsc_dsigned_ltx_reader::load_and_verify(u8* buffer, u32 const size)
     xr_strcat(dsign_section, dsign_sect_size, ltx_date.c_str());
     u32 new_size = size - dsign_sect_size + ltx_date.size() + 1; // xr_strlen(buffer)
     if (!verify(buffer, new_size, ltx_dsign))
-    {
         return false;
-    }
```
Xottab-DUTY, line 81: It's better to insert one empty line here, after `return false;` (but not before `if (!verify(...))`)

Xottab-DUTY:
@kacejot hello! What's the progress on the pull request?

kacejot:
@Xottab-DUTY, hi! I'am missing testing environment at the moment. Do not merge this PR, until my changes tested, please.

Xottab-DUTY:
@kacejot, any news? 🙂

kacejot:
@Xottab-DUTY, nothing for now. I plan to close these comments on the weekend.

src/xrGame/configs_dump_verifyer.cpp
Xottab-DUTY: What's the point of having brackets then? If it will be initialized in any way, then there's no need to have brackets.
```suggestion
    crypto::xr_sha1::hash_t tmp_checksum;
```

src/xrCore/Crypto/xr_dsa_verifyer.cpp
```diff
@@ -11,17 +11,16 @@ xr_dsa_verifyer::xr_dsa_verifyer(u8 const p_number[crypto::xr_dsa::public_key_le
...
 #ifdef DEBUG
     IWriter* verify_data = FS.w_open("$logs$", "verify");
     verify_data->w(data, data_size);
     verify_data->w_string("sha_checksum");
-    verify_data->w(m_sha.pointer(), m_sha.digest_length);
+    verify_data->w(hash.data(), crypto::xr_sha1::DigestSize);
     FS.w_close(verify_data);
 #endif
-    return m_dsa.verify(m_public_key, m_sha.pointer(), m_sha.digest_length, dsign);
+    bool success = m_dsa.verify(m_public_key, hash.data(), crypto::xr_sha1::DigestSize, dsign);
```
Xottab-DUTY:
```suggestion
    const bool success = m_dsa.verify(m_public_key, hash.data(), crypto::xr_sha1::DigestSize, dsign);
```

src/xrGame/configs_dump_verifyer.cpp
```diff
@@ -196,21 +196,16 @@ bool const configs_verifyer::verify(u8* data, u32 data_size, string256& diff)
         tmp_ini.r_string(cd_info_secion, cd_player_digest_key), tmp_ini.r_string(cd_info_secion, cd_creation_date));
 
     m_orig_config_body.w_stringZ(add_str);
+    auto hash = crypto::xr_sha1::calculate(m_orig_config_body.pointer(), m_orig_config_body.tell());
```
Xottab-DUTY, line 199: I'd replaced all that `auto` with explicit `crypto::xr_sha1::hash_t`, but the `auto` is a place for possible compiler optimizations, so, probably, `auto` is better.

Xottab-DUTY (changes requested):
I just reviewed the PR from scratch and found some details that I didn't paid attention previously. The final review part.

Xottab-DUTY (approved):

src/xrGame/configs_dump_verifyer.cpp
Xottab-DUTY: @kacejot, any post-merge thoughts?)

Xottab-DUTY:
Thank you!

PR #451, opened 2019-08-13
Fix engine initialization issues in Linux

Zegeri:
Certain changes in b21c156bcb51 and some faults in the Event class definition for Linux introduced some race conditions in the initialization, that caused random crashes when starting the engine. This pull request fixes them

### DIFF
--- a/src/xrEngine/device.cpp
+++ b/src/xrEngine/device.cpp
@@ -175,13 +175,13 @@ void CRenderDevice::PrimaryThreadProc(void* context)
     GEnv.Render->MakeContextCurrent(false);
     device.deviceCreated.Set();
 
-    device.deviceReadyToRun.Wait(); // ping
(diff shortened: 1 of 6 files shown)


src/xrEngine/device.cpp
```diff
@@ -520,10 +520,10 @@ void CRenderDevice::Run()
 
     // Pre start
     GEnv.Render->MakeContextCurrent(false);
-    deviceReadyToRun.Set();  // ping
+    deviceReadyToRun.Set();
 
-    while (!deviceReadyToRun.Wait(MaximalWaitTime))
```
Xottab-DUTY, line 525: Oh.. So, it pings and receives pong from itself..

Reasonable. I will remember that, to not to repeat such a mistake.

Xottab-DUTY (approved):
Thank you very much! This problem has been bothering us for a long time.

PR #454, opened 2019-08-27
Now timestamp is cross-platform (use strftime)

eagleivg:
Rewrote the function for time stamping (so that you can take screenshots in Linux)

### DIFF
--- a/src/xrCore/_std_extensions.cpp
+++ b/src/xrCore/_std_extensions.cpp
@@ -13,29 +13,11 @@ int xr_strcmp(const char* S1, const char* S2)
 
 char* timestamp(string64& dest)
 {
-    string64 temp;
+    time_t     now = time(nullptr);
+    struct tm  tstruct;
+    tstruct = *localtime(&now);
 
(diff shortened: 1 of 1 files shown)


src/xrCore/_std_extensions.cpp
```diff
@@ -13,29 +13,11 @@ int xr_strcmp(const char* S1, const char* S2)
...
+    tstruct = *localtime(&now);
 
-    /* Set time zone from TZ environment variable. If TZ is not set,
-    * the operating system is queried to obtain the default value
-    * for the variable.
-    */
-#if defined(WINDOWS)
-    _tzset();
-    u32 it;
+    strftime(dest, sizeof(dest), "%Y-%m-%d_%H-%M-%S", &tstruct);
```
Xottab-DUTY: It is better to use day-month-year format
```suggestion
    strftime(dest, sizeof(dest), "%m-%d-%y_%H-%M-%S", &tstruct);
```

Xottab-DUTY (approved):
All is good. But it's better to use original US date format.

Xottab-DUTY (approved):

PR #456, opened 2019-09-01
make timestamp() function cross-platform and thread-safe

ulyachka:
fix for https://github.com/OpenXRay/xray-16/commit/e6402cdefe35b0b852605d18c1af11a58f973460 and https://github.com/OpenXRay/xray-16/commit/42ef1ccc58345f4935ee0d3e6de4f5c504f155eb

### DIFF
--- a/src/xrCore/_std_extensions.cpp
+++ b/src/xrCore/_std_extensions.cpp
@@ -15,7 +15,12 @@ char* timestamp(string64& dest)
 {
     time_t     now = time(nullptr);
     struct tm  tstruct;
-    tstruct = *localtime(&now); // localtime is NOT thread-safe
+
+#if defined(WINDOWS)
+    localtime_s(&tstruct, &now);// thread-safe for windows
+#else
+    localtime_r(&now, &tstruct);// thread-safe for posix systems
+#endif
 
     strftime(dest, sizeof(dest), "%m-%d-%y_%H-%M-%S", &tstruct);
 
--- a/src/xrCore/_std_extensions.h
+++ b/src/xrCore/_std_extensions.h
@@ -265,7 +265,6 @@ inline void MemFill32(void* dst, u32 value, size_t dstSize)
         *ptr++ = value;
 }
 
-// This function is NOT thread-safe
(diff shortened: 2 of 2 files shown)


src/xrCore/_std_extensions.cpp
```diff
@@ -15,7 +15,12 @@ char* timestamp(string64& dest)
 {
     time_t     now = time(nullptr);
     struct tm  tstruct;
-    tstruct = *localtime(&now); // localtime is NOT thread-safe
+
+#if defined(WINDOWS)
+    localtime_s(&tstruct, &now);// thread-safe for windows
+#else
+    localtime_r(&now, &tstruct);// thread-safe for posix systems
```
Xottab-DUTY, line 22: We can also use localtime_s for POSIX. But it's not thread-safe there, right?

src/xrCore/_std_extensions.cpp
```diff
@@ -15,7 +15,12 @@ char* timestamp(string64& dest)
 {
     time_t     now = time(nullptr);
     struct tm  tstruct;
-    tstruct = *localtime(&now); // localtime is NOT thread-safe
+
+#if defined(WINDOWS)
+    localtime_s(&tstruct, &now);// thread-safe for windows
```
Xottab-DUTY, line 20: According to documentation, it's recommended to define `__STDC_WANT_LIB_EXT1__` in `Common/PlatformWindows.inl`
https://en.cppreference.com/w/c/chrono/localtime

src/xrCore/_std_extensions.cpp
ulyachka: No, POSIX doesn't provide the localtime_s funtion.
localtime_s is a windows implementation for localtime_r.

https://pubs.opengroup.org/onlinepubs/009695399/functions/localtime.html

Xottab-DUTY: Ah, ok.

src/xrCore/_std_extensions.cpp
Xottab-DUTY: Ah, we don't need it for Windows.

PR #459, opened 2019-09-04
Ability to unload all weapons in inventory via context menu.

Dx74rghs:
Functionality to unload all ammo from available weapons in the backpack.


Xottab-DUTY (changes requested):
I can't accept this PR as is. This introduces gamedata changes, which we try to keep close to vanilla as possible.

You can do it another way: try to add a context menu entry with string "Unload all". Yes, it's not a hotkey, but this is a good compromise between introducing new features and being close to vanilla as possible.

Also, do not put the translations into original `ui_st_keybinding.xml`, instead use `openxray.xml`.

Dx74rghs:
added as context menu item

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you!

PR #465, opened 2019-09-08
CHW: WSI surface parameters encapsulated

vTurbine:
This removes implementation specific access to rendering surface dimensions.

### DIFF
--- a/src/Layers/xrRenderDX10/dx10HW.cpp
+++ b/src/Layers/xrRenderDX10/dx10HW.cpp
@@ -449,7 +449,8 @@ void CHW::UpdateViews()
 
 std::pair<u32, u32> CHW::GetSurfaceSize() const
 {
-    return {
+    return
+    {
         m_ChainDesc.BufferDesc.Width,
         m_ChainDesc.BufferDesc.Height
     };
--- a/src/Layers/xrRenderDX10/dx10HW.cpp
+++ b/src/Layers/xrRenderDX10/dx10HW.cpp
@@ -446,3 +446,11 @@ void CHW::UpdateViews()
 
     _RELEASE(pDepthStencil);
 }
+
+SIZE CHW::GetSurfaceSize() const
+{
+    return {
+        LONG(m_ChainDesc.BufferDesc.Width),
+        LONG(m_ChainDesc.BufferDesc.Height)
+    };
+}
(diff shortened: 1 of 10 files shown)


Xottab-DUTY (changes requested):
I'm not sure, that `SIZE` is Linux compatible. But even if it is, consider to use std::pair and C++17 syntax:
```cpp
class HW
{
	...
	std::pair<u32, u32> GetSurfaceSize() const;
	...
};
...

auto [width, height] = HW.GetSurfaceSize();
```

Xottab-DUTY:
Yes, according to Travis CI, type `SIZE` is Windows-only, which is not suitable for us.

vTurbine:
Expected smth like this. But saw GL renderer using ‘HDC’

vTurbine:
The master is broken

src/Layers/xrRenderDX10/dx10HW.cpp
```diff
@@ -446,3 +446,11 @@ void CHW::UpdateViews()
 
     _RELEASE(pDepthStencil);
 }
+
+std::pair<u32, u32> CHW::GetSurfaceSize() const
+{
+    return {
```
Xottab-DUTY: `{` should be on the next line.

src/Layers/xrRenderGL/glHW.cpp
```diff
@@ -187,3 +187,11 @@ HRESULT CHW::Present(UINT /*SyncInterval*/, UINT /*Flags*/)
     SDL_GL_SwapWindow(m_hWnd);
     return S_OK;
 }
+
+std::pair<u32, u32> CHW::GetSurfaceSize() const
+{
+    return {
```
Xottab-DUTY: `{` should be on the next line.

src/Layers/xrRenderDX9/dx9HW.cpp
```diff
@@ -378,3 +378,16 @@ BOOL CHW::support(D3DFORMAT fmt, DWORD type, DWORD usage)
         return FALSE;
     return TRUE;
 }
+
+std::pair<u32, u32> CHW::GetSurfaceSize() const
+{
+    return {
```
Xottab-DUTY: `{` should be on the next line.

Xottab-DUTY (approved):

PR #466, opened 2019-09-10
Build fixed after profiler resurrection

vTurbine:


src/utils/xrAI/factory_api.h
```diff
@@ -3,12 +3,12 @@
 #include "xrServerEntities/xrServer_Object_Base.h"
 
 extern "C" {
-typedef XR_IMPORT IServerEntity* __stdcall Factory_Create(LPCSTR section);
-typedef XR_IMPORT void __stdcall Factory_Destroy(IServerEntity*&);
+using CreateEntity = XR_IMPORT IServerEntity* __stdcall(LPCSTR section);
+using DestroyEntity = XR_IMPORT void __stdcall(IServerEntity*&);
```
Xottab-DUTY, line 7: This is unnecessary and not related to this PR. Also, probably in future, xrSE_Factory will be refactored, so, this change can cause conflicts or just unnecessary changes.

Xottab-DUTY (changes requested):
Thank you for fast reaction on what I broke. 😄 
I ask you to remove typedefs changes in `xrAI/factory_api.h` for reasons I described in my comment to this file.

Also, sorry for the delay in reviewing this PR. I was far from home and couldn't answer.

src/utils/xrAI/factory_api.h
vTurbine: Actually related. After addind xrEngine in the scope you get collision for Factory_Create/Destroy names. This is a fix for that.

vTurbine: This time is my turn to be AFK for some time so please manage this PR by yourself or drop it in case of better solution. It was just a quick fix

Xottab-DUTY: Oh.. I didn't think about it. Ok, then!

Xottab-DUTY: > This time is my turn to be AFK for some time so please manage this PR by yourself or drop it in case of better solution. It was just a quick fix

Ok! Just renamed second commit to somewhat better name.

PR #471, opened 2019-09-24
HW: presentation and surface state encapsulated

vTurbine:


src/Layers/xrRender/D3DXRenderBase.cpp
```diff
@@ -327,23 +299,7 @@ void D3DXRenderBase::End()
         overdrawEnd();
     RCache.OnFrameEnd();
     DoAsyncScreenshot();
```
Xottab-DUTY, line 301: Previously, DoAsyncScreenshot() was called before HW.pDevice->EndScene(). Now it's contrariwise.

I think, (just in case) it's better to create two new functions for `CHW`: Begin and End. In future, these functions may be useful for some purpose or be redundant at all after DX9 removal.

src/Layers/xrRenderGL/glHW.h
```diff
@@ -26,8 +26,8 @@ class CHW
     void ClearRenderTargetView(GLuint pRenderTargetView, const FLOAT ColorRGBA[4]);
 
     void ClearDepthStencilView(GLuint pDepthStencilView, UINT ClearFlags, FLOAT Depth, UINT8 Stencil);
-
```
Xottab-DUTY, line 29: It's better to not delete this whitespace.

Xottab-DUTY (commented):
If you can, please squash two your commits into one.

src/Layers/xrRender/D3DXRenderBase.cpp
vTurbine: No difference. The ‘EndScene’ submits batch while ‘GetRenderTargetData’ accesses rendering result and has implicit sync point to make sure the batch completed (the same as ‘glReadPixels’). It’s more intuitive to have smth drawn before getting the image.

As for HW Begin/End, I do not see usage for that. In case of Vk/DX12 there is a need for cmd buffer finalization/submission but in practice it’s better to be done in the ‘RCache.OnFrameEnd’

Xottab-DUTY (approved):

PR #486, opened 2019-10-10
xrRender: Backend: `SetTransform` removed

vTurbine:
`SetTransform` is legacy from fixed pipeline world. Can be safely removed


Xottab-DUTY:
Wont it break something on R1? It was created for DX8 and then ported to DX9 and it can probably still use this. Not sure about it.

vTurbine:
Right. Even I didn't see any regression in visuals after change, it seems the debug renderer is affected. No idea how to check it fully and will return to it one day.

Issue #489, opened 2019-10-11
Split renderer front-end and back-end


Xottab-DUTY:
Currently, our renderers front-ends can call graphics APIs (DirectX, OpenGL) functions directly. This is a mess of front and back-ends. Should be fixed to make porting renderer to graphics APIs easier.

PR #490, opened 2019-10-11
xrRender: static geometry buffers code unification (#489)

vTurbine:
- Added staging buffer wrapper for static geometry;
- Common buffer interface for DX9/DX10/DX11/VK
- Backend top file simplified and made common

Please check if it works fine on Linux as well: I don't have OpenGL 4.x on my VM


Xottab-DUTY:
AppVeyor failed to compile this PR =(

vTurbine:
Right, sorry. `BufferUtils.h` was lost in pile of untracked files. Re-pushed with it

vTurbine:
It turned out that after this modification we able to share some of render targets files(!) across R2/R3/R4/GL since it became fully portable. Also this can give a good boost for #492

src/Layers/xrRenderDX9/CommonTypes.h
```diff
@@ -20,3 +20,12 @@ typedef IDirect3DVolumeTexture9 ID3DTexture3D;
 typedef IDirect3DStateBlock9 ID3DState;
 
 #define DX10_ONLY(expr) do {} while (0)
+
+using Unused = int[0];
```
Xottab-DUTY:
```suggestion
using unused_t = int[0];
```

src/Layers/xrRenderDX10/dx10BufferUtils.cpp
```diff
@@ -145,3 +145,37 @@ void ConvertVertexDeclaration(const xr_vector<D3DVERTEXELEMENT9>& declIn, xr_vec
...
+}
+
+void IndexStagingBuffer::Flush()
+{
+    // Upload data to device
+    dx10BufferUtils::CreateIndexBuffer(&m_DeviceBuffer, m_HostData, m_Size, false);
+    HW.stats_manager.increment_stats_ib(m_DeviceBuffer);
+    // Free host memory
+    xr_delete(m_HostData);
+    m_HostData = nullptr;
```
SkyLoaderr: xr_delete already has `ptr = nullptr`

SkyLoaderr:
Has anyone tested this PR on Linux?

src/Layers/xrRenderPC_R2/r2_loader.cpp
```diff
@@ -264,13 +258,11 @@ void CRender::LoadBuffers(CStreamReader* base_fs, bool alternative)
             Msg("* [Loading IB] %d indices, %d Kb", iCount, (iCount * 2) / 1024);
 
             // Create and fill
-            BYTE* pData = 0;
-            R_CHK(HW.pDevice->CreateIndexBuffer(iCount * 2, dwUsage, D3DFMT_INDEX16, D3DPOOL_MANAGED, &_IB[i], 0));
-            HW.stats_manager.increment_stats_ib(_IB[i]);
-            R_CHK(_IB[i]->Lock(0, 0, (void**)&pData, 0));
+            _IB[i].Create(iCount * 2);
```
SkyLoaderr, line 262: Old code doesn't use `D3DUSAGE_SOFTWAREPROCESSING` for `dwUsage` in software mode, but `_IB[i].Create` has it.

src/Layers/xrRender/BufferUtils.h
```diff
@@ -31,6 +31,19 @@ class VertexStagingBuffer
...
 
+    void AddRef()
+    {
+        ++m_RefCounter;
+    }
+
+    u32 Release()
+    {
+        VERIFY(m_RefCounter); // attempting to release unused object
+        if (--m_RefCounter == 0)
```
SkyLoaderr: Maybe separate these two operations?

src/Layers/xrRender/BufferUtils.h
```diff
@@ -0,0 +1,114 @@
...
+    void Flush();
+
+    void AddRef()
+    {
+        ++m_RefCounter;
+    }
+
+    u32 Release()
+    {
+        VERIFY(m_RefCounter); // attempting to release unused object
```
Xottab-DUTY:
```suggestion
        VERIFY2(m_RefCounter, "Attempt to release unused object");
```

src/Layers/xrRenderPC_R2/r2_loader.cpp
vTurbine: This odd for level geometry loading only. I believe this typo came from past times and the flag should be set. See R1 for reference.
Any reasons to treat this particular geometry buffers in different manner? Am I missing smth?

src/Layers/xrRenderPC_R2/r2_rendertarget_accum_point_geom.cpp
```diff
@@ -2,328 +2,38 @@
 
 #include "Layers/xrRender/du_sphere.h"
 
-/*
-Fvector du_sphere_vertices[DU_SPHERE_NUMVERTEX]=
```
Xottab-DUTY: Please, don't remove this, because dead code cleanup and this code itself is not related to this PR.

src/Layers/xrRenderPC_R2/r2_rendertarget_accum_spot_geom.cpp
```diff
@@ -1,107 +1,41 @@
 #include "stdafx.h"
 
 #include "Layers/xrRender/du_cone.h"
-/*
-Fvector du_cone_vertices[DU_CONE_NUMVERTEX]=
```
Xottab-DUTY: Please, don't remove this, because dead code cleanup and this code itself is not related to this PR.

src/Layers/xrRender/BufferUtils.h
```diff
@@ -0,0 +1,116 @@
...
+{
+public:
+    VertexStagingBuffer();
+    ~VertexStagingBuffer();
+
+    void Create(size_t size);
+    void Destroy();
+
+    bool IsValid() const;
+    void* GetHostPointer();
```
Xottab-DUTY: It is better to rename this function to `Lock`.
```suggestion
    void* Lock();
```

src/Layers/xrRender/BufferUtils.h
```diff
@@ -0,0 +1,116 @@
...
+    VertexBufferHandle GetBufferHandle() const;
+    void Flush();
```
Xottab-DUTY: It is better to rename this function to `Unlock`.
```suggestion
    void Unlock();
```

src/Layers/xrRender/BufferUtils.h
```diff
@@ -0,0 +1,116 @@
...
+    IndexStagingBuffer();
+    ~IndexStagingBuffer();
+
+    void Create(size_t size);
+    void Destroy();
+
+    bool IsValid() const;
+    void* GetHostPointer();
+    IndexBufferHandle GetBufferHandle() const;
+    void Flush();
```
Xottab-DUTY: It is better to rename this function to `Unlock`.
```suggestion
    void Unlock();
```

src/Layers/xrRender/BufferUtils.h
```diff
@@ -0,0 +1,116 @@
...
+{
+public:
+    IndexStagingBuffer();
+    ~IndexStagingBuffer();
+
+    void Create(size_t size);
+    void Destroy();
+
+    bool IsValid() const;
+    void* GetHostPointer();
```
Xottab-DUTY: It is better to rename this function to `Lock`.
```suggestion
    void* Lock();
```

src/Layers/xrRender/BufferUtils.h
SkyLoaderr: I agree

007exe:
I checked on ArchLinux in conjunction with the GeForce 9600 GT on proprietary drivers, I did not notice any new bugs

src/Layers/xrRender/BufferUtils.h
vTurbine: Changed to `Map`/`Unmap` pair to be consistent with forthcoming streaming buffers

Xottab-DUTY: It's even more complicated now and should be made easier.

Xottab-DUTY: It is absolutely obvious to call `Unlock()` if you called `Lock()`.
It is less, but obvious to call `Unmap()` if you called `Map()`.

But it is non-obvious to call `Flush()` if you called `Map()`

vTurbine: This complexity was added with purpose. In the case of skinned meshes loading, we will need access to buffer data again.
I can add all the code reworked but this will increase the PR. Or can share my local branch with all geometry stuff reworked

vTurbine: Agree with Map/Unmap inconsistency

Xottab-DUTY: > This complexity was added with purpose. In the case of skinned meshes loading, we will need access to buffer data again.
I can add all the code reworked but this will increase the PR. Or can share my local branch with all geometry stuff reworked

Then we should remove `Unmap()` function and rename `Flush()` to `FlushAndUnmap()`. This is very consistent:
1. Linked with `Map()` functions (function name tells you that you should call it)
2. Corresponds with what the functions does on DX10+/OpenGL. (It flushes the buffer and then discards the host buffer)

vTurbine:
Thank you for review. Let's decide now how to solve the issue with naming.
Two use cases are possible:
1. Lock(map?) host buffer to fill the buffer obtained with data and then upload it to GPU (unlock + flush)
2. Lock host buffer again to read data back and then unlock without flushing since no content changed

It will end up with this patterns:
1.
```
Create(size);
ptr = Lock();
< fill ptr >
Unlock();
Flush();
```

2.
```
Create(size, true);
ptr = Lock();
< fill ptr >
Unlock();
Flush();
...
ptr = Lock(true); // for read
Unlock();
```

So,
- should I add explicit `Unlock` after each buffer use?
- isn't `map/unmap` names fit better for case 2?
- Codacy reports wrongly scoped variables came from vanilla code. Do they need to be fixed too?

Xottab-DUTY:
> isn't map/unmap names fit better for case 2?

Since it doesn't actually Lock/Unlock buffers on DX10+/OpenGL and we want to deprecate DX9 (which is the only implementation which actually does Lock/Unlock) it is better to Map/Unmap, rather than Lock/Unlock.

> should I add explicit Unlock after each buffer use?

Summing up the above and the fact that `Unmap()` does nothing on DX10+/OpenGL, we can just use `Map()`, `Unmap()` and `FlushAndUnmap()`, so no explicit `Unlock()` calls are required.

Eventually, it can look like that:
1.
```
Create(size);
ptr = Map();
< fill ptr >
FlushAndUnmap();
```

2.
```
Create(size, true);
ptr = Map();
< fill ptr >
FlushAndUnmap();
...
ptr = Map(true); // for read
Unmap();
```

The main thing with my suggestion is that then Map, Unmap and FlushAndUnmap functions should be documented well.

`Layers/xrRender/BufferUtils.h` should contain behaviour description for all implementations (DX9/DX10+/OpenGL)

API specific implementations `Layers/xrRender*/*BufferUtils.cpp` should contain behaviour description specific to that API.

> Codacy reports wrongly scoped variables came from vanilla code. Do they need to be fixed too?

Will fix that myself)

Xottab-DUTY:
AppVeyor compilation failed

vTurbine:
> Summing up the above and the fact that Unmap() does nothing on DX10+/OpenGL, we can just use Map(), Unmap() and FlushAndUnmap(), so no explicit Unlock() calls are required.

Will be `Unmap(doFlush=true)` easier then?

1.
```
Create(size);
ptr = Map();
< fill ptr >
Unmap(true); // flush
```
2.
```
Create(size, true); // read capable
ptr = Map();
< fill ptr >
Unmap();
...
ptr = Map(true); // for read
< read ptr>
Unmap();
...
DiscardHostBuffer();
```

Xottab-DUTY:
> Will be Unmap(doFlush=true) easier then?

Yes, it will.

vTurbine:
Sorry for late update: my PC burnt out and it will take a couple of days to bring the next changes

Xottab-DUTY:
Hope you will successfully recover your PC! 😃

src/Layers/xrRenderGL/glBufferUtils.cpp
```diff
@@ -278,4 +280,197 @@ u32 GetDeclLength(const D3DVERTEXELEMENT9* decl)
...
+{
+    return m_HostBuffer ? m_Size : 0;
+}
+
+size_t VertexStagingBuffer::GetVideoMemoryUsage() const
+{
+    if (!m_DeviceBuffer)
+        return 0;
+
+    GLint VB_size;
```
Xottab-DUTY: Local variables should be in camelCase.

src/Layers/xrRenderPC_R2/r2_loader.cpp
Xottab-DUTY: > This odd for level geometry loading only. I believe this typo came from past times and the flag should be set. See R1 for reference.
Any reasons to treat this particular geometry buffers in different manner? Am I missing smth?

@SkyLoaderr, so, do you have something to say about it?

src/Layers/xrRenderGL/glBufferUtils.cpp
vTurbine: Yeah, silly copy-paste issue

src/Layers/xrRenderPC_R2/r2_loader.cpp
Xottab-DUTY: I bet this is not a typo and was made intentional. But, however, R2 probably won't even launch without hardware features, so this should be safe.

Xottab-DUTY (approved):
@vTurbine, anything more to apply or we may merge?

vTurbine:
It's fine, no more changes on this step.
BTW, I removed the code which restores pre-computed indices in `QuadIB` (see `CBackend::RestoreQuadIBData()`). Not sure if anyone uses it so I'm going to remove it completely right after this PR merging.  Is it ok or I need to return this back?

Xottab-DUTY:
It's better to just remove it right in this PR. Why removing in two steps? (partial code removing and then full removal in other PR)
We can and it's better to remove it right now.

Some history:
`CBackend::RestoreQuadIBData()` related to the old bug with rain particles corruption, many programmers tried to fix it in SoC and CS. (the bug appeared even before SoC was released)
It was finally fixed by Armen Abroyan in CoP.

vTurbine:
Thank you for explanation.
Done. Will keep an eye on regression

Xottab-DUTY:
> It was finally fixed by Armen Abroyan in CoP.

The fix is simple. Here's my backport to Clear Sky: https://github.com/OpenXRay/xray-15/commit/b8fd8b925c8593e75addd158deed327c0aff8d0b

PR #498, opened 2019-10-20
Luajit nuget package feature

collectioner:
In this PR:
1. Removed LuaJIT-repository dependency.
2. Updated engine in order to use `openxray.luajit` nuget-package: https://www.nuget.org/packages/openxray.luajit/


collectioner:
Although it works on Windows, Linux build will be broken until we'll update `cmake` file in order to use use our nuget-feed, if it's possible. Or find another solution for Unix-systems.

collectioner:
After quick googling I found that it's possible to use nuget-feed with cmake, but our nuget-package contains `.props` file that adds additional paths for linker and contains command to copy `.dll` file after successful build. So I guess we need to solve how to deal with this file.

collectioner:
Okay. Guys, I updated my branch to the latest version of "openxray.luajit" package - "2.2.2-alpha".
It contains binaries for Linux (for both x64 and x86 systems):
`x86 - native/lib/x86/linux`
and
`x64 - native/lib/x64/linux`
Both binaries have `xrLuajit.so` name.
@eagleivg please, feel free to update our cmake file in order to get binaries from the NuGet and setup additional directories for linker.

q4a:
Having one compiled lib file will not work very good for Linux:
1. Distros have very different dependencies.
2. It will be a bit painfull to support libs for x86, x64, arm, arm64 and rebuild+upload it each time, then it changes.

Xottab-DUTY:
> It will be a bit painfull to support libs for x86, x64, arm, arm64 and rebuild+upload it each time, then it changes.

Not at all. At least AppVeyor can make NuGet packages and upload them automatically.

Issue #446, opened 2019-08-02
Remove -fpermissive flag for Linux builds


kacejot:
Consider removing it for two reasons:

1. Projects used to not have warnings in builds, but this flag downgrades errors to warnings. That is bad and can be a source of serious problems in future.
2. Once this flag is removed, a large part of code will be refactored for successful project builds. This will solve problems like include cycles and deprecated language constructions.

Xottab-DUTY:
Done.

PR #503, opened 2019-10-30
Remove -fpermissive flag for Linux builds

eagleivg:
For #446 Issue

### DIFF
--- a/src/xrAICore/Components/problem_solver_inline.h
+++ b/src/xrAICore/Components/problem_solver_inline.h
@@ -362,7 +362,7 @@ IC void CProblemSolverAbstract::solve()
     m_solution_changed = true;
     m_current_state.clear();
 
-    m_failed = !ai().graph_engine().search(*this, reverse_search ? target_state() : current_state(),
+    m_failed = !GEnv.AISpace->graph_engine().search(*this, reverse_search ? target_state() : current_state(),
         reverse_search ? current_state() : target_state(), &m_solution,
         GraphEngineSpace::CSolverBaseParameters(
             GraphEngineSpace::_solver_dist_type(-1), GraphEngineSpace::_solver_condition_type(-1), 8000));
--- a/src/xrAICore/Navigation/graph_abstract.h
+++ b/src/xrAICore/Navigation/graph_abstract.h
@@ -23,7 +23,7 @@ class CGraphAbstract
 public:
     typedef CVertex<_data_type, _vertex_id_type, CGraphAbstract> CVrtx;
 
-    typedef CEdge<_edge_weight_type, CVrtx, _edge_data_type> CEdge;
+    typedef CEdgeT<_edge_weight_type, CVrtx, _edge_data_type> CEdge;
 
 public:
     typedef xr_map<_vertex_id_type, CVrtx*> VERTICES;
--- a/src/xrAICore/Navigation/graph_abstract.h
+++ b/src/xrAICore/Navigation/graph_abstract.h
@@ -21,13 +21,13 @@ template <typename _data_type = Loki::EmptyType, typename _edge_weight_type = fl
 class CGraphAbstract
 {
 public:
-    typedef CVertex<_data_type, _vertex_id_type, CGraphAbstract> CVertex;
+    typedef CVertex<_data_type, _vertex_id_type, CGraphAbstract> CVrtx;
 
-    typedef CEdge<_edge_weight_type, CVertex, _edge_data_type> CEdge;
+    typedef CEdge<_edge_weight_type, CVrtx, _edge_data_type> CEdge;
 
 public:
-    typedef xr_map<_vertex_id_type, CVertex*> VERTICES;
-    typedef typename CVertex::EDGES EDGES;
+    typedef xr_map<_vertex_id_type, CVrtx*> VERTICES;
+    typedef typename CVrtx::EDGES EDGES;
 
 public:
     typedef typename VERTICES::const_iterator const_vertex_iterator;
@@ -57,8 +57,8 @@ public:
     IC bool empty() const;
     IC const VERTICES& vertices() const;
     IC VERTICES& vertices();
-    IC const CVertex* vertex(const _vertex_id_type& vertex_id) const;
-    IC CVertex* vertex(const _vertex_id_type& vertex_id);
+    IC const CVrtx* vertex(const _vertex_id_type& vertex_id) const;
+    IC CVrtx* vertex(const _vertex_id_type& vertex_id);
     IC const CEdge* edge(const _vertex_id_type& vertex_id0, const _vertex_id_type& vertex_id1) const;
     IC CEdge* edge(const _vertex_id_type& vertex_id0, const _vertex_id_type& vertex_id1);
     IC const CGraphAbstract& header() const;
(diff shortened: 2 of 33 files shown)


src/xrAICore/Navigation/graph_abstract.h
```diff
@@ -21,13 +21,13 @@ template <typename _data_type = Loki::EmptyType, typename _edge_weight_type = fl
 class CGraphAbstract
 {
 public:
-    typedef CVertex<_data_type, _vertex_id_type, CGraphAbstract> CVertex;
+    typedef CVertex<_data_type, _vertex_id_type, CGraphAbstract> CVrtx;
```
Xottab-DUTY, line 24: CAbstractGraphVertex is probably better name than CVrtx

src/xrAICore/Navigation/graph_edge.h
```diff
@@ -30,7 +30,7 @@ class CEdgeBase
 };
 
 template <typename _edge_weight_type, typename _vertex_type, typename _edge_data_type>
-class CEdge : public CEdgeBase<_edge_weight_type, _vertex_type>
+class CEdgeT : public CEdgeBase<_edge_weight_type, _vertex_type>
```
Xottab-DUTY, line 33: It's conflicting with some class with the same name?

src/xrGame/map_manager.h
```diff
@@ -16,7 +16,7 @@ class CMapManager
     CMapManager();
     ~CMapManager();
     void __stdcall Update();
-    /*ICF */ Locations& Locations(); //{return *m_locations;}
+    /*ICF */ Locations& getLocations(); //{return *m_locations;}
```
Xottab-DUTY: What's the problem with this functions?

Anyway, it's better to name it with PascalCase: `GetLocations` (according to the code style)

src/xrGame/ui/UIMapWnd.h
```diff
@@ -156,10 +156,10 @@ class CUIMapWnd : public CUIWindow, public CUIWndCallback
     };
     void AddMapToRender(CUICustomMap*);
     void RemoveMapToRender(CUICustomMap*);
-    CUIGlobalMap* GlobalMap() { return m_GlobalMap; };
-    const GameMaps& GameMaps() { return m_GameMaps; };
+    CUIGlobalMap* GlobalMap() { return m_GlobalMap; }
+    const GameMaps& getGameMaps() { return m_GameMaps; }
```
Xottab-DUTY: The same question as to GetLocations

src/xrAICore/Components/problem_solver_inline.h
```diff
@@ -362,7 +362,7 @@ IC void CProblemSolverAbstract::solve()
     m_solution_changed = true;
     m_current_state.clear();
 
-    m_failed = !ai().graph_engine().search(*this, reverse_search ? target_state() : current_state(),
+    m_failed = !GEnv.AISpace->graph_engine().search(*this, reverse_search ? target_state() : current_state(),
```
Xottab-DUTY, line 365: There's a problem with using `ai()` in this place??

src/xrAICore/Navigation/graph_edge.h
eagleivg: Its conflict with this: `typedef CVertex<_data_type, _vertex_id_type, CGraphAbstract> CVertex`

src/xrAICore/Navigation/graph_abstract.h
eagleivg: A long name is worse because this typedef is used in other templates, and they are so poorly readable.

src/xrGame/map_manager.h
eagleivg: Problem with same name of returning type and method. Ok, PascalCase acceptable.

src/xrGame/ui/UIMapWnd.h
eagleivg: Same as GetLocations

src/xrAICore/Components/problem_solver_inline.h
eagleivg: `ai()` is inline method for CAI_Space (even not in namespace) from xrGame, but this method solve() belongs of parent AISpaceBase of xrUICore:
`class CAI_Space : public AISpaceBase`
Actually, I don’t know how to do it right. The `ai()` method is a transfer artifact from xrGame, probably is need to rewrite the `solve()` method, but I do not understand the logic of its operation.

src/xrAICore/Navigation/graph_abstract.h
Xottab-DUTY: `CGraphVertex`? Just `GraphVertex`?

src/xrAICore/Navigation/graph_edge.h
Xottab-DUTY: CEdge conflicts with CVertex??

src/xrGame/ui/UIMapWnd.h
Xottab-DUTY: So, ok then, let's just use PascalCase)

src/xrAICore/Navigation/graph_edge.h
eagleivg: CVertex conflict with CVertex
and CEdge conflict with CEdge
Both cases template name and typedef is same, this cause error (but with -fpermissive can be ignored)

src/xrAICore/Navigation/graph_abstract.h
GeorgeIvlev: Why type definition is named with `C` prefix? C stands for "class" in short name. Maybe it should be named with `T` prefix? (Type) or if it confusing with Template we can call it just `GraphVrx` since it's just typedef member of  `CGraphAbstract`. What you think about that?

src/xrGame/ui/UIMapWnd.h
```diff
@@ -156,10 +156,10 @@ class CUIMapWnd : public CUIWindow, public CUIWndCallback
     };
     void AddMapToRender(CUICustomMap*);
     void RemoveMapToRender(CUICustomMap*);
-    CUIGlobalMap* GlobalMap() { return m_GlobalMap; };
-    const GameMaps& GameMaps() { return m_GameMaps; };
+    CUIGlobalMap* GlobalMap() { return m_GlobalMap; }
+//    const GameMaps& GetGameMaps() { return m_GameMaps; }
```
Xottab-DUTY, line 160: Why commented? Unused?
Then, it's better to comment it like this:
```suggestion
    //const GameMaps& GetGameMaps() { return m_GameMaps; }
```

src/xrAICore/Navigation/graph_edge.h
Xottab-DUTY: It is better to rename the typedef, not the class!

eagleivg: It`s was do less changes in code

Xottab-DUTY: ~~Maybe, then rename it to CGraphEdge? It seems like a better choise, because CEdgeT causes questions about what `T` actually means.~~

Xottab-DUTY: But, however, renaming the typedef is still better even if it will cause much code changes.

src/xrAICore/Navigation/graph_abstract.h
Xottab-DUTY: `vertex_t`?

kacejot: If we change the code that contains `typedef`, we can replace it with `using`. That will be more expressive

src/xrAICore/Navigation/graph_edge.h
kacejot: `edge_t` is the option here :)

src/xrGame/ui/UIMapWnd.h
kacejot: It is better to remove the comment, I think

src/xrGame/ui/UIMapWnd.h
kacejot: But that is camelCase

Xottab-DUTY: PascalCase should be used for publics and globals.

eagleivg:
Moved to eagleivg:permissive

Issue #258, opened 2018-10-21
OpenGL Renderer: use shaders cross-compiler


Xottab-DUTY:
Source code compilers:
- [glslang](https://github.com/KhronosGroup/glslang) and [SPIRV-Cross](https://github.com/KhronosGroup/spirv-cross)
  - We should check if it supports compilation from HLSL to SPIR-V for **OpenGL**. (https://github.com/KhronosGroup/glslang/issues/1417)
  - [shaderc](https://github.com/google/shaderc)
- [DirectXShaderCompiler](https://github.com/Microsoft/DirectXShaderCompiler) and [SPIRV-Cross](https://github.com/KhronosGroup/spirv-cross)
  - [ShaderConductor](https://github.com/microsoft/ShaderConductor)
- [vkd3d-shader](https://gitlab.winehq.org/wine/vkd3d/-/tree/master/libs/vkd3d-shader)
- [hlsloptconv](https://github.com/snake5/hlsloptconv) (HLSL Optimizing Converter, SM3.0 – SM4.0)
- [hlslparser](https://github.com/unknownworlds/hlslparser)
  - [HLSLParser](https://github.com/Thekla/hlslparser) (Only SM3.0, no preprocessor, so doesn't support #includes)
- [hlsl2glslfork](https://github.com/aras-p/hlsl2glslfork) (Only SM3.0)
- [shgen](https://github.com/pperehozhih/shgen)
- [XShaderCompiler](https://github.com/LukasBanana/XShaderCompiler) (SM4 – SM5) (won't work for us currently, see https://github.com/LukasBanana/XShaderCompiler/issues/98)
- https://zz85.github.io/hlslparser/
- https://github.com/DiligentGraphics/DiligentCore/tree/master/Graphics/HLSL2GLSLConverterLib
  - https://diligentgraphics.com/diligent-engine/shader-converter/
  - http://diligentgraphics.com/diligent-engine/shader-converter/supported-features/
  - https://www.gamedeveloper.com/programming/state-of-the-art-hlsl-to-glsl-converter

Bytecode compilers are not suitable, we need to compile from sources.
However, here's the list of some bytecode compilers:
- [HLSLCrossCompiler](https://github.com/MrShoor/HLSLCrossCompiler)
  - [Amazon's Lumberyard version](https://github.com/aws/lumberyard/tree/master/dev/Code/Tools/HLSLCrossCompiler)
[...]

PR #511, opened 2019-11-05
Lower required GL version to 3.3 + extension

375gnu:
just an experiment. instead of `GL_ARB_gpu_shader5` `GL_ARB_texture_gather` must be used as in https://github.com/dolphin-emu/dolphin/pull/1259/files.

potentially it's possible, so if there is an interest i can complete this pr

### DIFF
--- a/res/gamedata/shaders/gl/accum_base.ps
+++ b/res/gamedata/shaders/gl/accum_base.ps
@@ -33,7 +33,7 @@ float4 _main( p_volume I )
 
 	float2	tcProj			= I.tc.xy / I.tc.w;
 
-	gbuffer_data gbd = gbuffer_load_data( GLD_P(tcProj, pos2d, ISAMPLE) );
+	gbuffer_data gbd = gbuffer_load_data( GLD_P(tcProj, pos2d, uint(ISAMPLE)) );
 
 #ifdef	GBUFFER_OPTIMIZATION
 	//	Emulate virtual offset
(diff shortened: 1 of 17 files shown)


vTurbine:
IMO the `GL_ARB_texture_gather` is useless because of, at least, these reasons:
-  `textureGather` returns only the first swizzle component of texels and particular channel specification is available only in GLSL 4;
- functions from the extension doesn't work with multisampled textures and MSAA will be completely broken;
-  the workaround in the PR you mentioned will cause to significant performance drop and, again, doesn't take MSAA into account.

I think it's ok to go with `GL_ARB_gpu_shader5`. But in this case you don't need to change shaders code because the extension already have support for implicit cast between int<->uint.

375gnu:
But if `GL_ARB_gpu_shader5` is supported then, unless a user has outdated drivers, GL4.1 is supported too. The only case i'm aware of when hardware supports GL_ARB_gpu_shader5 but not GL4.1 is AMD HDXXXX series (with exception of HD58xx and HD69xx) because they don't have almost unused in gamed fp64 support. But recent Mesa versions provide for those cards emulated fp64 so they expose GL4.5. And even with older Mesas it's possible to use MESA_GL_VERSION_OVERRIDE.

So if you want GL renderer for older hardware unable to support `GL_ARB_gpu_shader5` or maybe even `GL_ARB_gpu_shader4` then it might be worth to provide GL3.x and disable MSAA (anyway those cards are not powerful enough), on the other hand, GL4.1 capable hardware is as old as CoP, so only a few users still have so outdated hw.

Xottab-DUTY:
It's a good idea to lower the minimum required OpenGL version. It would be excellent, if it can be solved with minimum drawbacks: disabled MSAA is OK, but decreased performance is bad and we need to try to minimize impact on performance.

Xottab-DUTY:
The topmost priority is implementing cross-compilation from the original HLSL shaders. (#258)
We must direct all our efforts in this direction, rather than supporting special GLSL shaders.

PR #512, opened 2019-11-07
xrRender: staged geometry loading (#489)

vTurbine:
These are final changes to switch all static geometry buffers to staging loading. I'd like to highlight one particular part in `FSkinned.cpp` which restores HW skinning for DX1x and GL renderers.


src/Layers/xrRender/FSkinned.cpp
```diff
@@ -765,27 +578,28 @@ void CSkeletonX_ext::_CollectBoneFaces(Fvisual* V, u32 iBase, u32 iCount)
...
+        vertHW_1W* vertices = static_cast<vertHW_1W*>(data);
+
         for (u32 idx = 0; idx < iCount; idx++)
         {
             vertHW_1W& v = vertices[indices[idx]];
             CBoneData& BD = Parent->LL_GetData(v.get_bone());
             BD.AppendFace(ChildIDX, (u16)(idx / 3));
         }
-        V->p_rm_Vertices->Unlock();
+        V->p_rm_Vertices->Unmap(); // ? keep ?
```
Xottab-DUTY, line 591: To keep or not to keep?

Let me guess: the code is working good, but this is a potential place for optimization?

Xottab-DUTY (approved):
> I'd like to highlight one particular part in FSkinned.cpp which restores HW skinning for DX1x and GL renderers.

And this is awesome! I didn't even know about it)

src/Layers/xrRender/FSkinned.cpp
vTurbine: Yes, you're right. :)
The idea behind this is simple. While loading animated meshes some kinematics setups access their geometry data. It's reasonable to free host heap buffers right after the meshes get processed but I not sure yet where is the best place for it.

Xottab-DUTY: Aha... Ok then. We should note that we can improve this, but I believe, it's better to do this optimization in another PR)

Maybe, @SkyLoaderr can give us a piece of advice on this?

Xottab-DUTY:
@SkyLoaderr said that currently he is very busy and can't provide any feedback on this PR. =(

PR #519, opened 2019-11-24
Gamespy multimaster

FreeZoneMods:
Makes it possible to observe multiplayer servers running all versions of the game.

### DIFF
--- a/src/xrGameSpy/GameSpy_BrowsersWrapper.cpp
+++ b/src/xrGameSpy/GameSpy_BrowsersWrapper.cpp
@@ -221,7 +221,9 @@ GSUpdateStatus CGameSpy_BrowsersWrapper::Update()
     ScopeLock sl(&servers_lock);
     if (last_update_status != GSUpdateStatus::Success && last_update_status != GSUpdateStatus::ConnectingToMaster)
     {
-        return last_update_status;
+        // After first reporting 'bad' status we need to report 'success' one;
+        // otherwise the error dialogs will be shown to user in cycle
+        return GSUpdateStatus::Success;
     }
 
     for (auto& bro : browsers)
--- a/src/xrGameSpy/GameSpy_BrowsersWrapper.cpp
+++ b/src/xrGameSpy/GameSpy_BrowsersWrapper.cpp
@@ -54,11 +54,11 @@ struct SBrowserConfig
     bool active;
 };
 
-static constexpr SBrowserConfig cop_master_bro = { {GAMESPY_GAMENAME, GAMESPY_GAMEKEY}, true};
-static constexpr SBrowserConfig cs_master_bro  = { {GAMESPY_GAMENAME_CS, GAMESPY_GAMEKEY_CS}, true};
-static constexpr SBrowserConfig soc_master_bro = { {GAMESPY_GAMENAME_SOC, GAMESPY_GAMEKEY_SOC}, true};
+static constexpr SBrowserConfig cop_master_bro = { { GAMESPY_GAMENAME, GAMESPY_GAMEKEY }, true };
+static constexpr SBrowserConfig cs_master_bro  = { { GAMESPY_GAMENAME_CS, GAMESPY_GAMEKEY_CS }, true };
+static constexpr SBrowserConfig soc_master_bro = { { GAMESPY_GAMENAME_SOC, GAMESPY_GAMEKEY_SOC }, true };
 
-static constexpr xr_array<SBrowserConfig, 3> master_lists = {cop_master_bro, cs_master_bro, soc_master_bro};
+static constexpr xr_array<SBrowserConfig, 3> master_lists = { cop_master_bro, cs_master_bro, soc_master_bro };
 
 CGameSpy_BrowsersWrapper::CGameSpy_BrowsersWrapper()
 {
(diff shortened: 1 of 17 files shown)


src/xrGameSpy/xrGameSpy.cpp
```diff
@@ -1,16 +1,7 @@
...
-{
-    secretKey[0] = 'L';
-    secretKey[1] = 'T';
-    secretKey[2] = 'U';
-    secretKey[3] = '2';
-    secretKey[4] = 'z';
-    secretKey[5] = '2';
-    secretKey[6] = '\0';
-}
+void FillSecretKey(char* secretKey) { strcpy(secretKey, GAMESPY_GAMEKEY); }
```
Xottab-DUTY:
```suggestion
void FillSecretKey(char* secretKey) { xr_strcpy(secretKey, GAMESPY_GAMEKEY); }
```

src/xrGameSpy/GameSpy_BrowsersWrapper.cpp
```diff
@@ -0,0 +1,290 @@
...
+{
+    CGameSpy_Browser::SMasterListConfig cfg;
+    bool active;
+};
+
+static constexpr SBrowserConfig cop_master_bro = {{GAMESPY_GAMENAME, GAMESPY_GAMEKEY}, true};
+static constexpr SBrowserConfig cs_master_bro = {{GAMESPY_GAMENAME_CS, GAMESPY_GAMEKEY_CS}, true};
+static constexpr SBrowserConfig soc_master_bro = {{GAMESPY_GAMENAME_SOC, GAMESPY_GAMEKEY_SOC}, true};
+
+static constexpr xr_array<SBrowserConfig, 3> master_lists = {{cop_master_bro, cs_master_bro, soc_master_bro}};
```
Xottab-DUTY: It's better to "open" it like that:
```suggestion
static constexpr xr_array<SBrowserConfig, 3> master_lists = {{ cop_master_bro, cs_master_bro, soc_master_bro }};
```

src/xrGameSpy/GameSpy_BrowsersWrapper.cpp
```diff
@@ -0,0 +1,290 @@
...
+            break;
+    }
+    return res;
+}
+
+GSUpdateStatus CGSUpdateStatusAccumulator::GetPessimistic() const
+{
+    ScopeLock sl(&lock);
+    GSUpdateStatus res = GSUpdateStatus::Success;
+    for (auto s : accumulator)
```
Xottab-DUTY: Use `auto` only if it's `std::or_not_std::some_iterator_with_long_name_occupying_entire_line` 🙂 
Otherwise, we should write the type explicitly.

Xottab-DUTY (changes requested):

src/xrGameSpy/GameSpy_BrowsersWrapper.cpp
Xottab-DUTY: Also, why double braces? If VS 2017 does support compilation with single braces, then it's better to remove them.

src/xrGameSpy/xrGameSpy.cpp
FreeZoneMods: Yes, you are right. It's sad there is no simple way to mark std-implementations as deprecated.

Xottab-DUTY: We can: 816fca973cae3cf0a7df28b45b965f7fd92f6742

src/xrGameSpy/GameSpy_BrowsersWrapper.cpp
FreeZoneMods: The auto-formatter doesn't think the code with the spaces is better :) However, looks like it's really possible to get rid of the double braces for master_lists array. Hope GPP will not mind.

src/xrGameSpy/GameSpy_BrowsersWrapper.cpp
FreeZoneMods: Is it a rule from coding style? If yes I think it's... not a good rule. The 'AAA' (Almost-Always-Auto) is often recognized as a recommended practice. It often prevents us from possible bugs, helps us to make less changes in code after changing types of variables, forces the compiler to generate more optimal code and so on. The main counter argument is that sometimes we can't understand the real type quickly while lookung through the code. But it's a not appliciable for this case - the class is quite simple and we can easily know the 'guessed' type.
Do you think that 'for (GSUpdateStatus s : accumulator)' looks better? Or maybe you'd prefer 'for (auto it=accumulator.begin(); it!=accumulator.end(); ++it)? I don't like both of these variants.

Xottab-DUTY: > The main counter argument is that sometimes we can't understand the real type quickly while looking through the code.

Quick type understanding is critical and should be taken into account even in such small case.

> Do you think that 'for (GSUpdateStatus s : accumulator)' looks better?

So, it may look worse, but quick type understanding is more important. However, in my opinion, this variant actually look better.

Xottab-DUTY: But if `auto` will help compiler to produce more effective optimizations then it's a dilemma 🙂

FreeZoneMods: In this case the type deducing is quite obvious (from assigning explicitely-typed variable 'res').

Xottab-DUTY (approved):

PR #521, opened 2019-11-28
Viewport setup abstraction

Xottab-DUTY:


vTurbine:
I believe viewport setup should be done through backend

Xottab-DUTY:
Why?

vTurbine:
Well, in case of vk changing of viewport can be done in runtime via dynamic pipeline state. This will require recording a command into the command buffer which is handled by backend. The HW class should care of command buffers allocating only, not recording

vTurbine:
Also, the viewport is a regular pipeline state like clipping, z-test, color mask and so forth. In the engine, the CBackend is responsible for handling the pipeline states, so it’s better to keep things together

Xottab-DUTY:
Done. Anything else in this PR that should be fixed?

src/Layers/xrRender/R_Backend.h
```diff
@@ -358,6 +358,7 @@ class ECORE_API CBackend
     void set_ClipPlanes(u32 _enable, Fplane* _planes = nullptr, u32 count = 0);
     void set_ClipPlanes(u32 _enable, Fmatrix* _xform = nullptr, u32 fmask = 0xff);
     IC void set_Scissor(Irect* rect = nullptr);
+    IC void SetViewport(D3D_VIEWPORT viewport) const;
```
SkyLoaderr: I think it's not good to pass structure. Maybe use a reference of structure?

Xottab-DUTY: I think it's even better to pass r-value reference.

vTurbine: Yeah. And probably it’s time to rename ‘D3D_VIEWPORT’ into smth like ‘Viewport’

Xottab-DUTY: I think it's a good name, which shows that DirectX is a head renderer and other backends implementations should mimic to DirectX behaviour.

Xottab-DUTY: And personally I can't even imagine a good name for the type that would fit.

`Viewport` could be good, but...
```cpp
// public member variables should be PascalCase

struct ThatWasMadeForSomeReason
{
	SomeType SomeVar;
	... // more variables
	Viewport Viewport; // I'm not sure, but GCC won't be happy about this probably
                       // But even if it will, this looks bad.
}

File: src/Layers/xrRender/R_Backend.h [hunk 2]
```diff
@@ -358,6 +358,8 @@ class ECORE_API CBackend
     void set_ClipPlanes(u32 _enable, Fplane* _planes = nullptr, u32 count = 0);
     void set_ClipPlanes(u32 _enable, Fmatrix* _xform = nullptr, u32 fmask = 0xff);
     IC void set_Scissor(Irect* rect = nullptr);
+    IC void SetViewport(const D3D_VIEWPORT& viewport) const;
+    IC void SetViewport(D3D_VIEWPORT&& viewport) const;
```
vTurbine: It will be fine to have only the first implementation

Xottab-DUTY:
Anything to do?

vTurbine (approved):
Looks good to me 👍

PR #524, opened 2019-12-02
xrRenderGL: speed up texture loading on OpenGL

eagleivg:
The result of checking the engine with a profiler. Removed or rewritten places using `strstr`, speeding up the texture loading stage by about 20ms, potentially accelerating the transition of the NPC to online.

### DIFF
--- a/src/Layers/xrRenderGL/glSH_Texture.cpp
+++ b/src/Layers/xrRenderGL/glSH_Texture.cpp
@@ -164,7 +164,7 @@ void CTexture::Load()
     if (nullptr == *cName)
         return;
     if (0 == xr_stricmp(*cName, "$null")) return;
-    if (nullptr != strstr(*cName, "$user$"))
+    if (0 == strncmp(*cName, "$user$", 6))
     {
         flags.bUser = true;
         return;
--- a/src/Layers/xrRenderPC_GL/glTexture.cpp
+++ b/src/Layers/xrRenderPC_GL/glTexture.cpp
@@ -19,58 +19,12 @@ void fix_texture_name(LPSTR fn)
         *_ext = 0;
 }
 
-int get_texture_load_lod(LPCSTR fn)
-{
-    CInifile::Sect& sect = pSettings->r_section("reduce_lod_texture_list");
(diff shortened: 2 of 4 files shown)


src/Layers/xrRenderGL/glSH_Texture.cpp
```diff
@@ -164,7 +164,7 @@ void CTexture::Load()
     if (nullptr == *cName)
         return;
     if (0 == xr_stricmp(*cName, "$null")) return;
-    if (nullptr != strstr(*cName, "$user$"))
+    if (0 == strncmp(*cName, "$user$", 6))
```
Xottab-DUTY, line 167: Please, do the same optimization for `SH_Texture.cpp` (DX9) and `dx10SH_Texture.cpp` (DX10/11) also

src/Layers/xrRenderPC_GL/glTexture.cpp
```diff
@@ -209,8 +163,7 @@ GLuint CRender::texture_load(LPCSTR fRName, u32& ret_msize, GLenum& ret_desc)
 
         xr_strlwr(fn);
         ret_desc = target;
-        img_loaded_lod = is_target_cube(texture.target()) ? img_loaded_lod : get_texture_load_lod(fn);
```
Xottab-DUTY, line 212: Why do you removed it?

eagleivg: `get_texture_load_lod` called `strstr` only to determine the LOD, in order to eventually find out the size of the texture loaded into memory! At the same time, GLI can determine the size of the loaded texture taking into account the LOD. Therefore, it’s impossible to transfer exactly this edit to DirectX - there you have to calculate in the old way.

Xottab-DUTY: You're talking about `img_loaded_lod`, but I'm talking about `strstr`.

Xottab-DUTY: Just open dx10SH_Texture.cpp

PR #533, opened 2020-01-04
Added auto add rule for bash-completion

007exe:
Enhanced auto-completion for Bash. Conveniently running the Linux terminal engine


Xottab-DUTY:
Please, place the xr_3da file into res/bash-completion folder

PR #534, opened 2020-01-08
xrRender: fixed constant table merging

vTurbine:
The merge code is wrong since the `get` method relies on the assumption of a sorted table. While adding a new constant the table gets modified which breaks `std::lower_bound` logic.
The bug is silent and all you'll have is unexpected graphics artifacts.

It's tricky and had never struck because of merging order in pass compilation:
```
// r_Pass()
    ctable.merge(&ps->constants);
    ctable.merge(&vs->constants);
```
The PS usually has no crossings with VS constants and, even it does, this won't lead to adding a new one. The fun begins if the merge order has changed.

### DIFF
--- a/src/Layers/xrRender/r_constants.cpp
+++ b/src/Layers/xrRender/r_constants.cpp
@@ -204,6 +204,8 @@ void R_constant_table::merge(R_constant_table* T)
         dx9compatibility = true;
 
     // Real merge
+    xr_vector<ref_constant> table_tmp{};
+    table_tmp.reserve(table.size());
     for (u32 it = 0; it < T->table.size(); it++)
     {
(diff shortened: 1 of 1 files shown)


src/Layers/xrRender/r_constants.cpp
```diff
@@ -204,6 +204,8 @@ void R_constant_table::merge(R_constant_table* T)
         dx9compatibility = true;
 
     // Real merge
+    xr_vector<ref_constant> table_tmp{};
```
Xottab-DUTY: No need for `{} `. It would be initialized anyway.

Xottab-DUTY (approved):

PR #558, opened 2020-03-19
Add Vulkan-stub render (from Red panda)

q4a:


Xottab-DUTY:
No new renderers will be added. New API support should be implemented through introducing only new backends, not entire renderers. Moreover, this one is empty stub.

Xottab-DUTY:
Although, the project infrastructure is not ready for adding such new backends; it's only ready for adding new renderers. (as usual in X-Ray)
If it would be ready to render at least simple image, I've could be added it, but I can't approve empty stub.

PR #559, opened 2020-03-22
luajit: fixed VM target extension

vTurbine:
Incorrectly defined compilation target leads to swallowing of #ifdef directives by AS,
hence VM dispatch table contains wrong offsets to byte code implementation (see BC_POW for example).

Note: it's also possible to use `.S` Instead of `.sx`, but if one uses vfat partition on Linux the issue won't be resolved.
Ref: https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html

Should be submitted with https://github.com/OpenXRay/LuaJIT/pull/1


Xottab-DUTY (approved):
I just merged https://github.com/OpenXRay/LuaJIT/pull/1. Now you need to update LuaJIT submodule in this pull request and then PR can be merged. (you can merge it yourself)

PR #564, opened 2020-03-24
xrCommon/xr_string.h: Fix for incorrect hash function, it's also disabled on Windows now

ShokerStlk:
1) Original one was calculating hash for pointer, not string itself. So it's was almost always fail to compare two strings by it's hash. 
2) It's not needed for Windows - it has it's own hash implementation in VC++ Redist


q4a (approved):
Good changes. Thanks!

PR #567, opened 2020-03-31
Moved pBaseRT and pBaseZB from HW to CRenderTarget

Xottab-DUTY:


src/Layers/xrRender/R_Backend_Runtime.cpp
```diff
@@ -44,8 +44,8 @@ void CBackend::OnFrameBegin()
 #ifndef USE_OGL
         RImplementation.rmNormal();
 #endif
-        set_RT(HW.pBaseRT);
-        set_ZB(HW.pBaseZB);
+        set_RT(RImplementation.Target->rt_Base->pRT);
```
vTurbine: It turned out that low level entity (backend) accesses a high level one (target in frontend). I see two ways to resolve this:
- get these targets from the RM by id (requires to make RM global somehow);
- move this logic with targets binding into frontend

IMO, the second option is better

src/Layers/xrRenderDX10/dx10HW.cpp
```diff
@@ -180,8 +180,16 @@ void CHW::CreateDevice(SDL_Window* sdlWnd)
...
+    // Select depth-stencil format
+    constexpr DXGI_FORMAT formats[] =
+    {
+        DXGI_FORMAT_D32_FLOAT_S8X24_UINT,
+        DXGI_FORMAT_D24_UNORM_S8_UINT,
+        DXGI_FORMAT_D32_FLOAT,
+        DXGI_FORMAT_D16_UNORM
+    };
+    const DXGI_FORMAT selectedFormat = SelectFormat(D3D_FORMAT_SUPPORT_DEPTH_STENCIL, formats);
+    Caps.fDepth = dx10TextureUtils::ConvertTextureFormat(selectedFormat);
```
vTurbine, line 195: It’s good to put assertion if `UNKNOWN` was returned

src/Layers/xrRenderPC_R4/r4_rendertarget.cpp
```diff
@@ -367,6 +367,9 @@ CRenderTarget::CRenderTarget()
     //	NORMAL
     {
         u32 w = Device.dwWidth, h = Device.dwHeight;
+        rt_Base.create(BASE_RT(0), w, h, HW.Caps.fTarget, 1, { CRT::CreateBase });
```
vTurbine: The amount of backbuffers should depend on swapchain buffers count. I would recommend put some commentary here and add an assertion if the SC was created with > 1 buffer

src/Layers/xrRender/dxEnvironmentRender.cpp
```diff
@@ -271,7 +271,7 @@ void dxEnvironmentRender::RenderSky(CEnvironment& env)
 #ifdef USE_OGL
 	// Sun must be rendered to generic0 only as it is done in DX
     if (!RImplementation.o.dx10_msaa)
-        RImplementation.Target->u_setrt(RImplementation.Target->rt_Generic_0, nullptr, nullptr, HW.pBaseZB);
+        RImplementation.Target->u_setrt(RImplementation.Target->rt_Generic_0, nullptr, nullptr, RImplementation.Target->rt_Base_Depth->pZRT);
```
vTurbine: How about wrapping this access to the RT with smth like `GetBaseRTV()` and `GetBaseDSV()` ?

Xottab-DUTY: Should it be a member of `CRenderTarget` or global? The first is better, probably.

vTurbine: Definitely a member of `CRenderTarget`

src/Layers/xrRender/R_Backend_Runtime.cpp
Xottab-DUTY: Since OpenGL does the same (low level accesses high level) I suggest to fix both problems not in this pull request, but rather in the next pull request.

src/Layers/xrRender/r__dsgraph_render.cpp
```diff
@@ -443,7 +443,11 @@ void D3DXRenderBase::r_dsgraph_render_hud_ui()
     const ref_rt rt_null;
     RCache.set_RT(0, 1);
     RCache.set_RT(0, 2);
-    auto zb = HW.pBaseZB;
+#ifdef USE_DX9
```
vTurbine: Same code in both cases

src/Layers/xrRender/r__pixel_calculator.cpp
```diff
@@ -22,8 +22,12 @@ void r_pixel_calculator::end()
 {
     Device.RenderEnd();
 
-    RCache.set_RT(HW.pBaseRT);
-    RCache.set_ZB(HW.pBaseZB);
+    RCache.set_RT(RImplementation.Target->get_base_rt());
+#ifdef USE_DX9
```
vTurbine: The same for both cases

vTurbine (approved):

src/Layers/xrRender/r__pixel_calculator.cpp
Xottab-DUTY: Fixed

PR #571, opened 2020-04-05
xrRender: a better place for UI renderer initialisation

vTurbine:
Needed to draw console/stat when loading of xrGame is skipped


src/xrGame/xrgame_dll_detach.cpp
```diff
@@ -32,14 +32,12 @@ extern STORY_PAIRS spawn_story_ids;
 
 extern void release_smart_cast_stats();
 extern void clean_wnd_rects();
-extern void CreateUIGeom();
 extern void DestroyUIGeom();
```
Xottab-DUTY: Forgot to remove?

Xottab-DUTY (approved):

PR #574, opened 2020-04-08
CMake: added option to select memory allocator

vTurbine:
- now memory allocator can be selected with `MEMORY_ALLOCATOR` CMake option. Valid values: mimalloc(default), tbb, standard
- mimalloc tests and shared library are switched off


src/xrCore/xrMemory.cpp
```diff
@@ -10,9 +10,12 @@
 #include <sys/resource.h>
 #endif
 
-#define USE_MIMALLOC
-//#define USE_TBB_MALLOC
-//#define USE_PURE_ALLOC
+#if !defined(XR_PLATFORM_LINUX)
```
Xottab-DUTY:
```suggestion
// On Linux and other platforms these options are controlled by CMake
#if defined(XR_PLATFORM_WINDOWS)
```

src/xrCore/xrMemory.cpp
```diff
@@ -10,9 +10,12 @@
...
+// On Linux these options are controlled by CMake
```
Xottab-DUTY:
```suggestion
```

Xottab-DUTY (approved):

PR #575, opened 2020-04-09
Proper operator new overloading

Xottab-DUTY:
`inline` overloading is not allowed by the standard and it's undefined behaviour. This PR is intended to fix this problem.


Xottab-DUTY:
I need someone who can help me with CMake and consult on how to fix problems with mimalloc. (it overrides operator `new` too and linker does not know what overload to choose)

PR #578, opened 2020-04-18
E2K: added initial support of MCST Elbrus 2000 CPU architecture

r-a-sattarov:
- added initial support of e2k in xrCore and Cmake

E2K (Elbrus 2000) - this is VLIW/EPIC architecture, like Intel Itanium (IA-64) architecture.

About this architecture:
- https://en.wikipedia.org/wiki/Elbrus_2000


src/xrCore/xrDebug.cpp
```diff
@@ -67,6 +67,8 @@ static BOOL bException = FALSE;
 #define MACHINE_TYPE IMAGE_FILE_MACHINE_ARM64
 #elif defined XR_ARCHITECTURE_ARM
 #define MACHINE_TYPE IMAGE_FILE_MACHINE_ARM
+#elif defined XR_ARCHITECTURE_E2K
+#define MACHINE_TYPE IMAGE_FILE_MACHINE_E2K
```
Xottab-DUTY: `MACHINE_TYPE IMAGE_FILE_MACHINE_*` is Windows-specific and it's not needed in our case. (and it's not even supported)

Xottab-DUTY (approved):

PR #580, opened 2020-04-21
xrEngine: allow to run without game module

vTurbine:
Also do not crash if there are no levels data

### DIFF
--- a/src/xrEngine/EngineAPI.cpp
+++ b/src/xrEngine/EngineAPI.cpp
@@ -38,14 +38,20 @@ static bool r2_available = false;
 // Construction/Destruction
 //////////////////////////////////////////////////////////////////////
 
+namespace
+{
 void __cdecl dummy(void) {}
+IFactoryObject* __cdecl dummy(CLASS_ID) { return nullptr; }
+template<typename T>
+void __cdecl dummy(T* p) { R_ASSERT2(p == nullptr, "Attempting to release an object that shouldn't be allocated"); }
+};
 
 CEngineAPI::CEngineAPI()
 {
     hGame = nullptr;
     hTuner = nullptr;
-    pCreate = nullptr;
-    pDestroy = nullptr;
+    pCreate = dummy;
+    pDestroy = dummy;
     tune_enabled = false;
     tune_pause = dummy;
     tune_resume = dummy;
@@ -130,13 +136,16 @@ void CEngineAPI::Initialize(void)
     InitializeRenderers();
 
     hGame = XRay::LoadModule("xrGame");
-    R_ASSERT2(hGame->IsLoaded(), "Game DLL raised exception during loading or there is no game DLL at all");
+    if (!CanSkipGameModuleLoading())
+    {
+        R_ASSERT2(hGame->IsLoaded(), "! Game DLL raised exception during loading or there is no game DLL at all");
 
-    pCreate = (Factory_Create*)hGame->GetProcAddress("xrFactory_Create");
-    R_ASSERT(pCreate);
+        pCreate = reinterpret_cast<Factory_Create*>(hGame->GetProcAddress("xrFactory_Create"));
+        R_ASSERT(pCreate);
 
-    pDestroy = (Factory_Destroy*)hGame->GetProcAddress("xrFactory_Destroy");
-    R_ASSERT(pDestroy);
+        pDestroy = reinterpret_cast<Factory_Destroy*>(hGame->GetProcAddress("xrFactory_Destroy"));
+        R_ASSERT(pDestroy);
+    }
 
     //////////////////////////////////////////////////////////////////////////
     // vTune
--- a/src/xrEngine/EngineAPI.h
+++ b/src/xrEngine/EngineAPI.h
@@ -4,12 +4,13 @@
 // Support for extension DLLs
 //****************************************************************************
 #pragma once
-#include <memory>
 
 #include "xrEngine/Engine.h"
 #include "xrCore/ModuleLookup.hpp"
 #include "xrCore/clsid.h"
 #include "xrCore/xrCore_benchmark_macros.h"
+
+#include <memory>
 
 class IFactoryObject
 {
@@ -76,6 +77,7 @@ public:
     void Destroy();
 
     void CreateRendererList();
+    bool CanSkipGameModuleLoading() const { return !!strstr(Core.Params, "-nogame"); }
 
     CEngineAPI();
     ~CEngineAPI();
--- a/src/xrEngine/main.cpp
+++ b/src/xrEngine/main.cpp
@@ -324,7 +324,7 @@ ENGINE_API void Startup()
     Device.WaitEvent(spatialCreated);
 
     g_pGamePersistent = dynamic_cast<IGame_Persistent*>(NEW_INSTANCE(CLSID_GAME_PERSISTANT));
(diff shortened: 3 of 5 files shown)


src/xrEngine/EngineAPI.cpp
```diff
@@ -130,13 +130,18 @@ void CEngineAPI::Initialize(void)
     InitializeRenderers();
 
     hGame = XRay::LoadModule("xrGame");
-    R_ASSERT2(hGame->IsLoaded(), "Game DLL raised exception during loading or there is no game DLL at all");
-
-    pCreate = (Factory_Create*)hGame->GetProcAddress("xrFactory_Create");
-    R_ASSERT(pCreate);
+    if (!hGame->IsLoaded())
+    {
+        Msg("! Game DLL raised exception during loading or there is no game DLL at all");
```
Xottab-DUTY: This is not a good thing to allow xrGame loading to fail by default, consider adding `-no_game` key which will just disable xrGame loading.

src/xrEngine/EngineAPI.h
```diff
@@ -83,9 +83,9 @@ class ENGINE_API CEngineAPI
 
 ENGINE_API bool is_enough_address_space_available();
 
-#define NEW_INSTANCE(a) Engine.External.pCreate(a)
+#define NEW_INSTANCE(a) Engine.External.pCreate ? Engine.External.pCreate(a) : nullptr
```
Xottab-DUTY: It's better and more safe to assign pointer to a dummy function:
- create dummy will just return nullptr 
- destroy dummy will assert that passed pointer is nullptr.

src/xrEngine/x_ray.cpp
```diff
@@ -393,7 +393,11 @@ void CApplication::Level_Scan()
     Levels.clear();
 
     xr_vector<char*>* folder = FS.file_list_open("$game_levels$", FS_ListFolders | FS_RootOnly);
-    //. R_ASSERT (folder&&folder->size());
+    if (!folder)
+    {
+        Msg("! No levels found in game data");
```
Xottab-DUTY: There's no need to call Msg if we don't need to format log.
```suggestion
        Log("! No levels found in game data");
```

src/xrEngine/x_ray.cpp
```diff
@@ -182,7 +182,6 @@ void CApplication::OnEvent(EVENT E, u64 P1, u64 P2)
         LPSTR op_client = LPSTR(P2);
         Level_Current = u32(-1);
         R_ASSERT(nullptr == g_pGameLevel);
-        R_ASSERT(nullptr != g_pGamePersistent);
```
Xottab-DUTY: This assert should be here.

src/xrEngine/main.cpp
```diff
@@ -324,7 +324,6 @@ ENGINE_API void Startup()
     Device.WaitEvent(spatialCreated);
 
     g_pGamePersistent = dynamic_cast<IGame_Persistent*>(NEW_INSTANCE(CLSID_GAME_PERSISTANT));
-    R_ASSERT(g_pGamePersistent);
```
Xottab-DUTY, line 327: Here should be something like this:
```cpp
R_ASSERT(g_pGamePersistent || Engine.External.IsNoGame());
```

src/xrEngine/x_ray.cpp
```diff
@@ -232,7 +233,6 @@ void CApplication::OnEvent(EVENT E, u64 P1, u64 P2)
         LPSTR demo_file = LPSTR(P1);
 
         R_ASSERT(nullptr == g_pGameLevel);
-        R_ASSERT(nullptr != g_pGamePersistent);
```
Xottab-DUTY: This should be here too. I'm not sure how demo can be started without xrGame 🙂

Xottab-DUTY (changes requested):

src/xrEngine/EngineAPI.cpp
```diff
@@ -165,8 +172,8 @@ void CEngineAPI::Destroy(void)
 {
     hGame = nullptr;
     hTuner = nullptr;
-    pCreate = nullptr;
-    pDestroy = nullptr;
+    factoryCreate  = dummyCreate;
+    factoryDestroy = dummyDestroy;
```
Xottab-DUTY: They should be nullptr.

src/xrEngine/EngineAPI.cpp
```diff
@@ -39,13 +39,15 @@ static bool r2_available = false;
 //////////////////////////////////////////////////////////////////////
 
 void __cdecl dummy(void) {}
+constexpr auto dummyCreate  = [](auto){ return nullptr; };
+constexpr auto dummyDestroy = [](auto){ };
```
Xottab-DUTY: It's better to make it plain functions. Also, destroy should assert that passed pointer is nullptr.

src/xrEngine/EngineAPI.h
```diff
@@ -60,8 +62,8 @@ class ENGINE_API CEngineAPI
 public:
     BENCH_SEC_SCRAMBLEMEMBER1
 
-    Factory_Create*  pCreate;
-    Factory_Destroy* pDestroy;
+    std::function<Factory_Create>  factoryCreate;
+    std::function<Factory_Destroy> factoryDestroy;
```
Xottab-DUTY: What's the point on using std::function? My argument against using it – is that it is slower than plain pointer.

src/xrEngine/EngineAPI.h
```diff
@@ -4,13 +4,15 @@
...
 #pragma once
-#include <memory>
 
 #include "xrEngine/Engine.h"
 #include "xrCore/ModuleLookup.hpp"
 #include "xrCore/clsid.h"
 #include "xrCore/xrCore_benchmark_macros.h"
 
+#include <functional>
+#include <memory>
```
Xottab-DUTY, line 13: Saying just in case: this change is good, don't move <memory> back to top.

src/xrEngine/EngineAPI.h
```diff
@@ -76,16 +78,17 @@ class ENGINE_API CEngineAPI
     void Destroy();
 
     void CreateRendererList();
+    bool SkipGameModuleLoading() const { return !!strstr(Core.Params, "-nogame"); }
```
Xottab-DUTY: Just to be explicit =))
```suggestion
    bool CanSkipGameModuleLoading() const { return !!strstr(Core.Params, "-nogame"); }
```

src/xrEngine/main.cpp
```diff
@@ -324,7 +324,7 @@ ENGINE_API void Startup()
...
+    R_ASSERT(g_pGamePersistent || Engine.External.SkipGameModuleLoading());
```
Xottab-DUTY: Just to be explicit =))
```suggestion
    R_ASSERT(g_pGamePersistent || Engine.External.CanSkipGameModuleLoading());
```

Xottab-DUTY (changes requested):

src/xrEngine/EngineAPI.cpp
vTurbine: Nope. This will crash when trying to release pGamePersistent with `-nogame` option

Xottab-DUTY (approved):
Thank you for your patience!

PR #583, opened 2020-04-25
Cmake improvements ( Linux ). Place all libraries in one output folder 

GeorgeIvlev:
Reconfigure cmake to output compiled libraries and executable to one folder, for example:
  1) With cmake configuration: `cmake -DCMAKE_BUILD_TYPE=Debug ..` files will be placed into:
        `<project_root>/bin/x64/Debug` - for 64-bit system
        `<project_root>/bin/x86/Debug` - for 32-bit system
   2)   With cmake configuration: `cmake -DCMAKE_BUILD_TYPE=Release ..` files will be placed into:
        `<project_root>/bin/x64/Release` - for 64-bit system
        `<project_root>/bin/x86/Release` - for 32-bit system
   3) From there you need to run next command to be sure that libraries will be found by linker in build  output folder first: 
        `export LD_LIBRARY_PATH=<project_root>/bin/<arch_type>/<build_type>`
For example on my system: `export LD_LIBARY_PATH=/home/george/Desktop/xray-16/bin/x64/Debug`
   4) Then you can simply run: `<project_root>/bin/x86/Debug/xr_3da -fsltx /<path_to_gameroot>/fsgame.ltx`


cmake/utils.cmake
```diff
@@ -10,3 +10,13 @@ macro(add_dir DIRS)
...
 endmacro()
+
+
+# ------------------------------------------
+# Detect arch type ( x86 or x64 )
+# ------------------------------------------
+if (CMAKE_SIZEOF_VOID_P EQUAL 8)
+  set(ARCH_TYPE x64)
+else(CMAKE_SIZEOF_VOID_P EQUAL 4)
+  set(ARCH_TYPE x86)
```
r-a-sattarov, line 21: Will it properly work for non-x86 archs (ARM/ARM64/E2K)?

GeorgeIvlev: Yes it should work, but for sure better to check it out, i have an Raspberry Pi 3, but i can check only on next weekend's. Would be great if someone who have that types of processor architecture take a test.

eagleivg: What about cross-compile (32 target on 64 host?)

eagleivg (approved):

Xottab-DUTY:
Hard-reverted the merge of this PR because 9 useless merge commits.

PR #585, opened 2020-04-26
xrRender: It's now possible to not duplicate textures in RAM (all renders)

ShokerStlk:
**Experimental**, off by default. Can be moved to game options.
Thnx to **iOrange** for idea and help.

Added new console command  "texture_disable_managed" and new launch key "-nomanagedtex". Both do the same thing.

OpenGL already have such behavior by default.

In vanila engine almost every texture is **duplicated** to RAM.
It's has point for DX9 since vid_restart require to reload every texture. But for DX10+ it's not needed.

Currently the only drawback i see - R1\R2 on vid_restart require to reload all active textures from HDD. Still it takes ~3 seconds for SSD.

With not vanila gamedata (additional textures) i got RAM usage drop:
R1\R2: from 1.3 GB to 800 Mb (32-bit)
R3\R4: from 2.2 GB to 650-900 Mb (32-bit)

**Not tested for Release build.** But should be fine.

### DIFF
--- a/src/Layers/xrRender/D3DXRenderBase.cpp
+++ b/src/Layers/xrRender/D3DXRenderBase.cpp
@@ -81,7 +81,17 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
 
     Resources->reset_begin();
     Memory.mem_compact();
+
+#if (RENDER == R_R1) || (RENDER == R_R2)
+    bool bNoRamTex = RImplementation.o.no_ram_textures;
+    if (bNoRamTex) ResourcesDeferredUnload();
+#endif //(RENDER == R_R1) || (RENDER == R_R2)
+
     HW.Reset();
+
+#if (RENDER == R_R1) || (RENDER == R_R2)
+    if (bNoRamTex) ResourcesDeferredUpload();
+#endif //(RENDER == R_R1) || (RENDER == R_R2)
 
     std::tie(dwWidth, dwHeight) = HW.GetSurfaceSize();
 
--- a/src/Layers/xrRender/D3DXRenderBase.cpp
+++ b/src/Layers/xrRender/D3DXRenderBase.cpp
@@ -83,7 +83,7 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
     Memory.mem_compact();
 
 #ifdef USE_DX9
-    bool noTexturesInRAM = RImplementation.o.no_ram_textures;
+    const bool noTexturesInRAM = RImplementation.o.no_ram_textures;
     if (noTexturesInRAM)
         ResourcesDeferredUnload();
 #endif
--- a/src/Layers/xrRender/D3DXRenderBase.cpp
+++ b/src/Layers/xrRender/D3DXRenderBase.cpp
@@ -155,6 +155,10 @@ void D3DXRenderBase::ResourcesDeferredUpload()
 {
     Resources->DeferredUpload();
 }
+void D3DXRenderBase::ResourcesDeferredUnload()
+{
+    Resources->DeferredUnload();
+}
 void D3DXRenderBase::ResourcesGetMemoryUsage(u32& m_base, u32& c_base, u32& m_lmaps, u32& c_lmaps)
 {
     if (Resources)
--- a/src/Layers/xrRender/D3DXRenderBase.cpp
+++ b/src/Layers/xrRender/D3DXRenderBase.cpp
@@ -82,16 +82,18 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
     Resources->reset_begin();
     Memory.mem_compact();
 
-#if (RENDER == R_R1) || (RENDER == R_R2)
-    bool bNoRamTex = RImplementation.o.no_ram_textures;
-    if (bNoRamTex) ResourcesDeferredUnload();
-#endif //(RENDER == R_R1) || (RENDER == R_R2)
+#ifdef USE_DX9
+    bool noTexturesInRAM = RImplementation.o.no_ram_textures;
+    if (noTexturesInRAM)
+        ResourcesDeferredUnload();
+#endif
 
     HW.Reset();
 
-#if (RENDER == R_R1) || (RENDER == R_R2)
-    if (bNoRamTex) ResourcesDeferredUpload();
-#endif //(RENDER == R_R1) || (RENDER == R_R2)
+#ifdef USE_DX9
+    if (noTexturesInRAM)
+        ResourcesDeferredUpload();
+#endif
 
     std::tie(dwWidth, dwHeight) = HW.GetSurfaceSize();
 
--- a/src/Layers/xrRender/ResourceManager.cpp
+++ b/src/Layers/xrRender/ResourceManager.cpp
@@ -414,14 +414,20 @@ void CResourceManager::DeferredUpload()
         texture.second->Load();
 #endif
 }
-/*
-void	CResourceManager::DeferredUnload	()
+
+void CResourceManager::DeferredUnload()
 {
-    if (!RDEVICE.b_is_Ready)				return;
-    for (auto t=m_textures.begin(); t!=m_textures.end(); t++)
-        t->second->Unload();
+    if (!RDEVICE.b_is_Ready)
+        return;
+
+#ifndef USE_OGL
+    tbb::parallel_for_each(m_textures, [&](auto m_tex) { m_tex.second->Unload(); });
+#else
+    for (auto& texture : m_textures)
+        texture.second->Unload();
+#endif
 }
-*/
+
 #ifdef _EDITOR
 void CResourceManager::ED_UpdateTextures(AStringVec* names)
 {
--- a/src/Layers/xrRender/xrRender_console.cpp
+++ b/src/Layers/xrRender/xrRender_console.cpp
@@ -80,7 +80,7 @@ const xr_token qminmax_sm_token[] = {{"off", 0}, {"on", 1}, {"auto", 2}, {"autod
 extern int psSkeletonUpdate;
 extern float r__dtex_range;
 
-Flags32 ps_r__common_flags = {0 /* | RFLAG_NO_RAM_TEXTURES */}; // All renders
+Flags32 ps_r__common_flags = {/*RFLAG_NO_RAM_TEXTURES*/}; // All renders
 
 //int ps_r__Supersample = 1;
 int ps_r__LightSleepFrames = 10;
@@ -204,7 +204,6 @@ float ps_r3_dyn_wet_surf_near = 5.f; // 10.0f
 float ps_r3_dyn_wet_surf_far = 20.f; // 30.0f
 int ps_r3_dyn_wet_surf_sm_res = 256; // 256
 
-//AVO: detail draw radius
 u32 ps_steep_parallax = 0;
 int ps_r__detail_radius = 49;
 
--- a/src/Layers/xrRender/xrRender_console.cpp
+++ b/src/Layers/xrRender/xrRender_console.cpp
@@ -80,6 +80,8 @@ const xr_token qminmax_sm_token[] = {{"off", 0}, {"on", 1}, {"auto", 2}, {"autod
 extern int psSkeletonUpdate;
 extern float r__dtex_range;
 
+Flags32 ps_r__common_flags = {0 /* | RFLAG_NO_RAM_TEXTURES */}; // All renders
+
 //int ps_r__Supersample = 1;
 int ps_r__LightSleepFrames = 10;
 
@@ -203,7 +205,6 @@ float ps_r3_dyn_wet_surf_far = 20.f; // 30.0f
 int ps_r3_dyn_wet_surf_sm_res = 256; // 256
 
 //AVO: detail draw radius
-Flags32 ps_common_flags = {0}; // r1-only
 u32 ps_steep_parallax = 0;
 int ps_r__detail_radius = 49;
 
@@ -727,6 +728,8 @@ void xrRender_initconsole()
     tw_max.set(2, 2, 2);
     CMD4(CCC_Vector3, "r__d_tree_wave", &ps_r__Tree_Wave, tw_min, tw_max);
 #endif // DEBUG
+
+    CMD3(CCC_Mask, "r__no_ram_textures", &ps_r__common_flags, RFLAG_NO_RAM_TEXTURES);
 
     CMD2(CCC_tf_Aniso, "r__tf_aniso", &ps_r__tf_Anisotropic); // {1..16}
     CMD2(CCC_tf_MipBias, "r1_tf_mipbias", &ps_r__tf_Mipbias); // {-3 +3}
--- a/src/Layers/xrRenderDX10/dx10Texture.cpp
+++ b/src/Layers/xrRenderDX10/dx10Texture.cpp
@@ -308,7 +308,7 @@ ID3DBaseTexture* CRender::texture_load(LPCSTR fRName, u32& ret_msize, bool bStag
     ZeroMemory(&IMG, sizeof(IMG));
 
     //  Staging control
-    static bool bAllowStaging = !strstr(Core.Params, "-no_staging");
+    static bool bAllowStaging = (!!RImplementation.o.managed_tex_disabled == false && !strstr(Core.Params, "-no_staging"));
     bStaging &= bAllowStaging;
 
     ID3DBaseTexture* pTexture2D = NULL;
@@ -408,7 +408,7 @@ _DDS_CUBE:
     }
     else
     {
-        LoadInfo.Usage = D3D_USAGE_DEFAULT;
+        LoadInfo.Usage = D3D_USAGE_IMMUTABLE;
         LoadInfo.BindFlags = D3D_BIND_SHADER_RESOURCE;
     }
 
@@ -475,7 +475,7 @@ _DDS_2D:
     }
     else
     {
-        LoadInfo.Usage = D3D_USAGE_DEFAULT;
+        LoadInfo.Usage = D3D_USAGE_IMMUTABLE;
         LoadInfo.BindFlags = D3D_BIND_SHADER_RESOURCE;
     }
     LoadInfo.pSrcInfo = &IMG;
--- a/src/Layers/xrRenderDX10/dx10Texture.cpp
+++ b/src/Layers/xrRenderDX10/dx10Texture.cpp
@@ -308,7 +308,7 @@ ID3DBaseTexture* CRender::texture_load(LPCSTR fRName, u32& ret_msize, bool bStag
     ZeroMemory(&IMG, sizeof(IMG));
 
     //  Staging control
-    static bool bAllowStaging = (RImplementation.o.no_ram_textures == false && !strstr(Core.Params, "-no_staging"));
+    static bool bAllowStaging = !RImplementation.o.no_ram_textures;
     bStaging &= bAllowStaging;
 
     ID3DBaseTexture* pTexture2D = NULL;
--- a/src/Layers/xrRenderDX10/dx10Texture.cpp
+++ b/src/Layers/xrRenderDX10/dx10Texture.cpp
@@ -308,7 +308,7 @@ ID3DBaseTexture* CRender::texture_load(LPCSTR fRName, u32& ret_msize, bool bStag
     ZeroMemory(&IMG, sizeof(IMG));
 
     //  Staging control
-    static bool bAllowStaging = (!!RImplementation.o.managed_tex_disabled == false && !strstr(Core.Params, "-no_staging"));
+    static bool bAllowStaging = (RImplementation.o.no_ram_textures == false && !strstr(Core.Params, "-no_staging"));
     bStaging &= bAllowStaging;
 
     ID3DBaseTexture* pTexture2D = NULL;
--- a/src/Layers/xrRenderDX9/dx9HW.cpp
+++ b/src/Layers/xrRenderDX9/dx9HW.cpp
@@ -246,15 +246,27 @@ void CHW::Reset()
         DevPP.FullScreen_RefreshRateInHz = D3DPRESENT_RATE_DEFAULT;
     }
 
+    bool bNoManagedTex = !!RImplementation.o.managed_tex_disabled;
     while (true)
     {
-        const HRESULT result = HW.pDevice->Reset(&DevPP);
-        
+        HRESULT result;
+
+        if (bNoManagedTex)
+        {
+            GEnv.Render->ResourcesDeferredUnload();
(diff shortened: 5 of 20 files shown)


src/Layers/xrRender/ResourceManager.cpp
```diff
@@ -414,14 +414,20 @@ void CResourceManager::DeferredUpload()
...
-    if (!RDEVICE.b_is_Ready)				return;
-    for (auto t=m_textures.begin(); t!=m_textures.end(); t++)
-        t->second->Unload();
+    if (!RDEVICE.b_is_Ready)
+        return;
+
+#ifndef USE_OGL
+    tbb::parallel_for_each(m_textures, [&](auto m_tex) { m_tex.second->Unload(); });
+#else
+    for (auto& texture : m_textures)
```
vTurbine, line 426: Just curious: why don't use parallel unloading too?

src/Layers/xrRenderDX10/dx10Texture.cpp
```diff
@@ -408,7 +408,7 @@ ID3DBaseTexture* CRender::texture_load(LPCSTR fRName, u32& ret_msize, bool bStag
     }
     else
     {
-        LoadInfo.Usage = D3D_USAGE_DEFAULT;
+        LoadInfo.Usage = D3D_USAGE_IMMUTABLE;
```
vTurbine, line 411: Good catch!

src/Layers/xrRenderDX10/dx10Texture.cpp
```diff
@@ -308,7 +308,7 @@ ID3DBaseTexture* CRender::texture_load(LPCSTR fRName, u32& ret_msize, bool bStag
     ZeroMemory(&IMG, sizeof(IMG));
 
     //  Staging control
-    static bool bAllowStaging = !strstr(Core.Params, "-no_staging");
+    static bool bAllowStaging = (!!RImplementation.o.managed_tex_disabled == false && !strstr(Core.Params, "-no_staging"));
```
vTurbine: This render option is ready to use as a boolean type already. No need to explicit comparison and conversion

src/Layers/xrRenderDX9/dx9HW.cpp
```diff
@@ -246,15 +246,27 @@ void CHW::Reset()
...
+        {
+            result = HW.pDevice->Reset(&DevPP);
+        }
+
         if (SUCCEEDED(result))
             break;
 
         Msg("! ERROR: [%dx%d]: %s", DevPP.BackBufferWidth, DevPP.BackBufferHeight, xrDebug::ErrorToString(result));
-        Sleep(100);
+        Sleep((bNoManagedTex ? 3000 : 100)); // xxx: it's better to crash the game
```
vTurbine: This should be an assertion indeed. I guess the `Sleep` was put here to give the driver some span to bring the device up before it got re-created. The increase of delaying time has no effect since uploading of managed data is synchronous operation.

src/xrEngine/defines.h
```diff
@@ -47,7 +47,8 @@ enum
 
     rsShowFPS = (1ul << 24ul), // 24 was reserved for editor
     rsDrawParticles = (1ul << 25ul), // 25 was reserved for editor
-    // 26-32 bit - reserved to Editor
+    rsDisableManagedTex = (1ul << 26ul), // 26 was reserved for editor
```
vTurbine: Shouldn't be a device flag. This setting is related to renderers only -- no reason to make it so global.

src/xrEngine/xr_ioc_cmd.cpp
```diff
@@ -881,6 +881,7 @@ void CCC_Register()
     // Texture manager
     CMD4(CCC_Integer, "texture_lod", &psTextureLOD, 0, 4);
     CMD4(CCC_Integer, "net_dedicated_sleep", &psNET_DedicatedSleep, 0, 64);
+    CMD3(CCC_Mask, "texture_disable_managed", &psDeviceFlags, rsDisableManagedTex);
```
vTurbine: Since all renderers got an option to control texture uploading policy I prefer to use a custom command to access `o` member's field.

src/Layers/xrRenderPC_R1/FStaticRender.cpp
```diff
@@ -91,6 +91,10 @@ void CRender::create()
     o.no_detail_textures = !ps_r2_ls_flags.test(R1FLAG_DETAIL_TEXTURES);
     c_ldynamic_props = "L_dynamic_props";
 
+    o.managed_tex_disabled = (strstr(Core.Params, "-nomanagedtex")) ? TRUE : psDeviceFlags.test(rsDisableManagedTex);
```
vTurbine: Let's change CLI option name to `-no_staging` to be consistent with R3+

src/Layers/xrRenderPC_R1/FStaticRender.h
```diff
@@ -33,6 +33,7 @@ class CRender : public D3DXRenderBase
         u32 disasm : 1; // config
         u32 forceskinw : 1; // config
         u32 no_detail_textures : 1; // config
+        u32 managed_tex_disabled : 1; // don't keep textures in RAM
```
vTurbine: Kinda bad naming. This flag is just a runtime analog of `-no_staging` option.

vTurbine (approved):
Looks well, but personally I'm against adding an additional console command. Or are there some use cases I don't know about?

src/Layers/xrRender/ResourceManager.cpp
Xottab-DUTY: It probably was just copied from `Load()`:

`Load()` has separate singlethreaded path for OpenGL, because it requires that thread, that loads a texture, must own a context and we don't have efficient mechanism to do that with foreign threads such as TBB threads.

I don't know if `Unload()` can be done in parallel, but it's reasonable to do something to make both `Load()` and `Unload()` parallel.

src/xrEngine/xr_ioc_cmd.cpp
Xottab-DUTY: There's a reason to keep them separated so that console command may be enabled by the user, but disabled by renderer if some additional requirements are not met. So, I'm against giving direct access to `o` to the console.

How it works:
Renderer checks for user options (flags) and enables `o` options with some additional checks.

src/Layers/xrRenderDX9/dx9HW.cpp
```diff
@@ -246,15 +246,27 @@ void CHW::Reset()
...
+    bool bNoManagedTex = !!RImplementation.o.managed_tex_disabled;
     while (true)
     {
-        const HRESULT result = HW.pDevice->Reset(&DevPP);
-        
+        HRESULT result;
+
+        if (bNoManagedTex)
+        {
+            GEnv.Render->ResourcesDeferredUnload();
```
Xottab-DUTY: Bad place for this, should be done somewhere in `CRender` or `CResourceManager`.

src/xrEngine/defines.h
Xottab-DUTY: Agree.

src/xrEngine/xr_ioc_cmd.cpp
Xottab-DUTY: Also, it's better to rename it to something like `r__no_texture_staging` and move it to xrRender_Console.cpp

src/Layers/xrRenderPC_R1/FStaticRender.h
Xottab-DUTY: Agree, better name is `no_texture_staging`.

src/Layers/xrRenderDX10/dx10Texture.cpp
Xottab-DUTY: > Good catch!

Seems like something like a bugfix (?!) that should be pushed directly to the main branch.
image

Xottab-DUTY (commented):
We can completely remove `-no_staging` key and make a new console command `r__no_texture_staging` that will only apply after `vid_restart`

src/Layers/xrRenderDX10/dx10Texture.cpp
vTurbine: Not a bug fix but a slight performance improvement

src/Layers/xrRender/ResourceManager.cpp
vTurbine: Yeah, the context sharing. A good point to switch to fibers one day

src/xrEngine/xr_ioc_cmd.cpp
vTurbine: Sounds right, agree

src/Layers/xrRenderDX9/dx9HW.cpp
Xottab-DUTY: I think this is not a part of this PR, so, let's leave it as is in this PR.

ShokerStlk:
> Looks well, but personally I'm against adding an additional console command. Or are there some use cases I don't know about?

For new checkbox in game options (i use it in my fork).

> We can completely remove `-no_staging` key and make a new console command `r__no_texture_staging`

Should i rename console command like this?
-no_staging here is just for backward compatibility.

> ... that will only apply after `vid_restart`

Currently, you need to restart the entire game. I see no profit to bind this to vid_restart, but it's actually may be possible. Still it's just make things more complicated and bugged.

src/Layers/xrRenderDX10/dx10Texture.cpp
ShokerStlk: It's u32 actually:
`u32 managed_tex_disabled : 1;`

But maybe you right - i don't know this sintaxis well.

src/xrEngine/defines.h
ShokerStlk: Well... it was.

I need to store rsDisableManagedTex in FlagsX since it's required for console command to work properly.

There only CCC_Mask in Console (wich works with flags), and no CCC_Bool (maybe i missed something). 

It was CCC_Integer (0\1) previosly, but vanila UI-checkboxes in game options can't handle such things.
_______

So if i can't put it to psDeviceFlags. the i need to create entirely new Flags value only to store one option. Or cut off console command completely.

src/Layers/xrRenderPC_R1/FStaticRender.cpp
ShokerStlk: I can, but "staging" word is mostly from DX10\DX11 (D3D11_USAGE_**STAGING**)
For DX9 is D3DPOOL_**MANAGED**

Currently "-nomanagedtex" works for all renders, "-no_staging" only in DX10\11

src/Layers/xrRenderPC_R1/FStaticRender.h
ShokerStlk: Ok

src/Layers/xrRenderDX9/dx9HW.cpp
ShokerStlk: Will move to D3DXRenderBase::Reset()

src/xrEngine/defines.h
vTurbine: Yeah, you can solve this by making an additional renderer command (see the comment from @Xottab-DUTY below)

src/Layers/xrRenderPC_R1/FStaticRender.cpp
vTurbine: The "staging" is just a common word for the resources transferring technique. Managed resources behave the same as staged but this is hidden by driver logic. The meaning of this option is the same for all renderers

vTurbine: Yeah, because of the different lifetime of RAM textures (that the point of PR), just realized that `-nostaging` is confusing naming for DX9, indeed.
Since the option can be described as "keep textures in VRAM only" probably it's better to name it like `-noramtex`. And rename DX10 `-nostaging` to the same name accordingly.

ShokerStlk:
Update is here

src/Layers/xrRenderDX10/dx10Texture.cpp
```diff
@@ -308,7 +308,7 @@ ID3DBaseTexture* CRender::texture_load(LPCSTR fRName, u32& ret_msize, bool bStag
     ZeroMemory(&IMG, sizeof(IMG));
 
     //  Staging control
-    static bool bAllowStaging = !strstr(Core.Params, "-no_staging");
+    static bool bAllowStaging = (RImplementation.o.no_ram_textures == false && !strstr(Core.Params, "-no_staging"));
```
vTurbine: Now `-noramtex` and `-no_staging` became the same. So we can remove `-no_staging`

vTurbine (approved):

src/Layers/xrRenderDX10/dx10Texture.cpp
Xottab-DUTY:
```suggestion
    static bool bAllowStaging = !RImplementation.o.no_ram_textures;
```

src/Layers/xrRender/xrRender_console.cpp
```diff
@@ -203,7 +205,6 @@ float ps_r3_dyn_wet_surf_far = 20.f; // 30.0f
 int ps_r3_dyn_wet_surf_sm_res = 256; // 256
 
 //AVO: detail draw radius
```
Xottab-DUTY:
```suggestion
```

src/Layers/xrRender/xrRender_console.cpp
```diff
@@ -80,6 +80,8 @@ const xr_token qminmax_sm_token[] = {{"off", 0}, {"on", 1}, {"auto", 2}, {"autod
 extern int psSkeletonUpdate;
 extern float r__dtex_range;
 
+Flags32 ps_r__common_flags = {0 /* | RFLAG_NO_RAM_TEXTURES */}; // All renders
```
Xottab-DUTY: 0 is redundant.
```suggestion
Flags32 ps_r__common_flags = {/*RFLAG_NO_RAM_TEXTURES*/}; // All renders
```

src/Layers/xrRenderPC_R2/r2.h
```diff
@@ -68,6 +68,8 @@ class CRender : public D3DXRenderBase
         u32 disasm : 1;
         u32 advancedpp : 1; //	advanced post process (DOF, SSAO, volumetrics, etc.)
 
+        u32 no_ram_textures : 1; // don't keep textures in RAM
```
Xottab-DUTY: Please, move it upper, after `nulltr`. It's reasonable to separate just technical features from graphical features.

src/Layers/xrRenderPC_R2/r2.cpp
```diff
@@ -294,6 +294,10 @@ void CRender::create()
     o.disasm = (strstr(Core.Params, "-disasm")) ? TRUE : FALSE;
     o.forceskinw = (strstr(Core.Params, "-skinw")) ? TRUE : FALSE;
 
+    o.no_ram_textures = (strstr(Core.Params, "-noramtex")) ? TRUE : ps_r__common_flags.test(RFLAG_NO_RAM_TEXTURES);
+    if (o.no_ram_textures)
+        Msg("* Managed textures disabled");
+
```
Xottab-DUTY: This should be placed after nv-dbt code.

src/Layers/xrRenderPC_R3/r3.cpp
```diff
@@ -302,6 +302,10 @@ void CRender::create()
     o.disasm = (strstr(Core.Params, "-disasm")) ? TRUE : FALSE;
     o.forceskinw = (strstr(Core.Params, "-skinw")) ? TRUE : FALSE;
 
+    o.no_ram_textures = (strstr(Core.Params, "-noramtex")) ? TRUE : ps_r__common_flags.test(RFLAG_NO_RAM_TEXTURES);
+    if (o.no_ram_textures)
+        Msg("* Managed textures disabled");
```
Xottab-DUTY: The same as for R2.

src/Layers/xrRenderPC_R3/r3.h
```diff
@@ -87,6 +87,8 @@ class CRender : public D3DXRenderBase
         u32 advancedpp : 1; //	advanced post process (DOF, SSAO, volumetrics, etc.)
         u32 volumetricfog : 1;
 
+        u32 no_ram_textures : 1; // don't keep textures in RAM
```
Xottab-DUTY: The same as for R2.

src/Layers/xrRenderPC_R4/r4.cpp
```diff
@@ -307,6 +307,10 @@ void CRender::create()
     o.disasm = (strstr(Core.Params, "-disasm")) ? TRUE : FALSE;
     o.forceskinw = (strstr(Core.Params, "-skinw")) ? TRUE : FALSE;
 
+    o.no_ram_textures = (strstr(Core.Params, "-noramtex")) ? TRUE : ps_r__common_flags.test(RFLAG_NO_RAM_TEXTURES);
+    if (o.no_ram_textures)
+        Msg("* Managed textures disabled");
+
```
Xottab-DUTY: The same as for R2.

src/Layers/xrRenderPC_R4/r4.h
```diff
@@ -95,6 +95,8 @@ class CRender : public D3DXRenderBase
         u32 advancedpp : 1; //	advanced post process (DOF, SSAO, volumetrics, etc.)
         u32 volumetricfog : 1;
 
+        u32 no_ram_textures : 1; // don't keep textures in RAM
```
Xottab-DUTY: The same as for R2.

src/Layers/xrRender/D3DXRenderBase.cpp
```diff
@@ -81,8 +81,18 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
...
+#if (RENDER == R_R1) || (RENDER == R_R2)
+    bool bNoRamTex = RImplementation.o.no_ram_textures;
+    if (bNoRamTex) ResourcesDeferredUnload();
+#endif //(RENDER == R_R1) || (RENDER == R_R2)
+
     HW.Reset();
 
+#if (RENDER == R_R1) || (RENDER == R_R2)
+    if (bNoRamTex) ResourcesDeferredUpload();
+#endif //(RENDER == R_R1) || (RENDER == R_R2)
```
Xottab-DUTY:
```suggestion
#endif
```

src/Layers/xrRender/D3DXRenderBase.cpp
```diff
@@ -81,8 +81,18 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
...
     Memory.mem_compact();
+
+#if (RENDER == R_R1) || (RENDER == R_R2)
+    bool bNoRamTex = RImplementation.o.no_ram_textures;
+    if (bNoRamTex) ResourcesDeferredUnload();
+#endif //(RENDER == R_R1) || (RENDER == R_R2)
+
     HW.Reset();
 
+#if (RENDER == R_R1) || (RENDER == R_R2)
```
Xottab-DUTY:
```suggestion
#ifdef USE_DX9
```

src/Layers/xrRender/D3DXRenderBase.cpp
```diff
@@ -81,8 +81,18 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
...
+    if (bNoRamTex) ResourcesDeferredUpload();
```
Xottab-DUTY: Code style
```suggestion
    if (bNoRamTex)
        ResourcesDeferredUpload();
```

src/Layers/xrRender/D3DXRenderBase.cpp
```diff
@@ -81,8 +81,18 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
 
     Resources->reset_begin();
     Memory.mem_compact();
+
+#if (RENDER == R_R1) || (RENDER == R_R2)
+    bool bNoRamTex = RImplementation.o.no_ram_textures;
+    if (bNoRamTex) ResourcesDeferredUnload();
+#endif //(RENDER == R_R1) || (RENDER == R_R2)
```
Xottab-DUTY:
```suggestion
#endif
```

src/Layers/xrRender/D3DXRenderBase.cpp
```diff
@@ -81,8 +81,18 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
 
     Resources->reset_begin();
     Memory.mem_compact();
+
+#if (RENDER == R_R1) || (RENDER == R_R2)
+    bool bNoRamTex = RImplementation.o.no_ram_textures;
+    if (bNoRamTex) ResourcesDeferredUnload();
```
Xottab-DUTY: Code style
```suggestion
    if (noTexturesInRAM)
        ResourcesDeferredUnload();
```

src/Layers/xrRender/D3DXRenderBase.cpp
```diff
@@ -81,8 +81,18 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
 
     Resources->reset_begin();
     Memory.mem_compact();
+
+#if (RENDER == R_R1) || (RENDER == R_R2)
+    bool bNoRamTex = RImplementation.o.no_ram_textures;
```
Xottab-DUTY:
```suggestion
    const bool noTexturesInRAM = RImplementation.o.no_ram_textures;
```

src/Layers/xrRender/D3DXRenderBase.cpp
```diff
@@ -81,8 +81,18 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
 
     Resources->reset_begin();
     Memory.mem_compact();
+
+#if (RENDER == R_R1) || (RENDER == R_R2)
```
Xottab-DUTY:
```suggestion
#ifdef USE_DX9
```

Xottab-DUTY (changes requested):

ShokerStlk:
Updated again

src/Layers/xrRender/D3DXRenderBase.cpp
```diff
@@ -81,8 +81,20 @@ void D3DXRenderBase::Reset(SDL_Window* hWnd, u32& dwWidth, u32& dwHeight, float&
 
     Resources->reset_begin();
     Memory.mem_compact();
+
+#ifdef USE_DX9
+    bool noTexturesInRAM = RImplementation.o.no_ram_textures;
```
Xottab-DUTY:
```suggestion
    const bool noTexturesInRAM = RImplementation.o.no_ram_textures;
```

Xottab-DUTY (approved):

PR #588, opened 2020-04-29
xrRender: RGL: redundant flip phase removed

vTurbine:
-1 draw call / frame


Xottab-DUTY (commented):
I have feeling like `phase_flip()` can be useful in future (when blit extension not supported). So, to not forget about it, let's leave it in the code. (unused but existing)

After some time, we will remove it in the future when we clearly understand that we don't need this function.

Xottab-DUTY (approved):

PR #608, opened 2020-05-11
CMake: Support linking libraries from the current execution folder

Xottab-DUTY:
Fixes the problem when you're trying to launch the engine without installing it to the system.

Taken from #577, thanks to @GeorgeIvlev.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,10 @@ include(GNUInstallDirs)
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 include(${PROJECT_SOURCE_DIR}/cmake/utils.cmake)
 
+# Option to enable support for link libraries from current execution folder for app.
+# If it is on, the linker will search libs in system folder.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -171,6 +175,10 @@ if( USE_CRYPTOPP )
   add_definitions ( -DUSE_CRYPTOPP )
 endif()
 
+if ( CMAKE_BUILD_WITH_INSTALL_RPATH )
+  add_definitions ( -DBUILD_WITH_INSTALL_RPATH )
+endif()
```
Xottab-DUTY, line 155: ~~Please, don't pay attention to the code style. We will fix it later, but for now I just following the code style of the local code.~~
Fixed.

GeorgeIvlev: This is fine, except that `CMAKE_BUILD_WITH_INSTALL_RPATH` variable is `false` by default, you should change condition to:
```
if ( !CMAKE_BUILD_WITH_INSTALL_RPATH )
  add_definitions ( -DBUILD_WITH_INSTALL_RPATH )
endif()
```

GeorgeIvlev:
Works fine. Checked on Ubuntu 20.04 x64.

PR #621, opened 2020-05-15
Linux. CPack improvements.

GeorgeIvlev:
- Added support for dynamic detection of linux distribution for configuration of CPack for RPM and DEB packages.
 - Remove non needed parts from main CMakeLists.txt to utils.cmake and packaging.cmake helper files.
 - Added "build" folder to .gitignore file.

**Note**: Checked on Debian, Ubuntu, Fedora, CentOS. Fedora and CentOS users still need to install rpmbuild package ( `yum install rpm-build` )

PR #625, opened 2020-05-20
Linux. CPack improvements.

GeorgeIvlev:
- Added support for dynamic detection of linux distribution for configuration of CPack for RPM and DEB packages. 
- Remove non needed parts from main CMakeLists.txt to utils.cmake and packaging.cmake helper files. 
- Added "build" folder to .gitignore file. 

Note: Checked on Debian, Ubuntu, Fedora, CentOS. Fedora and CentOS users still need to install rpmbuild package ( `yum install rpm-build` )

This is a reopened version of https://github.com/OpenXRay/xray-16/pull/621

### DIFF
--- a/cmake/packaging.cmake
+++ b/cmake/packaging.cmake
@@ -1,0 +1,56 @@
+set(PACKAGE_PROJECT_NAME        "OpenXRay")
+set(PACKAGE_PROJECT_DESCRIPTION "OpenXRay is an improved version of the X-Ray Engine, the game engine used in the world-famous S.T.A.L.K.E.R. game series by GSC Game World.")
+set(PACKAGE_PROJECT_CONTACT     "OpenXRay <openxray@yahoo.com>")
+set(PACKAGE_PROJECT_VERSION     "1.6.02")
+set(PACKAGE_PROJECT_HOME_URL    "https://github.com/OpenXRay/xray-16")
+
+if (UNIX)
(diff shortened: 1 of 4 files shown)


cmake/packaging.cmake
```diff
@@ -0,0 +1,56 @@
...
+        set(CPACK_RESOURCE_FILE_README ${PROJECT_SOURCE_DIR}/README.md)
+        set(CPACK_RESOURCE_FILE_LICENSE ${PROJECT_SOURCE_DIR}/License.txt)
+
+        # --- SELECT PROPER CPACK GENERATOR ---
+        if (DEBIAN_FOUND)
+            set(CPACK_GENERATOR DEB)
+
+            set(CPACK_DEBIAN_FILE_NAME "DEB-DEFAULT")
+            set(CPACK_DEBIAN_PACKAGE_CONTROL_STRICT_PERMISSION TRUE)
+            set(CPACK_DEBIAN_REVISON "ubuntu-bionic-a1") # TODO: This one need to be detected dynamically
```
GeorgeIvlev, line 26: Not sure about existing and helpfulness of this parameters at all, because official documentation not include them https://cmake.org/cmake/help/v2.8.0/cmake.html#module:CPackDeb don't know from where it was taken. It was added by @eagleivg in this commit 7869f7 you should ask him about that. Maybe it was taken from that article: https://pmateusz.github.io/linux/2017/06/30/linux-secure-apt-repository.html

Xottab-DUTY (approved):

PR #631, opened 2020-06-06
xrGame: Fixed linux resource loading

eagleivg:
Fixed parallel loading of resources for linux. Is it possible that TaskSheduler does not work correctly?

### DIFF
--- a/src/xrGame/GamePersistent.cpp
+++ b/src/xrGame/GamePersistent.cpp
@@ -222,7 +222,7 @@ void CGamePersistent::OnAppStart()
         nullptr, nullptr, &globalsInitialized);
 
     // load game materials
-#ifdef XR_PLATFORM_LINUX
+#ifndef XR_PLATFORM_WINDOWS
     GMLib.Load();
 #else
     TaskScheduler->AddTask("GMLib.Load()", [&]()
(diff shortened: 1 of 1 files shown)


src/xrGame/GamePersistent.cpp
```diff
@@ -222,12 +222,18 @@ void CGamePersistent::OnAppStart()
         nullptr, nullptr, &globalsInitialized);
 
     // load game materials
+#ifdef XR_PLATFORM_LINUX
+    GEnv.Render->MakeContextCurrent(IRender::HelperContext); // free to use, so let's use it
+    GMLib.Load();
+    GEnv.Render->MakeContextCurrent(IRender::NoContext); // release it for other users
+#else
```
Xottab-DUTY, line 227: No need to set helper context here. The thread already owns main context.
Just GMLib.Load() is enough.

Xottab-DUTY (changes requested):

src/xrGame/GamePersistent.cpp
```diff
@@ -222,12 +222,16 @@ void CGamePersistent::OnAppStart()
         nullptr, nullptr, &globalsInitialized);
 
     // load game materials
+#ifdef XR_PLATFORM_LINUX
```
Xottab-DUTY:
```suggestion
#ifndef XR_PLATFORM_WINDOWS
```

Xottab-DUTY (approved):

PR #639, opened 2020-06-25
Fix UB in case of incorrect value in StringTable (game localization configs)

gunslingermod:


src/xrEngine/xr_input_xinput.cpp
```diff
@@ -522,6 +522,13 @@ keyboard_key* KeynameToPtr(pcstr name)
 void GetActionAllBinding(pcstr action, char* dst_buff, int dst_buff_sz)
 {
     int action_id = ActionNameToId(action);
+    if (action_id == kNOTBINDED)
```
eagleivg, line 525: Maybe, better make nifty-counters on StringTable()? It also happens sometimes on Linux
(see GetToken() in console_commands.cpp). Or did I misunderstand the patch's purpose?

src/xrEngine/xr_input_xinput.cpp
```diff
@@ -522,6 +522,13 @@ keyboard_key* KeynameToPtr(pcstr name)
...
+    {
+        if (dst_buff_sz > 0)
+            dst_buff[0] = '\0';
```
nitrocaster, line 528: There's no reason to write '\0' here, since it's a single char and not a string literal. Just put a plain 0 here.

nitrocaster:
What is the issue this PR attempts to fix and how do we test that it actually fixes it?

src/xrEngine/xr_input_xinput.cpp
Xottab-DUTY: @eagleivg, yes, you did misunderstand the purpose of the patch =)

PR #664, opened 2020-09-24
Some PVS-Studio static analyzer warning fixes

dimhotepus:
Product page: https://www.viva64.com/en/pvs-studio/
How to use for free: https://www.viva64.com/en/b/0600/

* Initialize class members to safe defaults instead of garbage.
* Small performance improvements,
* Make macroses safe to use in expressions (operators priority).
* Fix some out of buffer reads.
* Do not access destroyed temporary variables.
and more...

Feel free to discuss/fix commits, XRay is hard.

### DIFF
--- a/src/Layers/xrRenderGL/glBufferUtils.cpp
+++ b/src/Layers/xrRenderGL/glBufferUtils.cpp
@@ -163,6 +163,9 @@ void ConvertVertexDeclaration(const VertexElement* dxdecl, SDeclaration* decl)
         GLenum type = VertexTypeList[desc.Type];
         GLboolean normalized = VertexNormalizedList[desc.Type];
 
+        if (location < 0)
+            continue; // Unsupported
+
         location += desc.UsageIndex;
         CHK_GL(glVertexAttribFormat(location, size, type, normalized, desc.Offset));
         CHK_GL(glVertexAttribBinding(location, desc.Stream));
--- a/src/Layers/xrRenderGL/glBufferUtils.cpp
+++ b/src/Layers/xrRenderGL/glBufferUtils.cpp
@@ -163,9 +163,6 @@ void ConvertVertexDeclaration(const VertexElement* dxdecl, SDeclaration* decl)
         GLenum type = VertexTypeList[desc.Type];
         GLboolean normalized = VertexNormalizedList[desc.Type];
 
-        if (location < 0)
-            continue; // Unsupported
-
         location += desc.UsageIndex;
         CHK_GL(glVertexAttribFormat(location, size, type, normalized, desc.Offset));
         CHK_GL(glVertexAttribBinding(location, desc.Stream));
--- a/src/Layers/xrRenderGL/glState.cpp
+++ b/src/Layers/xrRenderGL/glState.cpp
@@ -194,7 +194,7 @@ void glState::UpdateRenderState(u32 name, u32 value)
 
 void glState::UpdateSamplerState(u32 stage, u32 name, u32 value)
 {
-    if (stage < 0 || CTexture::mtMaxCombinedShaderTextures < stage)
+    if (stage < 0 || stage >= CTexture::mtMaxCombinedShaderTextures)
         return;
 
     GLint currentFilter = (GLint)GL_NEAREST;
--- a/src/utils/ETools/resample.cpp
+++ b/src/utils/ETools/resample.cpp
@@ -240,9 +240,8 @@ static int push(res_state const* const state, SAMPLE* pool, int* const poolfill,
 
     SAMPLE *const destbase = dest, *poolhead = pool + *poolfill, *poolend = pool + state->taps, *newpool = pool;
     SAMPLE const *refill, *base, *endpoint;
-    int lencheck;
 
-    lencheck = res_push_check(state, srclen);
+    const int lencheck = res_push_check(state, srclen);
 
     /* fill the pool before diving in */
     while (poolhead < poolend && srclen > 0)
--- a/src/utils/ETools/resample.cpp
+++ b/src/utils/ETools/resample.cpp
@@ -230,10 +230,6 @@ static SAMPLE sum(
 static int push(res_state const* const state, SAMPLE* pool, int* const poolfill, int* const offset, SAMPLE* dest,
     int dststep, SAMPLE const* source, int srcstep, size_t srclen)
 {
-    SAMPLE *const destbase = dest, *poolhead = pool + *poolfill, *poolend = pool + state->taps, *newpool = pool;
-    SAMPLE const *refill, *base, *endpoint;
-    int lencheck;
-
     assert(state);
     assert(pool);
     assert(poolfill);
@@ -241,6 +237,10 @@ static int push(res_state const* const state, SAMPLE* pool, int* const poolfill,
     assert(source);
 
     assert(state->poolfill != -1);
+
+    SAMPLE *const destbase = dest, *poolhead = pool + *poolfill, *poolend = pool + state->taps, *newpool = pool;
+    SAMPLE const *refill, *base, *endpoint;
+    int lencheck;
 
     lencheck = res_push_check(state, srclen);
 
--- a/src/utils/xrLC/NvMender2002/nv_algebra.cpp
+++ b/src/utils/xrLC/NvMender2002/nv_algebra.cpp
@@ -691,7 +691,7 @@ mat4& perspective(mat4& M, const nv_scalar fovy, const nv_scalar aspect, const n
 
 const quat quat::Identity(0, 0, 0, 1);
 
-quat::quat(nv_scalar x, nv_scalar y, nv_scalar z, nv_scalar w) : x(x), y(y), z(z), w(w) {}
+quat::quat(nv_scalar x_, nv_scalar y_, nv_scalar z_, nv_scalar w_) : x(x_), y(y_), z(z_), w(w_) {}
 quat::quat(const quat& quat)
 {
     x = quat.x;
--- a/src/utils/xrLC/NvMender2002/nv_algebra.cpp
+++ b/src/utils/xrLC/NvMender2002/nv_algebra.cpp
@@ -691,7 +691,7 @@ mat4& perspective(mat4& M, const nv_scalar fovy, const nv_scalar aspect, const n
 
 const quat quat::Identity(0, 0, 0, 1);
 
-quat::quat(nv_scalar x_, nv_scalar y_, nv_scalar z_, nv_scalar w_) : x(x_), y(y_), z(z_), w(w_) {}
+quat::quat(nv_scalar xIn, nv_scalar yIn, nv_scalar zIn, nv_scalar wIn) : x(xIn), y(yIn), z(zIn), w(wIn) {}
 quat::quat(const quat& quat)
 {
     x = quat.x;
--- a/src/utils/xrLC/xrPhase_AdaptiveHT.cpp
+++ b/src/utils/xrLC/xrPhase_AdaptiveHT.cpp
@@ -143,7 +143,6 @@ public:
         for (u32 vit = _from; vit < _to; vit++)
         {
             base_color_c vC;
-            R_ASSERT(vit != u32(-1));
             vecVertex& verts = lc_global_data()->g_vertices();
             R_ASSERT(vit < verts.size());
             Vertex* V = verts[vit];
--- a/src/utils/xrLC/xrPhase_AdaptiveHT.cpp
+++ b/src/utils/xrLC/xrPhase_AdaptiveHT.cpp
@@ -145,7 +145,6 @@ public:
             base_color_c vC;
             R_ASSERT(vit != u32(-1));
             vecVertex& verts = lc_global_data()->g_vertices();
-            R_ASSERT(vit >= 0);
             R_ASSERT(vit < verts.size());
             Vertex* V = verts[vit];
 
--- a/src/utils/xrLC_Light/base_lighting.cpp
+++ b/src/utils/xrLC_Light/base_lighting.cpp
@@ -7,16 +7,15 @@ void base_lighting::select(xr_vector<R_Light>& dest, xr_vector<R_Light>& src, Fv
     Fsphere Sphere;
     Sphere.set(P, R);
     dest.clear();
-    R_Light* L = &*src.begin();
-    for (; L != &*src.end(); L++)
+    for (auto &&L : src)
     {
-        if (L->type == LT_POINT)
+        if (L.type == LT_POINT)
         {
-            float dist = Sphere.P.distance_to(L->position);
-            if (dist > (Sphere.R + L->range))
+            float dist = Sphere.P.distance_to(L.position);
+            if (dist > (Sphere.R + L.range))
                 continue;
         }
-        dest.push_back(*L);
+        dest.push_back(L);
     }
 }
 void base_lighting::select(base_lighting& from, Fvector& P, float R)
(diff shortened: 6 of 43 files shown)


src/xrAICore/Navigation/PathManagers/path_manager_generic_inline.h
```diff
@@ -21,6 +21,7 @@ IC CGenericPathManager::CPathManagerGeneric()
     graph = 0;
     data_storage = 0;
     path = 0;
+    max_visited_node_count = 0;
```
Xottab-DUTY, line 24:
```suggestion
    max_visited_node_count = 0;
    best_node_index = nullptr;
```

src/Layers/xrRenderGL/glBufferUtils.cpp
```diff
@@ -163,9 +163,6 @@ void ConvertVertexDeclaration(const VertexElement* dxdecl, SDeclaration* decl)
         GLenum type = VertexTypeList[desc.Type];
         GLboolean normalized = VertexNormalizedList[desc.Type];
 
-        if (location < 0)
-            continue; // Unsupported
```
Xottab-DUTY: Good catch!

`location` get its value from VertexUsageList, which contains `~0u` values, which are effectively `-1`.

So, the proper fix is to make `location` and `VertexUsageList` to be `GLint` instead of `GLuint` and make `~0u` values `-1`, but... We must ensure that this doesn't break anything. =)
For this pull request, let's not touch this code. (I will look at it and push a separate commit)

src/xrCore/xrDebug.cpp
```diff
@@ -388,36 +388,37 @@ void xrDebug::GatherInfo(char* assertionInfo, size_t bufferSize, const ErrorLoca
         expr = "<no expression>";
     bool extendedDesc = desc && strchr(desc, '\n');
     pcstr prefix = "[error] ";
-    buffer += xr_sprintf(buffer, bufferSize, "\nFATAL ERROR\n\n");
-    buffer += xr_sprintf(buffer, bufferSize, "%sExpression    : %s\n", prefix, expr);
-    buffer += xr_sprintf(buffer, bufferSize, "%sFunction      : %s\n", prefix, loc.Function);
-    buffer += xr_sprintf(buffer, bufferSize, "%sFile          : %s\n", prefix, loc.File);
-    buffer += xr_sprintf(buffer, bufferSize, "%sLine          : %d\n", prefix, loc.Line);
+    const char* oneAboveBuffer = assertionInfo + bufferSize;
```
Xottab-DUTY, line 391: Isn't it unsafe?
```cpp
const char* stringWithLength4096 = "................";
char buffer[4096];

// points to 4097th element
const char* oneAboveBuffer = assertionInfo + bufferSize;

// buffer points to 4096th element now
buffer += xr_sprintf(buffer, oneAboveBuffer - buffer, stringWithLength4096);

// writes to 4097th element
buffer += xr_sprintf(buffer, oneAboveBuffer - buffer, " ");
```

src/xrCore/LocatorAPI.cpp
```diff
@@ -612,7 +612,7 @@ void CLocatorAPI::ProcessOne(pcstr path, const _finddata_t& entry)
             return;
         if (0 == xr_strcmp(entry.name, ".."))
             return;
-        if(path[xr_strlen(path) - 1] != _DELIMITER || path[xr_strlen(path) - 1] != '/')
+        if(path[xr_strlen(path) - 1] != _DELIMITER && path[xr_strlen(path) - 1] != '/')
```
Xottab-DUTY: Let's fix the formatting just by the way...
```suggestion
        if (path[xr_strlen(path) - 1] != _DELIMITER && path[xr_strlen(path) - 1] != '/')
```

src/xr_3da/entry_point.cpp
```diff
@@ -43,12 +43,12 @@ int entry_point(pcstr commandLine)
         shortcuts.Disable();
 #endif
 
-    pcstr fsltx = "-fsltx ";
+    constexpr char fsltx[] = "-fsltx ";
     string_path fsgame = "";
-    if (strstr(commandLine, fsltx))
+    if (pcstr fsltxPos = strstr(commandLine, fsltx); fsltxPos)
```
Xottab-DUTY: Readability is much worse, while the impact on performance is very low.

This code runs just once in the early start stage. (right after `main`/`WinMain`)

src/xr_3da/entry_point.cpp
```diff
@@ -43,12 +43,12 @@ int entry_point(pcstr commandLine)
...
     {
-        const size_t sz = xr_strlen(fsltx);
-        sscanf(strstr(commandLine, fsltx) + sz, "%[^ ] ", fsgame);
+        constexpr size_t sz = std::size(fsltx) - 1;
```
Xottab-DUTY: Readability is a bit bad than before here too.

Xottab-DUTY (changes requested):
Thank you!

Everything is good, except a few notes:
1. It's better to revert glBufferUtils.cpp and entry_point.cpp changes.
2. There's a question in the xrDebug.cpp...
3. There are some suggestions from me :)

src/xr_3da/entry_point.cpp
dimhotepus: Reverted, thanks!

src/xr_3da/entry_point.cpp
dimhotepus: Reverted, thanks!

src/Layers/xrRenderGL/glBufferUtils.cpp
dimhotepus: Reverted, thanks!

src/xrCore/xrDebug.cpp
Xottab-DUTY: ~~Ah... It's safe. Just got confused somehow.
(points to 4096th, then writes to it and points to 4097th, everything is ok)~~

Will it write '\0` to the 4096th element or the string will be left NOT zero terminated?

dimhotepus: Well, `xr_sprintf` is (except `#Ifdef 0` version for linux?) calling safe versions of string functions (*_s), so invalid parameter call handler will be invoked (https://docs.microsoft.com/en-us/cpp/c-runtime-library/parameter-validation. Zero termination is not guaranteed is such case, as i've not found anything about in docs.

As we subscribed on such event in `xrDebug::OnThreadSpawn` via `_set_invalid_parameter_handler`, error will be processed, and no out of buffer write occurs.

dimhotepus: Or, to be more precise we get stack overflow in such case. That's because `_set_invalid_parameter_handler` handler calls `GatherInfo` :).

dimhotepus:
@Xottab-DUTY Thank you for review! I've fixed (almost?) all problems.

Could you, please, check last fixes portion added after first review?

src/utils/xrLC_Light/base_lighting.cpp
```diff
@@ -7,16 +7,15 @@ void base_lighting::select(xr_vector<R_Light>& dest, xr_vector<R_Light>& src, Fv
     Fsphere Sphere;
     Sphere.set(P, R);
     dest.clear();
-    R_Light* L = &*src.begin();
-    for (; L != &*src.end(); L++)
+    for (auto &&L : src)
```
Xottab-DUTY:
```suggestion
    for (const R_Light& L : src)
```

src/utils/ETools/resample.cpp
```diff
@@ -242,6 +238,10 @@ static int push(res_state const* const state, SAMPLE* pool, int* const poolfill,
 
     assert(state->poolfill != -1);
 
+    SAMPLE *const destbase = dest, *poolhead = pool + *poolfill, *poolend = pool + state->taps, *newpool = pool;
+    SAMPLE const *refill, *base, *endpoint;
+    int lencheck;
+
     lencheck = res_push_check(state, srclen);
```
Xottab-DUTY:
```suggestion

    const int lencheck = res_push_check(state, srclen);
```

src/utils/xrLC_Light/xrFace.cpp
```diff
@@ -245,8 +245,6 @@ void DataFace::AddChannel(Fvector2& p1, Fvector2& p2, Fvector2& p3)
 
 BOOL DataFace::hasImplicitLighting()
 {
-    if (0 == this)
-        return FALSE;
```
Xottab-DUTY, line 249: > 'this' pointer can never be NULL.

Unfortunately, this is not true for the MSVC compiler. (it preserves legacy behaviour)

This code should be completely reviewed and rewritten, but for now it's better to **not** remove this check.

src/utils/xrLC_Light/base_lighting.cpp
```diff
@@ -7,16 +7,15 @@ void base_lighting::select(xr_vector<R_Light>& dest, xr_vector<R_Light>& src, Fv
...
+        if (L.type == LT_POINT)
         {
-            float dist = Sphere.P.distance_to(L->position);
-            if (dist > (Sphere.R + L->range))
+            float dist = Sphere.P.distance_to(L.position);
+            if (dist > (Sphere.R + L.range))
                 continue;
         }
-        dest.push_back(*L);
+        dest.push_back(L);
```
Xottab-DUTY:
```suggestion
        dest.emplace_back(L);
```

src/utils/xrLC/xrPhase_AdaptiveHT.cpp
```diff
@@ -145,7 +145,6 @@ class CPrecalcBaseHemiThread : public CThread
             base_color_c vC;
             R_ASSERT(vit != u32(-1));
```
Xottab-DUTY: This check can also be removed.
```suggestion
```

src/xrAICore/Navigation/PatrolPath/patrol_point.cpp
```diff
@@ -20,11 +20,10 @@
...
 
 CPatrolPoint::CPatrolPoint(const CPatrolPath* path)
-{
 #ifdef DEBUG
-    m_path = path;
-    m_initialized = false;
+    : m_flags(0), m_level_vertex_id(0), m_game_vertex_id(0), m_initialized(false), m_path(path)
 #endif
+{
 }
```
Xottab-DUTY, line 29: Should be like this:
```cpp
CPatrolPoint::CPatrolPoint(const CPatrolPath* path)
    : m_flags(0), m_level_vertex_id(u32(-1)), m_game_vertex_id(u32(-1))
#ifdef DEBUG
	, m_initialized(false), m_path(path)
#endif
{
}
```

src/xrGame/battleye_system.cpp
```diff
@@ -348,7 +347,6 @@ bool BattlEyeSystem::InitDLL(LPCSTR dll_name, string_path& out_file)
         }
         fclose(ft);
     }
-    fclose(ft);
```
Xottab-DUTY, line 351: No, should not be removed!

src/Layers/xrRenderGL/glState.cpp
```diff
@@ -194,7 +194,7 @@ void glState::UpdateRenderState(u32 name, u32 value)
 
 void glState::UpdateSamplerState(u32 stage, u32 name, u32 value)
 {
-    if (stage < 0 || CTexture::mtMaxCombinedShaderTextures < stage)
+    if (stage < 0 || stage >= CTexture::mtMaxCombinedShaderTextures)
```
Xottab-DUTY, line 197: Effectively, this doesn't change anything :)

But, yes, now the code is less confusing.

src/utils/xrLC/NvMender2002/nv_algebra.cpp
```diff
@@ -691,7 +691,7 @@ mat4& perspective(mat4& M, const nv_scalar fovy, const nv_scalar aspect, const n
 
 const quat quat::Identity(0, 0, 0, 1);
 
-quat::quat(nv_scalar x, nv_scalar y, nv_scalar z, nv_scalar w) : x(x), y(y), z(z), w(w) {}
+quat::quat(nv_scalar x_, nv_scalar y_, nv_scalar z_, nv_scalar w_) : x(x_), y(y_), z(z_), w(w_) {}
```
Xottab-DUTY: The empty `_` doesn't correspond to the file codestyle.

It's better to write something like this:
```suggestion
quat::quat(nv_scalar xIn, nv_scalar yIn, nv_scalar zIn, nv_scalar wIn) : x(xIn), y(yIn), z(zIn), w(wIn) {}
```

Xottab-DUTY (changes requested):

src/xrGame/battleye_system.cpp
dimhotepus: Could you, please, explain? I see potential double close vulnerability here.

At least if `ft` is closed, we should set it to `nullptr` to prevent last close from trying to close already closed file.

src/xrGame/battleye_system.cpp
```diff
@@ -348,7 +347,6 @@ bool BattlEyeSystem::InitDLL(LPCSTR dll_name, string_path& out_file)
         }
         fclose(ft);
```
Xottab-DUTY, line 349:
```suggestion
        fclose(ft);
        return true;
```

src/xrGame/battleye_system.cpp
Xottab-DUTY: If the first attempt to open the file (line 319) will be successful, then it won't enter into the `if` body and just immediately jump to the end of the function. (line 351)
If we remove fclose(), then file handle will be leaked.

Xottab-DUTY: We can just insert another `return true;` to fix potential double close.

dimhotepus: Good catch, thanks!

Xottab-DUTY:
@dimhotepus, so, what about other two requested changes in `patrol_point.cpp` and `xrFace.cpp`?

dimhotepus:
> @dimhotepus, so, what about other two requested changes in `patrol_point.cpp` and `xrFace.cpp`?

Going to fix it today, was busy a bit.

src/xrAICore/Navigation/PatrolPath/patrol_point.cpp
dimhotepus: Done, thanks!

src/utils/xrLC_Light/xrFace.cpp
dimhotepus: That's UB, but restored anyway.

dimhotepus:
@Xottab-DUTY Thank you for review! If all is good, let's merge.

Xottab-DUTY (approved):

dimhotepus:
CI / build-linux (Debug, x64) (pull_request)

Failure not related to changes, can be ignored.

dimhotepus:
Rebased on latest fix for `CStatTimer`, running CI.

Xottab-DUTY:
Could you also remove commits that doing nothing using rebase, then?)
("[tag] description" commits and their reverting pairs)

dimhotepus:
> Could you also remove commits that doing nothing using rebase, then?)
> ("[tag] description" commits and their reverting pairs)

Removed (mostly) empty commits.

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you!

PR #666, opened 2020-09-27
PVS-Studio static analyzer warning fixes (part 2)

dimhotepus:
Part 2 of fixes by PVS-Studio static analyzer tool.

### DIFF
--- a/Externals/NVTT/src/nvimage/Filter.cpp
+++ b/Externals/NVTT/src/nvimage/Filter.cpp
@@ -297,7 +297,7 @@ Kernel1::Kernel1(const Filter & f, int iscale, int samples/*= 32*/)
 /// Dtor.
 Kernel1::~Kernel1()
 {
-	delete m_data;
+	delete[] m_data;
 }
 
 /// Print the kernel for debugging purposes.
@@ -329,7 +329,7 @@ Kernel2::Kernel2(const Kernel2 & k) : m_windowSize(k.m_windowSize)
 /// Dtor.
 Kernel2::~Kernel2()
 {
-	delete m_data;
+	delete[] m_data;
 }
 
 /// Normalize the filter.
@@ -602,4 +602,3 @@ void PolyphaseKernel::debugPrint() const
 		nvDebug("\n");
 	}
 }
-
--- a/src/Include/editor/ide.hpp
+++ b/src/Include/editor/ide.hpp
@@ -22,6 +22,7 @@ class property_holder_holder;
 class ide_base
 {
 public:
+    virtual ~ide_base() = default;
     virtual HWND main_handle() = 0;
     virtual HWND view_handle() = 0;
     virtual CEnvironment* environment() = 0;
--- a/src/Include/editor/property_holder_base.hpp
+++ b/src/Include/editor/property_holder_base.hpp
@@ -126,6 +126,7 @@ public:
     }; // enum value_editor_enum
 
 public:
+    virtual ~property_holder_base() = default;
     virtual property_holder_holder* holder() = 0;
     virtual void clear() = 0;
     virtual property_value* add_property(LPCSTR identifier, LPCSTR category, LPCSTR description,
--- a/src/Layers/xrRender/DetailManager.cpp
+++ b/src/Layers/xrRender/DetailManager.cpp
@@ -213,10 +213,10 @@ void CDetailManager::Unload()
 {
     hw_Unload();
 
-    for (DetailIt it = objects.begin(); it != objects.end(); it++)
+    for (CDetail* detailObject : objects)
     {
-        (*it)->Unload();
-        xr_delete(*it);
+        detailObject->Unload();
+        xr_delete(detailObject);
     }
 
     objects.clear();
--- a/src/editors/xrPostprocessEditor/EditorEngine.cs
+++ b/src/editors/xrPostprocessEditor/EditorEngine.cs
@@ -6,17 +6,27 @@ using XRay.ManagedApi.Core;
 
 namespace xrPostprocessEditor
 {
-    public struct NoiseParams
+    public struct NoiseParams : IEquatable<NoiseParams>
     {
         public float Intensity;
         public float Grain;
         public float FPS;
+
+        public bool Equals(NoiseParams other)
+        {
+            return Intensity == other.Intensity && Grain == other.Grain && FPS == other.FPS;
+        }
     }
 
-    public struct ColorMappingParams
+    public struct ColorMappingParams : IEquatable<ColorMappingParams>
     {
         public float Influence;
         public string Texture;
+
+        public bool Equals(ColorMappingParams other)
+        {
+            return Influence == other.Influence && Texture == other.Texture;
+        }
     }
 
     public class EditorEngine : IDisposable
--- a/src/editors/xrSdkControls/Controls/TreeViewSearchPanel/TreeViewSearchPanel.cs
+++ b/src/editors/xrSdkControls/Controls/TreeViewSearchPanel/TreeViewSearchPanel.cs
@@ -72,7 +72,7 @@ namespace XRay.SdkControls
                     MessageBox.Show("There is no matches!");
                 else if (findedNodes.Count == 1)
                     resultLabel.Text = "Result: 1 / 1";
-                else if (findedNodes.Count > 1)
+                else
                 {
                     resultLabel.Enabled = true;
                     searchNext.Enabled = true;
@@ -83,14 +83,6 @@ namespace XRay.SdkControls
                     findedNodes[0].EnsureVisible();
                     selectedNodePrevColor = findedNodes[0].BackColor;
                     findedNodes[0].BackColor = Color.Coral;
-                }
-                else
-                {
-                    resultLabel.Enabled = false;
-                    searchNext.Enabled = false;
-                    searchPrev.Enabled = false;
-                    findedNodes.Clear();
-                    searchResultIndex = -1;
                 }
             }
             if (Visible && e.KeyCode == Keys.Escape)
--- a/src/editors/xrWeatherEditor/property_converter_color.cpp
+++ b/src/editors/xrWeatherEditor/property_converter_color.cpp
@@ -82,10 +82,10 @@ bool property_converter_color::CanConvertFrom(ITypeDescriptorContext ^ context, 
 
 Object ^ property_converter_color::ConvertFrom(ITypeDescriptorContext ^ context, CultureInfo ^ culture, Object ^ value)
 {
-    String ^ string = dynamic_cast<String ^>(value);
     if (!value)
         return (inherited::ConvertFrom(context, culture, value));
 
+    String ^ string = dynamic_cast<String ^>(value);
     try
     {
         int comma = string->IndexOf(" ");
--- a/src/utils/ETools/audio.cpp
+++ b/src/utils/ETools/audio.cpp
@@ -226,7 +226,6 @@ int aiff_open(FILE* in, oe_enc_opt* opt, unsigned char* buf, int buflen)
     unsigned char* buffer;
     unsigned char buf2[8];
     aiff_fmt format;
-    aifffile* aiff = (aifffile*)malloc(sizeof(aifffile));
 
     if (buf[11] == 'C')
         aifc = 1;
@@ -258,7 +257,7 @@ int aiff_open(FILE* in, oe_enc_opt* opt, unsigned char* buf, int buflen)
     format.samplesize = READ_U16_BE(buffer + 6);
     format.rate = (int)read_IEEE80(buffer + 8);
 
-    aiff->bigendian = 1;
+    short bigendian = 1;
 
     if (aifc)
     {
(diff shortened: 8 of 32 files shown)


Externals/NVTT/src/nvimage/Filter.cpp
```diff
@@ -329,7 +329,7 @@ Kernel2::Kernel2(const Kernel2 & k) : m_windowSize(k.m_windowSize)
 /// Dtor.
 Kernel2::~Kernel2()
 {
-	delete m_data;
+	delete [] m_data;
```
Xottab-DUTY:
```suggestion
	delete[] m_data;
```

Externals/NVTT/src/nvimage/Filter.cpp
```diff
@@ -297,7 +297,7 @@ Kernel1::Kernel1(const Filter & f, int iscale, int samples/*= 32*/)
 /// Dtor.
 Kernel1::~Kernel1()
 {
-	delete m_data;
+	delete [] m_data;
```
Xottab-DUTY:
```suggestion
	delete[] m_data;
```

src/editors/xrPostprocessEditor/EditorEngine.cs
```diff
@@ -6,17 +6,27 @@
...
 
-    public struct ColorMappingParams
+    public struct ColorMappingParams : IEquatable<ColorMappingParams>
     {
         public float Influence;
         public string Texture;
+
+        public bool Equals(ColorMappingParams other)
+        {
+          return Influence == other.Influence && Texture == other.Texture;
```
Xottab-DUTY:
```suggestion
            return Influence == other.Influence && Texture == other.Texture;
```

src/Layers/xrRender/DetailManager.cpp
```diff
@@ -213,16 +213,19 @@ void CDetailManager::Unload()
...
 
     objects.clear();
-    m_visibles[0].clear();
-    m_visibles[1].clear();
-    m_visibles[2].clear();
+
+    for (vis_list& l : m_visibles)
+    {
+        l.clear();
+    }
```
Xottab-DUTY: Wastes too much space, introduces indentation to the right, too small – so, doesn't worth it.

src/utils/ETools/audio.cpp
```diff
@@ -776,6 +792,7 @@ void setup_downmix(oe_enc_opt* opt)
     if (opt->channels != 2)
     {
         fprintf(stderr, "Internal error! Please report this bug.\n");
+        free(d);
```
Xottab-DUTY: It's better to just move the memory allocation down.

src/utils/ETools/audio.cpp
```diff
@@ -250,6 +252,7 @@ int aiff_open(FILE* in, oe_enc_opt* opt, unsigned char* buf, int buflen)
     if (fread(buffer, 1, len, in) < len)
     {
         fprintf(stderr, _("Warning: Unexpected EOF in reading AIFF header\n"));
+        free(aiff);
```
Xottab-DUTY: We can just move the `aiff` allocation down to the place where it's actually used.

`aiff->bigendian =` can be replaced with `int bigendian =`

src/utils/ETools/audio.cpp
```diff
@@ -366,11 +375,15 @@ int wav_open(FILE* in, oe_enc_opt* opt, unsigned char* oldbuf, int buflen)
      */
 
     if (!find_wav_chunk(in, "fmt ", &len))
+    {
+        free(wav);
```
Xottab-DUTY: We can just move the `wav` allocation down to the place where it's actually used.

src/Include/editor/ide.hpp
```diff
@@ -22,6 +22,7 @@ class property_holder_holder;
 class ide_base
 {
 public:
+    virtual ~ide_base() {}
```
Xottab-DUTY:
```suggestion
    virtual ~ide_base() = default;
```

src/Include/editor/property_holder_base.hpp
```diff
@@ -126,6 +126,7 @@ class property_holder_base
     }; // enum value_editor_enum
 
 public:
+    virtual ~property_holder_base() {}
```
Xottab-DUTY:
```suggestion
    virtual ~property_holder_base() = default;
```

src/editors/xrSdkControls/Controls/TreeViewSearchPanel/TreeViewSearchPanel.cs
```diff
@@ -72,7 +72,7 @@ private void searchBox_KeyDown(object sender, KeyEventArgs e)
                     MessageBox.Show("There is no matches!");
                 else if (findedNodes.Count == 1)
                     resultLabel.Text = "Result: 1 / 1";
-                else if (findedNodes.Count > 1)
```
Xottab-DUTY, line 75: And now with have two `else { ... } else { ... }` branches

src/Layers/xrRender/DetailManager.cpp
dimhotepus: Ok, restored.

src/utils/ETools/audio.cpp
dimhotepus: Unfortunately, `aiff` is used in nested block, too. 

May be use `std::make_unique` + `release` where needed?

src/editors/xrSdkControls/Controls/TreeViewSearchPanel/TreeViewSearchPanel.cs
dimhotepus: Sorry, output window has hidden last else from me. Just removed dead code.

src/utils/ETools/audio.cpp
dimhotepus: Nevermind, fixed by `int bigendian =`

src/utils/ETools/audio.cpp
```diff
@@ -366,7 +366,9 @@ int wav_open(FILE* in, oe_enc_opt* opt, unsigned char* oldbuf, int buflen)
      */
 
     if (!find_wav_chunk(in, "fmt ", &len))
+    {
         return 0; /* EOF */
+    }
```
Xottab-DUTY: Unneeded formatting

src/Layers/xrRender/DetailManager.cpp
```diff
@@ -213,16 +213,17 @@ void CDetailManager::Unload()
...
-        xr_delete(*it);
+        o->Unload();
+        xr_delete(o);
     }
 
     objects.clear();
     m_visibles[0].clear();
     m_visibles[1].clear();
     m_visibles[2].clear();
+
```
Xottab-DUTY: Unneeded formatting

src/utils/ETools/audio.cpp
```diff
@@ -258,7 +257,7 @@ int aiff_open(FILE* in, oe_enc_opt* opt, unsigned char* buf, int buflen)
     format.samplesize = READ_U16_BE(buffer + 6);
     format.rate = (int)read_IEEE80(buffer + 8);
 
-    aiff->bigendian = 1;
+    int bigendian = 1;
```
Xottab-DUTY: Was wrong a bit. Let's make it 100% correspond to `aifffile::bigendian` type
```suggestion
    short bigendian = 1;
```

src/Layers/xrRender/DetailManager.cpp
```diff
@@ -213,16 +213,17 @@ void CDetailManager::Unload()
 {
     hw_Unload();
 
-    for (DetailIt it = objects.begin(); it != objects.end(); it++)
+    for (CDetail* o : objects)
```
Xottab-DUTY: It's better to name it `detailObject`.

dimhotepus:
Not sure about Codacity's `Do not check floating point equality with exact values, use a range instead.` issues.

I can add correct implementation of floating point comparison via `Math.Abs(expected - actual) <|<= Epsilon`, but two questions arise:
* Which `Epsilon` should i choose?
* Do we need this at all for now?

Xottab-DUTY:
[OpenXRay#666@aba138b](https://github.com/OpenXRay/xray-16/pull/666/commits/aba138b183420cb5165f7f87bf3e24248a29408c)
It would be unhandy to rebase this commit 😬

dimhotepus:
> [OpenXRay#666@aba138b](https://github.com/OpenXRay/xray-16/pull/666/commits/aba138b183420cb5165f7f87bf3e24248a29408c)
> It would be unhandy to rebase this commit 😬

May be do not rebase? Or you prefer `every commit is a release one` idiom?

Xottab-DUTY:
> May be do not rebase? Or you prefer every commit is a release one idiom?

Not exactly, but almost like that. I'm just trying to keep history clean.
The rule sounds something like this: `once a WIP branch is ready to merge, it should be cleaned up from commits that fix previous commits. (fixing commits should be fixup'ed into the commits they are fixing)`

> Not sure about Codacity's `Do not check floating point equality with exact values, use a range instead.` issues.
> 
> I can add correct implementation of floating point comparison via `Math.Abs(expected - actual) <|<= Epsilon`, but two questions arise:
> 
> * Which `Epsilon` should i choose?
> * Do we need this at all for now?

Not sure about this too... Probably, we can ignore the Codacy's warning... But, when IEquatable was unimplemented and reflection was used, how it was implemented? Was it checking exact values or using a proper comparison with epsilon?

dimhotepus:
> > May be do not rebase? Or you prefer every commit is a release one idiom?
> 
> Not exactly, but almost like that. I'm just trying to keep history clean.
> The rule sounds something like this: `once a WIP branch is ready to merge, it should be cleaned up from commits that fix previous commits. (fixing commits should be fixup'ed into the commits they are fixing)`

Ok, i'll split and rebase.

> > Not sure about Codacity's `Do not check floating point equality with exact values, use a range instead.` issues.
> > I can add correct implementation of floating point comparison via `Math.Abs(expected - actual) <|<= Epsilon`, but two questions arise:
> > 
> > * Which `Epsilon` should i choose?
> > * Do we need this at all for now?
> 
> Not sure about this too... Probably, we can ignore the Codacy's warning... But, when IEquatable was unimplemented and reflection was used, how it was implemented? Was it checking exact values or using a proper comparison with epsilon?

Hm. If i remember correctly, C# spec says value types default equality operator uses bitwise comparison, so behavior is the same as before except costly reflection calls are replaced by our custom `Equals` implementation.

Xottab-DUTY:
> Hm. If i remember correctly, C# spec says value types default equality operator uses bitwise comparison, so behavior is the same as before except costly reflection calls are replaced by our custom Equals implementation.

Then we can ignore the warning.

dimhotepus:
> > Hm. If i remember correctly, C# spec says value types default equality operator uses bitwise comparison, so behavior is the same as before except costly reflection calls are replaced by our custom Equals implementation.
> 
> Then we can ignore the warning.

Thank you for review. I've rebased and squashed all the things.

src/editors/xrWeatherEditor/property_converter_color.cpp
```diff
@@ -83,7 +83,7 @@ bool property_converter_color::CanConvertFrom(ITypeDescriptorContext ^ context,
 Object ^ property_converter_color::ConvertFrom(ITypeDescriptorContext ^ context, CultureInfo ^ culture, Object ^ value)
 {
     String ^ string = dynamic_cast<String ^>(value);
-    if (!value)
```
Xottab-DUTY, line 86: I just checked the code and I see that this replacement is wrong and the code was written like that intentionally.

Xottab-DUTY: But it makes sense to swap the code like that:
```
	if (!value)
		return (inherited::ConvertFrom(context, culture, value));

	String ^ string = dynamic_cast<String ^>(value);
	try
	{
		...
```

dimhotepus: Good to catch it before the merge.

I'll move cast below value check than.

dimhotepus: > But it makes sense to swap the code like that:
> 
> ```
> 	if (!value)
> 		return (inherited::ConvertFrom(context, culture, value));
> 
> 	String ^ string = dynamic_cast<String ^>(value);
> 	try
> 	{
> 		...
> ```

Done, thank you. Hope space doesn't matter here.

Xottab-DUTY: Well, it doesn't look good, sorry
oops

dimhotepus: It'll be better to use ClangFormat or similar tool to automate boring formatting task.

I'll fix formatting tomorrow, thanks!

dimhotepus: Done!

Xottab-DUTY: Thank you!

> It'll be better to use ClangFormat or similar tool to automate boring formatting task.

We have `.clang-format` file in the `src` dir. Doesn't it work?

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you!

PR #673, opened 2020-10-05
Enable GC64 on 64-bit platforms by default

ghost:
It's strange that even for 64-bit platforms GC32 is still used, after all, GC64 has long been declared stable. Enabling GC64 will help fix a known "not enough memory" bug.


Xottab-DUTY:
@jacendi thanks for the pull request!

@eagleivg, @007exe, can you test it?

ghost:
Got some troubles with LuaJIT, closed until until i fix it

PR #680, opened 2020-11-04
HUD: added switch for left handed HUD

vTurbine:
HL-like console command

image

### DIFF
--- a/src/Layers/xrRender/r__dsgraph_render.cpp
+++ b/src/Layers/xrRender/r__dsgraph_render.cpp
@@ -336,6 +336,8 @@ class hud_transform_helper
 {
     Fmatrix Pold;
     Fmatrix FTold;
+    static u32 cullMode;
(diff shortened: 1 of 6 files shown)


src/Layers/xrRender/r__dsgraph_render.cpp
```diff
@@ -379,9 +385,34 @@ class hud_transform_helper
...
+    {
+        if (!isActive)
+        {
+            return;
+        }
+
+        if (!psHUD_Flags.test(HUD_LEFT_HANDED))
+        {
+            return;
+        }
```
Xottab-DUTY:
```suggestion
        if (!isActive || !psHUD_Flags.test(HUD_LEFT_HANDED))
            return;
```

Xottab-DUTY (approved):
Just cosmetic code changes required.

PR #689, opened 2020-11-28
Builtin scripts or engine without LUA!

wh1t3lord:
# Warning

This pull request is not finished yet, but you can see some progress of my work and how to integrate scripts to your project which can be similar to OpenXRay's architecture. (See commits history of this PR)

# What you achieve 

- Improved performance
- Full-stable system
- Easy to debug
- You can easily remove LUA from your engine

# Minuses 

- You must know C++ for understanding
- You don't have runtime changes and as the result you need to compile your solution every time  or only just xrGame project

But you can fix it and start to use Jumbo (Unity) build or to start using modules (required C++20/23).

# Simple roadmap

- Connect my scripts to OpenXRay
- Testing schemes
- Write up most of UI elements
- Trying to play game from start to the end

(This message can be updated at anytime)

### DIFF
--- a/src/xrGame/xrGame.vcxproj.filters
+++ b/src/xrGame/xrGame.vcxproj.filters
@@ -2254,6 +2254,15 @@
     <Filter Include="UI\Multiplayer\MM_MP\VersionList">
       <UniqueIdentifier>{04c90fd3-e247-4f21-b715-8dc556986837}</UniqueIdentifier>
     </Filter>
+    <Filter Include="Scripts">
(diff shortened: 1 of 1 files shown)


KaMeHb-UA:
And what about 3rd-party mods that edits predefined scripts or even just use lua scripts? Are there plans to drop its support?

Xottab-DUTY:
This is unofficial pull request, but we are interested in it, so I've documented how it can be it be integrated into OpenXRay, see #782.
In short, mods and Lua support will never be dropped.

Also, I will close it since it's stale.

PR #701, opened 2020-12-17
xrRender: fixed OpenGL errors queue behavior (close #647)

### LINKED ISSUE #647: Green screen instead of game intro
Когда начинаешь новую игру, вместо игровой заставки просто зелёный экран.

И такая же проблема в настройках. 
Снимок экрана_2020-07-12_17-50-52

**OS**
Slackware Current(x86_64)

**OpenXRay Version:**
730-july-preview

**OpenXRay Log:**
[openxray_nemial.log](https://github.com/OpenXRay/xray-16/files/4908882/openxray_nemial.log)

GOG сборка игры.

eanmos:
I've managed to reproduce the bug. When the green screen appears the following message is seen in log:

```bash
	intro_start intro_game
	Invalid video stream: 0x500
```

This leads to file `xrRenderGL/glSH_Texture.cpp`:

```cpp
            GLenum err = glGetError();
            if (err != GL_NO_ERROR)
            {
                Msg("Invalid video stream: 0x%x", err);
                xr_delete(pTheora);
                pSurface = 0;
            }
```

However the actual errors appears a few lines above:

```cpp
	Preload();
```

After this function call the OpenGL errors queue contains an `0x500` error. If we check what is happening in the end of this function:

```
	void CTexture::Preload()
	{
	    m_bumpmap = RImplementation.Resources->m_textures_description.GetBumpName(cName);
	    m_material = RImplementation.Resources->m_textures_description.GetMaterial(cName);

	    if (glGetError() != GL_NO_ERROR)
		    Msg("Warning: can't preload texture %s", cName.c_str());
	}
```

we will get

```bash
	Warning: can't preload texture water\water_flowing_nmap
```

So, I guess this texture is just broken.

This check actually fixes the bug because it clears the OpenGL errors queue and the following code could work correctly.

### DIFF
--- a/src/Layers/xrRenderGL/glSH_Texture.cpp
+++ b/src/Layers/xrRenderGL/glSH_Texture.cpp
@@ -147,6 +147,9 @@ void CTexture::Preload()
 {
     m_bumpmap = RImplementation.Resources->m_textures_description.GetBumpName(cName);
     m_material = RImplementation.Resources->m_textures_description.GetMaterial(cName);
+
+    if (glGetError() != GL_NO_ERROR)
+	    Msg("Warning: can't preload texture %s", cName.c_str());
 }
 
 void CTexture::Load()


eanmos:
By the way, I've noticed that after this fix a new video intro appeared among the other videos with AMD and Radeon logos at the very game beginning.

Xottab-DUTY:
Preload() doesn't touch OpenGL at all, it just calls internal engine procedures... It seems that error flag is set to an error somewhere else. Can you please tell me, what value `glGetError` returns?

eanmos:
@Xottab-DUTY, `0x500` — `GL_INVALID_ENUM`.

eanmos:
Hmm… yeah, it seems I hurried with the conclusion. The „fix” works even if we pop the error **before** `Preload` call.

vTurbine:
Try to run _Mixed_ configuration to enable `CHK_GL` macro. I believe the error is happening on loading some of s3c compressed textures

Xottab-DUTY:
Tip: in Linux, Debug configuration is equivalent to Mixed on Windows.

eanmos:
@Xottab-DUTY, the wiki I was following [states](https://github.com/OpenXRay/xray-16/wiki/%5BEN%5D-How-to-build-and-setup-on-Linux)

> To enable debugging:
>
> `cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo`
>
>Note that `-DCMAKE_BUILD_TYPE=Debug` used to be supported, but is no longer supported.

But I can see this in the `CMakeLists.txt`:

```cmake
if(${CMAKE_BUILD_TYPE} STREQUAL "Debug")
    add_definitions(-DDEBUG -DMIXED)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Og")
    set(LUA_LIBRARIES xrLuajit-debug)
endif()
```

And Russian wiki [states](https://github.com/OpenXRay/xray-16/wiki/%5BRU%5D-%D0%9A%D0%B0%D0%BA-%D1%81%D0%BE%D0%B1%D1%80%D0%B0%D1%82%D1%8C-%D0%B4%D0%B2%D0%B8%D0%B6%D0%BE%D0%BA-%D0%BD%D0%B0-Linux):

> `DCMAKE_BUILD_TYPE` - Тип сборки может быть:
>
> - `Debug` - Отладка
> - `Mixed` - Смешанная
> - `RelWithDebInfo` - Релиз с отладочными символами
> - `Release` - Релиз значение по умолчанию (скорее всего это ваш выбор)

What to believe? Should we fix wiki?

eanmos:
@vTurbine, you are right. I've compiled the engine with `Debug` and got the following fatal error:

```plaintext
* Loaded: \home\eanmos\.local\share\GSC Game World\S.T.A.L.K.E.R. - Call of Pripyat\gamedata\textures\water\water_sbumpvolume.dds[1048704]b
GL_INVALID_ENUM in glTexStorage3D(internalformat = GL_COMPRESSED_RGBA_S3TC_DXT5_EXT) 2
 
FATAL ERROR
 
[error] Expression    : glTexStorage3D(target, static_cast<GLint>(texture.levels()), format.Internal, tex_extent.x, tex_extent.y, tex_extent.z)
[error] Function      : texture_load
[error] File          : /home/eanmos/dev/xray-16/src/Layers/xrRenderPC_GL/glTexture.cpp
[error] Line          : 142
[error] Description   : (null)
```

the `CHK_GL` fails in function [`CRender::texture_load`](https://github.com/OpenXRay/xray-16/blob/dev/src/Layers/xrRenderPC_GL/glTexture.cpp#L142):

```cpp
        case gli::TARGET_3D:
            CHK_GL(glTexStorage3D(target, static_cast<GLint>(texture.levels()), format.Internal,
                           tex_extent.x, tex_extent.y, tex_extent.z));
            break;
```

For this texture (`water_sbumpvolume.dds`):

- `format` is `1` (`FORMAT_RG4_UNORM_PACK8`);
- `target` is `806f` (`TARGET_3D`);
- `format.Internal` is `83f3` (`INTERNAL_RGBA_DXT5`).

I don't know why OpenGL returns `GL_INVALID_ENUM` in this case because `GL_COMPRESSED_RGBA_S3TC_DXT5_EXT` seems to be a valid internal format according to the [OpenGL documentation](https://www.khronos.org/opengl/wiki/GLAPI/glTexStorage3D).

---

I've tried to open the texture file in GIMP and got a warning:

> Warning: `DDSD_LINEARSIZE` is incorrectly not set for `DDPF_FOURCC`!

But the texture opens anyway.

If I export the texture in some other DDS format and load it instead of the original one, there is no fatal error anymore.

---

Do you have some ideas how to deal with it?

vTurbine:
Did you try to re-save it in DXT5 again? If the texture is broken it's probably better to ignore this error while texture loading. Otherwise, I bet you have an Intel integrated GPUs which has some quirks in S3C support (e.g. you may have an S3C extension available while some of the formats will be refused to load).
What GPU and OS you're using?

Xottab-DUTY:
Firstly, we need to make `R_CHK` macro for OpenGL renderer that will work on Release configurations also...
Then, we need to find what we can do about errors with S3C format...

OPNA2608:
> Did you try to re-save it in DXT5 again? Otherwise, I bet you have an Intel integrated GPUs which has some quirks in S3C support (e.g. you may have an S3C extension available while some of the formats will be refused to load).
> What GPU and OS you're using?

I also experience this green screen instead of the intro video on the first New Game after game launch. Quitting back to the menu and clicking New Game again makes the intro play properly. I run a Radeon RX 5700 XT on Linux.

Loading the texture into GIMP (I got no warning like in https://github.com/OpenXRay/xray-16/pull/701#issuecomment-750300572) and re-exporting it as "RXGB (DXT5)" fixes the intro video for me.

KiralyCraft:
This PR fixed my initial green screen issue on Arch with AMD, many thanks!

r-a-sattarov:
@eanmos hi! Any news on this PR?

eanmos:
@r-a-sattarov, hi! Honestly, I don't really know what would be the best way to handle it: either make `R_CHK` macro that will work in release mode (as @Xottab-DUTY suggested above), or just manually clean OpenGL error queue after the texture is loaded and leave some comment explaining this issue.

However, IIUC the second solution won't fix the root of problem. IMO, it is better for the lead developer to decide what to do in this case.

Xottab-DUTY:
Doing some housekeeping and reducing number of stale pull requests...

Closing due to:
> Preload() doesn't touch OpenGL at all... It seems that error flag is set to an error somewhere else.

The issue won't be left as is, though, and I hope that we'll find the way to fix it.

PR #703, opened 2020-12-28
xrEngine: flush SDL events queue on console opening

vTurbine:


Xottab-DUTY:
Oh reaaaaallllly, that's the only thing it was missing? Not bad!

PR #706, opened 2021-01-03
Fixed RPM package missing info about provided dependencies (closes #650)

### LINKED ISSUE #650: libODE.so Unresolved dependencies
Im install latest version OpenXRay make a Fedora package and when i try to install this package i get an error
```
error: Failed dependencies:
	libODE.so()(64bit) is needed by OpenXRay-1.6.02-1.x86_64

```

007exe:
Fix Cpack, solution to circular dependency. Now, when installing the generated OpenXRay RPM package, the package manager will not ask you to install the libODE.so OPCODE.so dependencies ... that the OpenXRay package provides, as a result of which it was not possible to install the OpenXRay package.


cmake/packaging.cmake
```diff
@@ -33,7 +33,10 @@ if (UNIX)
 
         if (FEDORA_FOUND OR REDHAT_FOUND OR CENTOS_FOUND)
             set(CPACK_GENERATOR RPM)
-
+
```
Xottab-DUTY:
```suggestion

```

cmake/packaging.cmake
```diff
@@ -33,7 +33,9 @@ if (UNIX)
 
         if (FEDORA_FOUND OR REDHAT_FOUND OR CENTOS_FOUND)
             set(CPACK_GENERATOR RPM)
-
+            # -- set(CPACK_RPM_PACKAGE_AUTOREQPROV ON)
```
Xottab-DUTY, line 37:
```suggestion

            # -- set(CPACK_RPM_PACKAGE_AUTOREQPROV ON)
```

Xottab-DUTY (approved):

Issue #210, opened 2018-05-31
Support static linking


Xottab-DUTY:
Support compilation mode when all engine binary files will be in a just one executable file.
- [x] Link engine libraries statically
- [ ] Link static version of the standard library

Xottab-DUTY:
The main goal is done: static build is supported on Windows and other platforms.
Linking to a static standard library runtime is just a small TODO. It can be tracked in a closed issue :)

PR #712, opened 2021-01-08
Visual Studio build system refactoring

Xottab-DUTY:
This pull request unifies the build system and:
- Makes us easier to add static build support. (#210)
- Makes easier to raise warnings levels. (#713)

VS2017 and Windows SDK 8.1 are now deprecated and not supported until we can test if it actually works.

If VS2017 works, then we need to add checks for PlatformToolset and WindowsTargetPlatformVersion for VS2017 and Windows SDK 8.1 in the xray_project_configuration.props.


Xottab-DUTY:
@OpenXRay/xray-contributors any thoughts?

PR #726, opened 2021-02-04
Various tiny fixes

vTurbine:


src/Layers/xrRender/r__sector.h
```diff
@@ -22,8 +22,17 @@ class CPortal : public IRender_Portal
                 public pureRender
 #endif
 {
+public:
+    using Poly = svector<Fvector, 6>;
```
drug007, line 26: In original version 8 elements, not 6, is it intentional change?

vTurbine: Serialized portals are stored in a way that they can't have more than 6 vertices (see `b_portal` and loading code around). When loaded, their polygon shape doesn't change anymore. Except for clipping code (see `CSector::traverse()` "Clip by frustum" section) where we can have an additional vertex from intersection with a frustum plane.

vTurbine: Oh, shi.. `ClipPoly()` re-uses existing container, so probably it's a good idea to increase the size of clipped portal poly.

drug007: I see

Xottab-DUTY (approved):

PR #733, opened 2021-02-27
Added search of FreeImagePlus library and its include path

r-a-sattarov:


Xottab-DUTY:
@eagleivg, @q4a, @Masterkatze, please look. All is ok?

Masterkatze:
Is FreeImagePlus and FreeImage interchangeable or one depends on another? Shouldn't we also have `find_package(FreeImagePlus REQUIRED)` in that case?

Otherwise looks fine to me. Although I would use lowercase but that's another story :)

Xottab-DUTY:
> Is FreeImagePlus and FreeImage interchangeable or one depends on another?

FreeImage is written in C and FreeImagePlus is C++ wrapper, so second depends on first.

> Shouldn't we also have find_package(FreeImagePlus REQUIRED) in that case?

I think it will be better, yes.

Masterkatze:
> I think it will be better, yes.

I think it would be better to have separate `FindFreeImagePlus.cmake` file then.

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you!

PR #737, opened 2021-03-06
Refactor luajit.cmake to match with LuaJIT makefile

Masterkatze:


CMakeLists.txt
```diff
@@ -171,7 +173,6 @@ set(LUA_FOUND true)
 message("CMake build ${CMAKE_BUILD_TYPE}, CFLAGS=${CMAKE_C_FLAGS}, CXXFLAGS=${CMAKE_CXX_FLAGS}")
 
 add_definitions(-D_MT -D_CPPUNWIND -DPURE_DYNAMIC_CAST -DDECLARE_SPECIALIZATION -DM_NOSTDCONTAINERS_EXT)
-add_definitions(-DLUABIND_DYNAMIC_LINK) #self-explanatory
```
Xottab-DUTY: Luabind can be included from any engine module, so this has to be global.

Masterkatze: This define is present in `Externals/luabind/CMakeLists.txt:23`

Xottab-DUTY: Oh, ok. And it is defined for any module too?
*Like, when you build xrEngine, -DLUABIND_DYNAMIC_LINK should be defined in xrEngine too.*

Xottab-DUTY: Like, when you build xrEngine, -DLUABIND_DYNAMIC_LINK should be defined in xrEngine too.

Xottab-DUTY: @Masterkatze

Masterkatze: Hmm, I think it's better to put it back for now.

Xottab-DUTY (approved):

CMakeLists.txt
```diff
@@ -4,6 +4,11 @@ project(OpenXRay)
 
 set(CMAKE_CXX_STANDARD 17)
 
+set(CMAKE_VERBOSE_MAKEFILE ON)
+if (CMAKE_VERBOSE_MAKEFILE)
+	set(CMAKE_EXECUTE_PROCESS_COMMAND_ECHO STDOUT)
```
Xottab-DUTY: Should use spaces instead of tabs.
```suggestion
    set(CMAKE_EXECUTE_PROCESS_COMMAND_ECHO STDOUT)
```

CMakeLists.txt
```diff
@@ -173,7 +178,7 @@ set(LUA_FOUND true)
 message("CMake build ${CMAKE_BUILD_TYPE}, CFLAGS=${CMAKE_C_FLAGS}, CXXFLAGS=${CMAKE_CXX_FLAGS}")
 
 add_definitions(-D_MT -D_CPPUNWIND -DPURE_DYNAMIC_CAST -DDECLARE_SPECIALIZATION -DM_NOSTDCONTAINERS_EXT)
-add_definitions(-DLUABIND_DYNAMIC_LINK) #self-explanatory
+add_definitions(-DLUABIND_DYNAMIC_LINK)
```
Xottab-DUTY: Don't unnecessary diffs.
```suggestion
add_definitions(-DLUABIND_DYNAMIC_LINK) #self-explanatory
```

PR #738, opened 2021-03-08
Remove D3DX9 dependency from R2 renderer

HeapRaid:
This is the first step in reducing the dependency on D3DX. It completely eliminates the D3DX9 dependency from the R2 renderer and most of the R1 renderer. It paves the way for removing the D3DX11 dependency at a later point.

It's not possible to completely remove the D3DX9 dependency from the R1 renderer, because the D3DCompiler doesn't support the `ps_1_1` shader model.

A couple of important points:
- ~This removes the ability to save screenshots as JPEGs, they are now always saved as TGAs. JPEG support could be added back in later if needed.~
- This enables O3 optimizations for shaders on DX9 backends, if this results in issues an exception could be added back in.


Xottab-DUTY:
What about PNG screenshots? PNG or JPEG is needed.

HeapRaid:
> What about PNG screenshots? PNG or JPEG is needed.

Would it be enough to support PNG on Windows and TGA on Linux? DirectXTex can only save DDS and TGA images in a cross-platform way. The OpenGL renderer has its own screenshot code, so it only affects DX9 on Linux (eventually).

HeapRaid:
I've restored the ability to take JPEG screenshots on Windows, let me know if you want PNG instead.

Xottab-DUTY:
Thanks!

I reworked the changed from this PR and almost got rid of D3DX in R1/R2 renderers in these commits:
546bcdac80dd4185e51b65e04776168938b93b90
a3eebde180585d2e508d9de88d9345b1edcef3bd
08d0f0a543a46fa0285df30f8ab812c16cf3c562
20256656f5b3ee0b74e89ab28bea07d50e0d09c6
85c10545826e90967d361090e8e137d2bcbb81fa
ab2a5e57ee34855fed89acf1ccbf71dc802152c1
95b7e99acdaccf6627ae689144896e5ff850c8d3
4c291994e71f31e823160f19c04249e05b34d23b
b024f89da627f86f186da52b9db6532a02a03bd7
e9ec3d3c3a75948abaa18f01893840ba0e68c8cc
a9faefed5d4fbc4466a2bed65ffd062f067608c4
b6ca0584141c9b5c5e035ebcd1242790b515251d
361b5f440c7983f4775a41f4cf93935b322469e5
8ce179260a67b4e377f20ce81cd8ac67f928e56a
b25ba96baa0eb4c52101934328ca6549b9a123df
8a01df48b07b861f8c6617a631fa0a07cb655b55
3cb8f9262a64d48e9f6ba1659d1428f05998f4f6

Only changes from Texture.cpp are left unmerged, because this PR removes LOD management from unmaged textures. This is not appropriate, especially since I removed managed textures completely in 42c225870f40137722624dc1a7be9f2d8b919469.
So, Texture.cpp should be ported to DirectXTex differently

PR #742, opened 2021-03-11
CMakeLists: Fix compiler output not accounting for the build type.

HeapRaid:
This happens when generating for the first time without specifying a build type.

Also removes the debug lines added in fba27f2.


Xottab-DUTY:
Thanks!

PR #743, opened 2021-03-13
luajit.cmake: Clean up old buildvm_arch.h header

HeapRaid:
This header is now generated in the build directory. However if an old buildvm_arch.h header is still present in the source directory it will be preferred by the compiler over the newly generated header in the build directory.

Automatically deleting the old header fixes clones that have built commits with the old luajit.cmake


Xottab-DUTY:
The addressed issue is fixed in #744.

PR #761, opened 2021-03-31
Rename cpuid.h header, fix __cpuid call

Masterkatze:


src/xrCore/xr_cpuid.cpp
```diff
@@ -1,7 +1,11 @@
 #include "stdafx.h"
 #pragma hdrstop
 
-#include "cpuid.h"
+#if (defined(XR_PLATFORM_LINUX) || defined(XR_PLATFORM_FREEBSD)) && defined(XR_COMPILER_GCC)
+#include <cpuid.h>
+#endif
+
+#include "xr_cpuid.h"
```
Xottab-DUTY: This should go before system include

PR #762, opened 2021-04-01
CDB Cache: setting time for archive files

### LINKED ISSUE #714: CDB cache
It looks like the engine file system doesn't flush the files immediately to disk. If the engine will be closed before the cache is flushed to disk, the cache files will be broken.

CDB cache was added in #698.

- [x] Flush to disk immediately (ad881d6071c5f8b7e48e11db3086d6cad799af92)
	- [x] ~~Sub-task: Rescan CDB cache folder (#709)~~
		- [x] Rescan not needed, it was a bug fixed in ad881d6071c5f8b7e48e11db3086d6cad799af92
- [x] Version check (#727, #762)
- [x] Integrity check (a0f045012700ffeed8ee07d9ad0ca6ab9a6c7e7c)

jjdredd:
This fixes CDB cache has 0 in files timestamp, if the file is from archive.
Needs testing for windows.

### DIFF
--- a/src/xrCore/LocatorAPI.cpp
+++ b/src/xrCore/LocatorAPI.cpp
@@ -472,13 +472,14 @@ void CLocatorAPI::LoadArchive(archive& A, pcstr entrypoint)
         hdr->r(&ptr, sizeof(ptr)); // Obtain internal pointer to the file in archive
 
         strconcat(sizeof full, full, fs_entry_point, name);
-        Register(full, A.vfs_idx, header.crc, ptr, header.size_real, header.size_compr, 0);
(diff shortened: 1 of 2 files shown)


jjdredd:
To test on windows, run the engine with -cdb_cache switch, load into the game (any save, etc) and check the first
dword in files `hom.bin,  objspace.bin,  portals.bin` in `_appdata_/cdb_cache/<levelname>/` directory.
It should be non-zero.

jjdredd:
Would somebody care to test this?

jjdredd:
I have tested this on windows 10 : it works fine.

PR #765, opened 2021-04-06
Fix Linux build issue in dxPixEventWrapper macro

Masterkatze:


Xottab-DUTY:
Thank you for the effort!
It appeared that for OpenGL we don't need wchar_t, hence the build was fixed automatically when I implemented `*PixEvent` functions for OpenGL.

PR #767, opened 2021-04-07
Fix thread names not being set in Linux

Masterkatze:


Xottab-DUTY:
Names without X-Ray are noncanonical

src/xrCore/Threading/ThreadUtil.cpp
```diff
@@ -39,10 +38,21 @@ void SetThreadName(ThreadHandle threadHandle, pcstr name)
     }
 }
 
+void SetThreadName(ThreadHandle threadHandle, pcstr name)
```
Xottab-DUTY, line 48: I think, we don't even need this function, SetCurrentThreadName should be enough :)

src/xrCore/Threading/ThreadUtil.cpp
```diff
@@ -11,10 +11,9 @@ ThreadId GetCurrThreadId() { return GetCurrentThreadId(); }
 
 ThreadHandle GetCurrentThreadHandle() { return GetCurrentThread(); }
 
-void SetThreadName(ThreadHandle threadHandle, pcstr name)
+void SetThreadNameImpl(DWORD threadId, pcstr name)
```
drug007, line 15: What is the reason behind replacing `ThreadHandle` by `DWORD`? The last is not cross platform as I know.

Masterkatze: Nothing being replaced, it's just implementation moved in its own function, `SetThreadName` and `SetCurrentThreadName` still has `ThreadHandle` as first parameter type.

src/xrCore/Threading/ThreadUtil.cpp
Masterkatze: I was thinking about that too, but decided to leave it in case someone will need it.

src/xrCore/Threading/ThreadUtil.cpp
```diff
@@ -24,9 +24,13 @@ void SetThreadName(ThreadHandle threadHandle, pcstr name)
         DWORD dwFlags;
     };
 
+    auto fullNameSize = sizeof(name) + 6;
```
Xottab-DUTY: Please, replace `6` with something like, maybe, `std::size("X-Ray")`

src/xrCore/Threading/ThreadUtil.cpp
```diff
@@ -24,9 +24,15 @@ void SetThreadName(ThreadHandle threadHandle, pcstr name)
         DWORD dwFlags;
     };
 
+    auto namePrefix = "X-Ray ";
+    constexpr auto namePrefixSize = sizeof(namePrefix);
```
Xottab-DUTY: sizeof namePrefix will return pointer size, because namePrefix will be evaluated to `const char*` instead of `char[]`. At least, this is what IntelliSense says.
```suggestion
    constexpr char namePrefix[] = "X-Ray ";
    constexpr auto namePrefixSize = std::size(namePrefix) - 1;
```

src/xrCore/Threading/ThreadUtil.cpp
```diff
@@ -24,9 +24,15 @@ void SetThreadName(ThreadHandle threadHandle, pcstr name)
...
+    auto fullNameSize = sizeof(name) + namePrefixSize;
```
Xottab-DUTY: sizeof is wrong here
```suggestion
    auto fullNameSize = xr_strlen(name) + namePrefixSize;
```

src/xrCore/Threading/ThreadUtil.cpp
Masterkatze: It was working fine (I had tested it in debugger), but your version looks better. And also `- 1` is not needed here.

src/xrCore/Threading/ThreadUtil.cpp
```diff
@@ -24,9 +24,15 @@ void SetThreadName(ThreadHandle threadHandle, pcstr name)
         DWORD dwFlags;
     };
 
+    constexpr char namePrefix[] = "X-Ray ";
+    constexpr auto namePrefixSize = std::size(namePrefix);
```
Xottab-DUTY:
```suggestion
    constexpr auto namePrefixSize = std::size(namePrefix); // includes null-character intentionally
```

src/xrCore/Threading/ThreadUtil.cpp
```diff
@@ -24,9 +24,15 @@ void SetThreadName(ThreadHandle threadHandle, pcstr name)
         DWORD dwFlags;
     };
 
+    constexpr char namePrefix[] = "X-Ray ";
+    constexpr auto namePrefixSize = std::size(namePrefix); // Includes null-character intentionally
```
Xottab-DUTY: Very important xD
```suggestion
    constexpr auto namePrefixSize = std::size(namePrefix); // includes null-character intentionally
```

Xottab-DUTY (approved):

Xottab-DUTY (approved):

src/xrCore/Threading/ThreadUtil.cpp
```diff
@@ -92,19 +109,24 @@ ThreadHandle GetCurrentThreadHandle() { return pthread_self(); }
 
 void SetThreadName(ThreadHandle threadHandle, pcstr name)
 {
-    if (!threadHandle)
+    if (auto error = pthread_setname_np(threadHandle, name) != 0)
     {
-        pthread_setname_np(pthread_self(), name);
-        return;
+        Msg("SetThreadName: failed to set thread name to '%s'. Errno: '%s'", name, error);
```
dimhotepus: `Errno: '%s'` where `error` is `int`. Either use `%d` or `strerror{_s}`.

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you!

PR #773, opened 2021-04-16
CMake hacks for MCST e2k (Elbrus 2000) architecture

r-a-sattarov:
- luajit.cmake: added hack for MCST version of LuaJIT
- xrGame/CMakeLists: added hack for MCST lcc compiler

### DIFF
--- a/src/xrGame/CMakeLists.txt
+++ b/src/xrGame/CMakeLists.txt
@@ -2673,6 +2673,13 @@ target_compile_options(${PROJECT_NAME}
     -Wno-trigraphs
 )
 
+if ((CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo") AND PROJECT_PLATFORM_E2K) # E2K: hack for debug compile (MCST lcc compiler does not support a file > 4Gb, this is a bug in EDG frontend)
+    target_compile_options(${PROJECT_NAME} 
+        PRIVATE
+        -g0 
+    )
+endif()
+
 set_target_properties(${PROJECT_NAME} PROPERTIES
     PREFIX ""
 )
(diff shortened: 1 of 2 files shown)


src/xrGame/CMakeLists.txt
```diff
@@ -2673,6 +2673,13 @@ target_compile_options(${PROJECT_NAME}
     -Wno-trigraphs
 )
 
+if ((CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo") AND PROJECT_PLATFORM_E2K) # E2K: hack for debug compile (MCST lcc compiler does not support a file > 4Gb, this is a bug in EDG frontend)
```
Masterkatze, line 2676: Use generator expression please

Masterkatze (dismissed):

src/xrGame/CMakeLists.txt
a1batross: What's the purpose of using generator expression here?

I would replace the debug check to something else though, but I forgot CMake (as a scary nightmare)

Masterkatze: The purpose is that we don't need to have multiple `target_compile_options` for single target.

In result, it will look like this (needs testing):
```
target_compile_options(${PROJECT_NAME} 
    PRIVATE
    -Wno-trigraphs
    $<$<AND:$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>,$<BOOL:PROJECT_PLATFORM_E2K>>:-g0>
)
```

Xottab-DUTY: @Masterkatze, it looks unreadable. Readability is preferred.

Xottab-DUTY: Maybe, it is better to do something like this:
```cmake
if (PROJECT_PLATFORM_E2K)
    target_compile_options($generator_expression$)
endif()
```

Also, I think it is ok have additional target_compile_options for workarounds like this. It makes them more visible and indicates the possibility for cleanup in future, when the bug will probably be fixed.

Masterkatze: Yes, the readability is not the best one :) let's leave it as it is.

Xottab-DUTY (approved):
@Masterkatze, thank you!

PR #781, opened 2021-04-27
Fixed OOM at startup on E2K and ARM

kosumosu:
Not filling `cpui` and leaving trash in it for ARM and E2K architectures was causing program to crash on startup inside `query_processor_info()`. The program was eating away all available memory and then crashing.
Since trying to determine CPU features x86-way on alternative architectures makes no sense, added separate implementations for them.

Also fixed incorrectly specified memory order for store operation. [Order must be one of std::memory_order_relaxed, std::memory_order_release or std::memory_order_seq_cst. Otherwise the behavior is undefined.](https://en.cppreference.com/w/cpp/atomic/atomic/store) This was causing Windows debug build to crash with failed assertion inside `std::atomic`.

### DIFF
--- a/src/xrCore/Threading/TaskManager.cpp
+++ b/src/xrCore/Threading/TaskManager.cpp
@@ -275,7 +275,7 @@ void TaskManager::WakeUpIfNeeded()
                 continue;
             if (worker->sleeps.load(std::memory_order_relaxed))
             {
-                worker->steal_from.store(steal_from);
+                worker->steal_from.store(steal_from, std::memory_order_relaxed);
                 worker->event.Set();
                 break;
             }
--- a/src/xrCore/Threading/TaskManager.cpp
+++ b/src/xrCore/Threading/TaskManager.cpp
@@ -275,9 +275,7 @@ void TaskManager::WakeUpIfNeeded()
                 continue;
             if (worker->sleeps.load(std::memory_order_relaxed))
             {
-                // Acquire semantic guarantees that event won't be set
-                // earlier than steal_from will be assigned
-                worker->steal_from.store(steal_from, std::memory_order_acquire);
+                worker->steal_from.store(steal_from);
                 worker->event.Set();
                 break;
             }
--- a/src/xrCore/xrDebug.cpp
+++ b/src/xrCore/xrDebug.cpp
@@ -446,30 +446,34 @@ void xrDebug::GatherInfo(char* assertionInfo, size_t bufferSize, const ErrorLoca
     int nptrs = backtrace(array, 20);     // get void*'s for all entries on the stack
     char **strings = backtrace_symbols(array, nptrs);
 
-    if(strings)
+    if (strings)
     {
+        size_t demangledBufSize = 0;
+        char* demangledName = nullptr;
         for (size_t i = 0; i < nptrs; i++)
         {
             char* functionName = strings[i];
-            char* demangledName = nullptr;
+
             Dl_info info;
 
             if (dladdr(array[i], &info))
             {
-                int status = -1;
-                demangledName = abi::__cxa_demangle(info.dli_sname, nullptr,
-                    nullptr, &status);
-                if (status == 0)
+                if (info.dli_sname)
                 {
-                    functionName = demangledName;
+                    int status = -1;
+                    demangledName = abi::__cxa_demangle(info.dli_sname, demangledName, &demangledBufSize, &status);
+                    if (status == 0)
+                    {
+                        functionName = demangledName;
+                    }
                 }
             }
             Log(functionName);
-    #ifdef USE_OWN_ERROR_MESSAGE_WINDOW
+#ifdef USE_OWN_ERROR_MESSAGE_WINDOW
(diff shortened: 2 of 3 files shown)


kosumosu:
Билды упали сами по себе. Не виноватая я.

kosumosu:
There'll gonna be more changes for this.

src/xrCore/xrDebug.cpp
```diff
@@ -446,30 +446,34 @@ void xrDebug::GatherInfo(char* assertionInfo, size_t bufferSize, const ErrorLoca
     int nptrs = backtrace(array, 20);     // get void*'s for all entries on the stack
     char **strings = backtrace_symbols(array, nptrs);
 
-    if(strings)
+    if (strings)
     {
+        size_t demangledBufSize = 0;
```
Xottab-DUTY, line 451: Should be initialized with braces: `size_t demangledBufSize{};`

src/xrCore/xrDebug.cpp
```diff
@@ -446,30 +446,34 @@ void xrDebug::GatherInfo(char* assertionInfo, size_t bufferSize, const ErrorLoca
...
             }
             Log(functionName);
-    #ifdef USE_OWN_ERROR_MESSAGE_WINDOW
+#ifdef USE_OWN_ERROR_MESSAGE_WINDOW
             buffer += xr_sprintf(buffer, bufferSize, "%s\n", functionName);
-    #endif // USE_OWN_ERROR_MESSAGE_WINDOW
-            ::free(demangledName);
+#endif // USE_OWN_ERROR_MESSAGE_WINDOW
         }
+        ::free(demangledName);
```
Xottab-DUTY, line 476: Why outside of for loop?

src/xrCore/xrDebug.cpp
kosumosu: Are you sure this is more readable?

src/xrCore/xrDebug.cpp
kosumosu: Because `abi::__cxa_demangle()` can reuse buffer and realloc it if needed. Less allocations == more performance. We can now crash faster.

Xottab-DUTY: > We can now crash faster.

Got you. 😎

src/xrCore/xrDebug.cpp
Xottab-DUTY: Less readable, but it is my code style.

Well, let it as is for now then, don't want to force it right now =)

src/xrCore/Threading/TaskManager.cpp
```diff
@@ -275,9 +275,7 @@ void TaskManager::WakeUpIfNeeded()
                 continue;
             if (worker->sleeps.load(std::memory_order_relaxed))
             {
-                // Acquire semantic guarantees that event won't be set
-                // earlier than steal_from will be assigned
-                worker->steal_from.store(steal_from, std::memory_order_acquire);
+                worker->steal_from.store(steal_from);
```
Xottab-DUTY: Well, I'm not sure, but maybe `Set()` function call just won't reordered, so we can just relax... But I'm not sure.
```suggestion
                worker->steal_from.store(steal_from, std::memory_order_relaxed);
```

kosumosu: Well, i'm not sure neither, but if it won't reorder anyway, should we even specify any order manually? We can just leave the default strict ordering.

Xottab-DUTY: The lesser restrictions – the better for performance. Except, that it should always be safe. It is safe to lower the restrictions here.

src/xrCore/xr_cpuid.cpp
```diff
@@ -193,67 +259,71 @@ bool query_processor_info(processor_info* pinfo)
...
+    
 
-    u32 byteOffset = 0;
-    while (byteOffset + sizeof(SYSTEM_LOGICAL_PROCESSOR_INFORMATION) <= returnedLength)
-    {
-        switch (buffer->Relationship)
-        {
-            case RelationProcessorCore:
-                processorCoreCount++;
+    #if (defined(__MMX__))
```
Xottab-DUTY: `#` should be always to the left. Other part of the macro can be aligned, though.
Also excessive parentheses are unneeded.
```suggestion
#if defined(__MMX__)
```

Xottab-DUTY: Remove the alignment.

> Other part of the macro can be aligned, though.

By this sentence I mean alignment in nesting cases, like this:
```cpp
#if defined(DEFINE)
#   if defined(DEFINE2)
#       define ONE
#	else
#       define TWO
#	endif
#endif
```
But there's no nesting and `#` character should always stick to the left side.

Xottab-DUTY (approved):

Xottab-DUTY:
Do you have any other changes or this can be marked as ready and then merged?)

kosumosu:
Yes, let's merge.

Xottab-DUTY:
Thank you!

PR #791, opened 2021-05-08
xrCDB/Frustum.cpp: reduced testSphere_dirty function code size

xrombik:


src/xrCDB/Frustum.cpp
```diff
@@ -90,48 +90,8 @@ EFC_Visible CFrustum::testSphere(Fvector& c, float r, u32& test_mask) const
...
-            return FALSE;
-    case 1:
-        if (planes[0].classify(c) > r)
-            return FALSE;
-    case 0: break;
-    default: NODEFAULT;
-    }
-    return TRUE;
+    XR_ASSUME(p_count);
+    return planes[p_count - 1U].classify(c) <= r;
```
Xottab-DUTY: This actually makes it unsafe, because NODEFAULT won't be triggered.

Xottab-DUTY (changes requested):

src/xrCDB/Frustum.cpp
drug007: Also '0' should be possible value and return `true`

src/xrCDB/Frustum.cpp
```diff
@@ -90,48 +90,9 @@ EFC_Visible CFrustum::testSphere(Fvector& c, float r, u32& test_mask) const
...
-    case 1:
-        if (planes[0].classify(c) > r)
-            return FALSE;
-    case 0: break;
-    default: NODEFAULT;
-    }
-    return TRUE;
+    if (p_count == 0U)
+        return TRUE;
+    return planes[p_count - 1U].classify(c) <= r;
```
Xottab-DUTY: It is still unsafe

drug007: I don't know what `NODEFAULT` is but probably this:
```
if (p_count > 12)
    NODEFAULT;
```

Xottab-DUTY:
```suggestion
	switch (p_count)
    {
    case 12:
    case 11:
    case 10:
    case 9:
    case 8:
    case 7:
    case 6:
    case 5:
    case 4:
    case 3:
    case 2:
    case 1:
        if (planes[p_count - 1].classify(c) > r)
            return false;
    case 0: break;
    default: NODEFAULT;
    }
    return true;
```

src/xrCDB/Frustum.cpp
```diff
@@ -90,48 +90,26 @@ EFC_Visible CFrustum::testSphere(Fvector& c, float r, u32& test_mask) const
 
 BOOL CFrustum::testSphere_dirty(Fvector& c, float r) const
 {
-    switch (p_count)
+	switch (p_count)
```
Xottab-DUTY:
```suggestion
    switch (p_count)
```

src/xrCDB/Frustum.cpp
```diff
@@ -90,48 +90,9 @@ EFC_Visible CFrustum::testSphere(Fvector& c, float r, u32& test_mask) const
...
-    case 1:
-        if (planes[0].classify(c) > r)
-            return FALSE;
-    case 0: break;
-    default: NODEFAULT;
-    }
-    return TRUE;
+    if (p_count == 0U)
+        return TRUE;
+    return planes[p_count - 1U].classify(c) <= r;
```
Xottab-DUTY:
```suggestion
	VERIFY(p_count <= FRUSTUM_MAXPLANES); // '<=' is not a typo, this check is correct
    if (p_count == 0)
        return true;
    return planes[p_count - 1].classify(c) <= r;
```

PR #798, opened 2021-05-12
Try to fix resource linking

### LINKED ISSUE #786: Resource linking ignores installation prefix
**Describe the bug**
https://github.com/OpenXRay/xray-16/blob/705d957fbbbb01cc5b86e83708917a5d17c4428e/src/xrCore/xrCore.cpp#L265-L287

This code ignores the user-configurable prefix for the build to which these file & directories actually get installed to.

https://github.com/OpenXRay/xray-16/blob/705d957fbbbb01cc5b86e83708917a5d17c4428e/CMakeLists.txt#L15-L16

https://github.com/OpenXRay/xray-16/blob/705d957fbbbb01cc5b86e83708917a5d17c4428e/res/CMakeLists.txt#L3-L7

We currently need to patch this in Nixpkgs to point to our build's prefix instead.

https://github.com/NixOS/nixpkgs/blob/ce93c98ce2243a8228419fcfbe97fd87f3c5f99e/pkgs/games/openxray/default.nix#L57-L60

**To Reproduce**
1. Pass `-DCMAKE_INSTALL_PREFIX=/some/path/that/isnt/the/default` to CMake
2. Build & install OpenXRay
3. Launch OpenXRay in a way that would enter the mentioned block of code

The created symlinks should point to invalid (or at least for this particular build incorrect) paths.

**Expected behavior**
The code should use the prefix the user passed to CMake (e.g. `/nix/store/v9s3qr21cf15lx3lgp1x4d4k8pdcc2an-openxray-822-december-preview`) instead of hardcoding `/usr`.

**Screenshots**
n/a

**BugTrap error report**
n/a

 - OS: Ubuntu 18.04.5 (with Nixpkgs) & NixOS 20.09
 - OpenXRay build version: 822

**Additional context**
I think `CMAKE_INSTALL_FULL_DATAROOTDIR` is the CMake variable the fixed code should somehow get access to. `CMAKE_INSTALL_DATAROOTDIR` works for moving files but it may be a relative path, while `CMAKE_INSTALL_FULL_DATAROOTDIR` is guaranteed to be an absolute one.

OPNA2608:
Fixes #786.

Problems:
- CMake supports setting an installation prefix, but the convenience resource linking ignores it and hardcodes `/usr/share` instead
- The resource linking uses `lstat` to determine if a link already exists. `lstat` gives details about the link itself, not what it links to - `stat` will follow symlinks to give details about their targets. It can happen that the link target stops existing at some point (e.g. reinstallation to new directory) which leads to invalid symlinks. These can produce cryptic errors for users.

Solutions:
1. Hardcode the CMake data prefix's (`CMAKE_INSTALL_DATAROOTDIR`) absolute path variant instead of `/usr/share`. Its default value is `share` relative to `CMAKE_INSTALL_PREFIX`, so it should evaluate to the same path with the default prefix of OpenXRay.
2. Use `stat` instead of `lstat` to determine if a link is needed. If `stat` fails, start preparing the resource linking. We then check `lstat` next to determine if an invalid link already exists (link path exists, but what it links to doesn't) - unlink it if it does. Lastly we make a new link to the resource.


src/xrCore/xrCore.cpp
```diff
@@ -263,26 +263,49 @@ void xrCore::Initialize(pcstr _ApplicationName, pcstr commandLine, LogCallback c
...
+             * -> delete the symlink if it is.
+             * Then, make a new symlink to our resources.
+             * This doesn't account for other stat errors (EACCES, ENAMETOOLONG, ENOENT caused by missing path component) */
+            int res = stat(tmp, &statbuf);
+            if (res != 0)
+            {
+                ZeroMemory(&statbuf, sizeof(statbuf));
+                res = lstat(tmp, &statbuf);
+                if (res == 0)
+                    unlink(tmp);
```
Xottab-DUTY:
```suggestion
                    xr_unlink(tmp);
```

src/xrCore/xrCore.cpp
```diff
@@ -263,26 +263,49 @@ void xrCore::Initialize(pcstr _ApplicationName, pcstr commandLine, LogCallback c
...
+            if (res != 0)
             {
-                mkdir("gamedata", S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
-                mkdir("gamedata/shaders", S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
-                symlink("/usr/share/openxray/gamedata/shaders/gl", tmp);
+                ZeroMemory(&statbuf, sizeof(statbuf));
+                res = lstat(tmp, &statbuf);
+                if (res == 0)
+                {
+                    unlink(tmp);
```
Xottab-DUTY:
```suggestion
                    xr_unlink(tmp);
```

src/xrCore/xrCore.cpp
```diff
@@ -263,26 +263,47 @@ void xrCore::Initialize(pcstr _ApplicationName, pcstr commandLine, LogCallback c
...
 
         /* A final decision must be made regarding the changed resources. Since only OpenGL shaders remain mandatory for Linux for the entire trilogy,
-        * I propose adding shaders from /usr/share/openxray/gamedata/shaders so that we remove unnecessary questions from users who want to start
+        * I propose adding shaders from <CMAKE_INSTALL_FULL_DATAROOTDIR>/openxray/gamedata/shaders so that we remove unnecessary questions from users who want to start
         * the game using resources not from the proposed ~/.local/share/GSC Game World/Game in this case, this section of code can be safely removed */
         if (!strstr(Core.Params, "-fsltx"))
         {
             chdir(ApplicationPath);
-            string_path tmp;
+            const string_path install_dir = MACRO_TO_STRING(CMAKE_INSTALL_FULL_DATAROOTDIR);
```
Xottab-DUTY:
```suggestion
            constexpr pcstr install_dir = MACRO_TO_STRING(CMAKE_INSTALL_FULL_DATAROOTDIR);
```

Xottab-DUTY (commented):
Seems ok to me. @eagleivg and @007exe, please look 🙂

Xottab-DUTY:
Thank you!

PR #815, opened 2021-06-03
Linux: Disabled C++ exceptions when MASTER_GOLD is defined

Xottab-DUTY:


Xottab-DUTY:
Maybe, I'll reopen that in feature.

PR #824, opened 2021-06-15
Optimize explosions

FreeZoneMods:
We don't need to restart explosion if multiple GE_GRENADE_EXPLODE events arrive.

### DIFF
--- a/src/xrGame/Explosive.cpp
+++ b/src/xrGame/Explosive.cpp
@@ -563,16 +563,19 @@ void CExplosive::OnEvent(NET_Packet& P, u16 type)
     {
     case GE_GRENADE_EXPLODE:
     {
-        Fvector pos, normal;
-        u16 parent_id;
-        P.r_u16(parent_id);
-        P.r_vec3(pos);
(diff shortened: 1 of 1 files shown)


Xottab-DUTY:
How that can happen?

FreeZoneMods:
It looks like in the vanilla game you won't be able to see performance drop. However, there are several ways to see the problem.
1) Write a script which tries to detonate explosive item using calls of 'explode' method. Each call of this method generates a new explosion event.
2) Fix a bug in CGrenade::Hit which prevents detonation of grenades when new hit arrives - condition 'CExplosive::Initiator() == u16(-1)' is always false. But if we remove this line we'll have problems with exploding a large heaps of grenades - explosion of each grenade will produce a new hit (and therefore a new event) for the other grenades in the heap.

Xottab-DUTY:
> explosion of each grenade will produce a new hit (and therefore a new event) for the other grenades in the heap.

So, it will trigger a new explosion event, when the grenade is already exploding?

Issue #570, opened 2020-04-05
Path to fsgame.ltx corrupted


xrombik:
This output:
```command line -fsltx /mnt/sda2/Programs/Steam/steamapps/common/Stalker Call of Pripyat/fsgame.ltx ```

and then this in same time:
```using fs-ltx /mnt/sda2/Programs/Steam/steamapps/common/Stalker```
Shows that path to `fsgame.ltx` is corrupted at runtime - it was split by white space.

<details> 
<summary> Console output </summary>
$:./run.sh
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/mnt/sdc1/stalkercop1/usr/games
LD_LIBRARY_PATH=:/mnt/sdc1/stalkercop1/usr/lib:/mnt/sdc1/stalkercop1/lib/mimalloc-1.6
! Couldn't create surface from image: Couldn't open logo.bmp
OpenXRay Rx64 build 7729, Apr  5 2020
Custom build from commit[34cd539079a836703401bca64bab435717c9b149] branch[xd_dev]
 
command line -fsltx /mnt/sda2/Programs/Steam/steamapps/common/Stalker Call of Pripyat/fsgame.ltx 

* CPU features: RDTSC, MMX, SSE, SSE2, SSE3, SSE4.1
* CPU cores/threads: 4/4
* CPU current freq: 3780 MHz

Initializing File System...
using fs-ltx /mnt/sda2/Programs/Steam/steamapps/common/Stalker
 
FATAL ERROR
 
[error] Expression    : file_size == (u32)r_bytes
[error] Function      : FileDownload
[error] File          : /mnt/sdc1/openxray/xray-16/src/xrCore/FS.cpp
[error] Line          : 146
[error] Description   : can't read from file : 
[error] Arguments     : \mnt\sda2\Programs\Steam\steamapps\
 

stack trace:

xrDebug::GatherInfo(char*, unsigned long, ErrorLocation const&, char const*, char const*, char const*, char const*)
xrDebug::Fail(bool&, ErrorLocation const&, char const*, char const*, char const*, char const*)
/mnt/sdc1/stalkercop1/usr/lib/xrCore.so(+0x2a822) [0x7f908d86e822]
CLocatorAPI::setup_fs_ltx(char const*)
CLocatorAPI::_initialize(unsigned int, char const*, char const*)
[...]

PR #846, opened 2021-07-12
command line options, started transition (#570)

### LINKED ISSUE #316: Sparse command line argument parsing with virtually no documentation on arguments
The engine has many commandline arguments, but they all are parsed "in place" by using strstr.

Can we create a single struct with all the flags and params pre-parsed?

This may require additional dependency or even code generation tool to make it work efficiently, but will give us well documented cmd arguments and will concentrate all the parsing in one place, will make code more clear.

Pros:
* no arguments-related `strstr` calls in code -> clear code and no ambiguity (i. e. when you have something like this: `./xr_3da -fsltx ../blablabla-globus/fsgame.ltx` this will trigger -gl option on windows which may cause unwanted behavior)
* documentation for arguments and `--help` option can be easily implemented
* this may be efective in some usecases of `strstr`-based parsing

Cons:
* This will require many lines of dull copy-paste code which we'll have to support
* or to add another dependency (which also will not be as efficient as just a struct with fields)
* or to use code-generation tool (which may bring more problems)

jjdredd:
I wish to discuss these changes. New system for handling command line options.

While  transitioning to the news system, I noticed that this approach is very inconvenient when
adding command line options that are used across different subsystems.
One has to define a public variable and declare it external in other places where it's used,
keeping in mind that the unit where external declaration is used must be compiled *after* 
the unit where the variable is defined. This is not very readable and extendable, although
not many options need to be used across different modules.

Another option is to have a public variable containing a map of command line options and access it like so:
`command_line_options_map.getvalue("-myoption")`.
This could lead to some overhead.

This PR is just for discussion.

### DIFF
--- a/src/xrCDB/xrCDB.cpp
+++ b/src/xrCDB/xrCDB.cpp
@@ -5,6 +5,7 @@
 
 #include "xrCDB.h"
 #include "xrCore/Threading/Lock.hpp"
+#include "xrCore/command_line_key.h"
 
 namespace Opcode
 {
@@ -13,6 +14,8 @@ namespace Opcode
 
 using namespace CDB;
 using namespace Opcode;
+
+static command_line_key<bool> mt_cdb("-mt_cdb", "mt_cdb", false);
 
 // Model building
 MODEL::MODEL() :
@@ -81,7 +84,7 @@ void MODEL::build(Fvector* V, int Vcnt, TRI* T, int Tcnt, build_callback* bc, vo
 #ifdef _EDITOR
     build_internal(V, Vcnt, T, Tcnt, bc, bcp);
 #else
-    if (!strstr(Core.Params, "-mt_cdb"))
+    if (!mt_cdb.OptionValue())
     {
         build_internal(V, Vcnt, T, Tcnt, bc, bcp);
         status = S_READY;
--- a/src/xrCDB/xrCDB.cpp
+++ b/src/xrCDB/xrCDB.cpp
@@ -5,7 +5,6 @@
 
 #include "xrCDB.h"
 #include "xrCore/Threading/Lock.hpp"
-#include "xrCore/command_line_key.h"
 
 namespace Opcode
 {
--- a/src/xrCDB/xrCDB.cpp
+++ b/src/xrCDB/xrCDB.cpp
@@ -14,7 +14,7 @@ namespace Opcode
 using namespace CDB;
 using namespace Opcode;
 
-static command_line_key<bool> mt_cdb("-mt_cdb", "mt_cdb", false);
+static command_line_key<bool> mt_cdb("mt_cdb", "mt_cdb", false);
 
 // Model building
 MODEL::MODEL() :
--- a/src/xrCore/command_line_key.cpp
+++ b/src/xrCore/command_line_key.cpp
@@ -93,7 +93,7 @@ void command_line_key<T>::PrintHelp()
     while (current_node != nullptr)
     {
         pcstr isreq = current_node->required ? "(mandatory)" : "(optional)";
-        Msg("%-20s \t %-10s \t %-25s", current_node->option_name,
+        Msg("-%-20s \t %-10s \t %-25s", current_node->option_name,
             isreq, current_node->description);
         current_node = current_node->l_next;
     }
@@ -110,9 +110,7 @@ bool ParseCommandLine(int argc, char** argv)
             continue;
         }
 
-        pcstr current_arg = argv[n];
-        // this is for later
-        // pcstr current_arg = &argv[n][1];
+        pcstr current_arg = &argv[n][1];
 
         // is this a bool option?
         if (auto clkey = command_line_key<bool>::find_option(current_arg))
--- a/src/xrCore/command_line_key.cpp
+++ b/src/xrCore/command_line_key.cpp
@@ -1,0 +1,216 @@
+#include "command_line_key.h"
+#include "xrMemory.h"
+#include "log.h"
+
+template<> command_line_key<bool> *command_line_key<bool>::l_head = nullptr;
+template<> command_line_key<int> *command_line_key<int>::l_head = nullptr;
+template<> command_line_key<pstr> *command_line_key<pstr>::l_head = nullptr;
+
+// command_line_key
+
+// when adding make sure it's beginning with a '-'
+template<typename T>
+command_line_key<T>::command_line_key(pcstr opname, pcstr desc, const T defval, bool req)
+    : provided(false), required(req)
+{
+    if(!IsOptionFlag(opname))
+    {
+        size_t opsz = xr_strlen(opname) + 1;
+        option_name = xr_alloc<char>(opsz);
+        option_name[0] = '-';
+        xr_strcpy(&option_name[1], opsz - 1, opname);
+    }
+    else
+    {
+        option_name = xr_strdup(opname);
+    }
+    description = xr_strdup(desc);
+    copy_argument(defval);
+
+    // add the option list node
+    l_next = l_head;
+    l_head = this;
+}
+
+
+template<typename T>
+void command_line_key<T>::copy_argument(T arg) {
+    argument = arg;
+}
+
+template<>
+void command_line_key<pstr>::copy_argument(pstr arg) {
+    argument = xr_strdup(arg);
+}
+
+template<typename T>
+void command_line_key<T>::free_argument() {
+    return;
+}
+
+template<>
+void command_line_key<pstr>::free_argument() {
+    xr_free(argument);
+}
+
+
+template<typename T>
+command_line_key<T> *command_line_key<T>::find_option(pcstr flag_name)
+{
+    command_line_key<T> *current_node =  command_line_key<T>::l_head;
+    while(current_node != nullptr)
+    {
+        if (!xr_strcmp(current_node->option_name, flag_name)) break;
+        current_node = current_node->l_next;
+    }
+    return current_node;
+}
+
+template<typename T>
+command_line_key<T>::~command_line_key()
+{
+    xr_free(option_name);
+    xr_free(description);
+    free_argument();
+
+    command_line_key<T> *current_node =  command_line_key<T>::l_head;
+    command_line_key<T> *prev_node = nullptr;
+    while(current_node != nullptr)
+    {
+        if (current_node == this) break;
+        prev_node = current_node;
+        current_node = current_node->l_next;
+    }
+
+    if(!current_node)
+        return;
+
+    if(prev_node)
+        prev_node->l_next = current_node->l_next;
+    else
+        command_line_key<T>::l_head = current_node->l_next;
+}
+
+template<typename T>
+bool command_line_key<T>::IsProvided() const
+{
+    return provided;
+}
+
+template<typename T>
+T command_line_key<T>::OptionValue() const
+{
+    return argument;
+}
+
+template<typename T>
+bool command_line_key<T>::CheckArguments()
+{
+    command_line_key<T> *current_node =  command_line_key<T>::l_head;
+    while(current_node != nullptr)
+    {
+        if(current_node->required && !current_node->provided)
+        {
+            Log("Error: Required option %s", current_node->option_name);
+            return false;
+        }
+        current_node = current_node->l_next;
+    }
+    return true;
+}
+
+template<>
+bool command_line_key<bool>::parse_option(pcstr option, pcstr arg)
+{
+    auto clkey = command_line_key<bool>::find_option(option);
+    if(!clkey) return false; // not found
+
+    clkey->argument = true;
+    clkey->provided = true;
+    return true;                // found and set
+}
+
+template<>
+bool command_line_key<int>::parse_option(pcstr option, pcstr arg)
+{
+    auto clkey = command_line_key<int>::find_option(option);
+    if(!clkey) return false; // not found
+
+    clkey->argument = std::stoi(arg); // may throw
+    clkey->provided = true;
+    return true;
+}
+
+template<>
+bool command_line_key<pstr>::parse_option(pcstr option, pcstr arg)
+{
+    auto clkey = command_line_key<pstr>::find_option(option);
+    if(!clkey) return false; // not found
+
+    clkey->argument = xr_strdup(arg);
+    clkey->provided = true;
+    return true;
+}
+
+template<typename T>
+void command_line_key<T>::PrintHelp() {
+    command_line_key<T> *current_node = command_line_key<T>::l_head;
+    while(current_node != nullptr)
+    {
+        pcstr isreq = current_node->required ? "(mandatory)" : "(optional)";
+        Msg("%-10s \t %-10s \t %-25s", current_node->option_name,
+            isreq, current_node->description);
+    }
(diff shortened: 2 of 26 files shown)


jjdredd:
The key files to look at:
```
src/xrCore/CLOptions.h
src/xrCore/CLOptions.cpp
src/xrCore/xrCore.cpp
src/xrCore/LocatorAPI.cpp
src/xrEngine/main.cpp
src/xr_3da/entry_point.cpp
```

vTurbine:
What you've done is just re-invention of [cxxopts](https://github.com/jarro2783/cxxopts)
The initial code has an issue (e.g. `MODEL::build()`) where an option gets checked every function call. In this case, we'd better cache its value in the class constructor/init function. With this, I don't care about any overhead added by shared map/string/whatever since all cmd line args will be parsed at the early module initialization only.

jjdredd:
> What you've done is just re-invention of [cxxopts](https://github.com/jarro2783/cxxopts)

This isn't my design, but I haven't heard of cxxopts

> The initial code has an issue (e.g. `MODEL::build()`) where an option gets checked every function call. In this case, we'd better cache its value in the class constructor/init function. With this, I don't care about any overhead added by shared map/string/whatever since all cmd line args will be parsed at the early module initialization only.

Hm, I wouldn't like to go through all classes and cache the relevant options as this would require a lot of refactoring.
Although this could be a little neater solution.

Edit:
I've looked at cxxopts and their designd seems to be quite different from the design in this patch.

jjdredd:
I think declaring a command line option as global variable in a header is dangerous, since a header may be included in several different compilation units and there will be multiple copies of that global variable.

Xottab-DUTY:
It's should not be an issue if you declare the variable as inline.

src/xrCore/command_line_key.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+template<typename T>
+command_line_key<T>::~command_line_key()
+{
+    xr_free(option_name);
+    xr_free(description);
+    free_argument();
+
+    command_line_key<T> *current_node =  command_line_key<T>::l_head;
```
Xottab-DUTY:
```suggestion
    auto current_node =  command_line_key<T>::l_head;
```

src/xrCore/command_line_key.cpp
```diff
@@ -0,0 +1,216 @@
...
+    command_line_key<T> *prev_node = nullptr;
+    while(current_node != nullptr)
```
Xottab-DUTY:
```suggestion
    while (current_node != nullptr)
```

src/xrCore/command_line_key.cpp
```diff
@@ -0,0 +1,216 @@
...
+    {
+        if (current_node == this) break;
```
Xottab-DUTY:
```suggestion
        if (current_node == this)
            break;
```

src/xrCore/command_line_key.cpp
```diff
@@ -0,0 +1,216 @@
...
+        prev_node = current_node;
+        current_node = current_node->l_next;
+    }
+
+    if(!current_node)
```
Xottab-DUTY:
```suggestion
    if (!current_node)
```

src/xrCore/command_line_key.cpp
```diff
@@ -0,0 +1,216 @@
...
+template<typename T>
+T command_line_key<T>::OptionValue() const
+{
+    return argument;
+}
+
+template<typename T>
+bool command_line_key<T>::CheckArguments()
+{
+    command_line_key<T> *current_node =  command_line_key<T>::l_head;
```
Xottab-DUTY:
```suggestion
    auto current_node = command_line_key<T>::l_head;
```

src/xrCore/command_line_key.cpp
```diff
@@ -0,0 +1,216 @@
...
+    clkey->argument = std::stoi(arg); // may throw
+    clkey->provided = true;
+    return true;
+}
+
+template<>
+bool command_line_key<pstr>::parse_option(pcstr option, pcstr arg)
+{
+    auto clkey = command_line_key<pstr>::find_option(option);
+    if(!clkey) return false; // not found
```
Xottab-DUTY:
```suggestion
    if (!clkey)
        return false; // not found
```

src/xrCore/command_line_key.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+
+template<typename T>
+void command_line_key<T>::copy_argument(T arg) {
+    argument = arg;
+}
+
+template<>
+void command_line_key<pstr>::copy_argument(pstr arg) {
```
Xottab-DUTY:
```suggestion
void command_line_key<pstr>::copy_argument(pstr arg)
{
```

src/xrCore/command_line_key.cpp
```diff
@@ -0,0 +1,216 @@
...
+    description = xr_strdup(desc);
+    copy_argument(defval);
+
+    // add the option list node
+    l_next = l_head;
+    l_head = this;
+}
+
+
+template<typename T>
```
Xottab-DUTY:
```suggestion
template < typename T >
```

Xottab-DUTY (commented):
Стиль кода надо привести к нашему

src/xrCore/command_line_key.h
```diff
@@ -0,0 +1,44 @@
...
+    static bool parse_option(pcstr option, pcstr arg);
+    static command_line_key<T> *l_head;
+};
+
+inline static bool IsOptionFlag(const char *buf)
+{
+    return (buf && buf[0] == '-');
+}
+
+XRCORE_API bool ParseCommandLine(int argc, char *argv[]);
```
Xottab-DUTY:
```suggestion
XRCORE_API bool ParseCommandLine(int argc, char* argv[]);
```

src/xrCore/xrCore.cpp
```diff
@@ -29,6 +30,18 @@ extern compression::ppmd::stream* trained_model;
 
 XRCORE_API xrCore Core;
 
+XRCORE_API command_line_key<pcstr> fsltx_path("-fsltx", "path to game config ltx", "");
```
Xottab-DUTY: It should be defined in LocatorAPI.cpp, and mentioned here as `extern`

src/xr_3da/entry_point.cpp
```diff
@@ -10,6 +13,7 @@
 #endif
 #include "xrEngine/main.h"
 #include "xrEngine/splash.h"
+#include "xrCore/command_line_key.h"
```
Xottab-DUTY: Just include this include into `xrCore.h` and then we don't need to include it everywhere anymore

jjdredd: Is it ok if after this I'll have to include xrCore.h in the source files it wasn't included before?

Xottab-DUTY: Hm, what file?

jjdredd:
What's happening with CI?

src/xr_3da/entry_point.cpp
jjdredd: LocatorAPI.cpp
LocatorAPI_auth.cpp
log.cpp
xrDebug.cpp
xrMemory.cpp

jjdredd:
I think this PR is almost ready for the second review. There are two things that I'm not sure of:

1. I don't think it's not sufficient to just add command_line_key.h into xrCore.h (see above comment)
2. I'm thinking about  getting rid of dynamic memory allocation. I think a static variable of 20 bytes is enough to store the name of the key and the oxr string2048 or string4096 to store the provided int/string argument. This would waste a few Kb of memory. If this is ok then I'll just use static variable. command line keys like `-start` isn't an issue since this is a separate case and needs another approach like passing a string with new keys to the command_line_key parser once again - we should discuss this case more.

src/xrCore/command_line_key.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+
+#include "xrCommon/xr_list.h"
+#include "_std_extensions.h"
+
+template < typename T > class XRCORE_API command_line_key;
+
```
Xottab-DUTY:
```suggestion
```

src/xrCore/command_line_key.h
```diff
@@ -0,0 +1,44 @@
...
+    bool IsProvided() const;          // was the option provided in the CLI?
+    T OptionValue() const;            // value provided with the option
+
+    static bool CheckArguments();
+    static void PrintHelp();
+    friend XRCORE_API bool ParseCommandLine(int argc, char* argv[]);
+
+private:
+    pstr option_name = nullptr;
+    pstr description = nullptr;
```
Xottab-DUTY: There's absolutely no reason for these variables to be pstr. If there is, eliminate that reason.
```suggestion
    pcstr option_name = nullptr;
    pcstr description = nullptr;
```

src/xrCDB/xrCDB.cpp
```diff
@@ -14,6 +15,8 @@ namespace Opcode
 using namespace CDB;
 using namespace Opcode;
 
+static command_line_key<bool> mt_cdb("-mt_cdb", "mt_cdb", false);
```
Xottab-DUTY: Ideally, the name of the key should be specified without `-` part. (`"mt_cdb"` instead of `-mt_cdb`)
And let the code in command_line_key.cpp check for that `-` automatically.

Xottab-DUTY (commented):
> I think this PR is almost ready for the second review. There are two things that I'm not sure of:
> 
> 1. I don't think it's not sufficient to just add command_line_key.h into xrCore.h (see above comment)
> 2. I'm thinking about  getting rid of dynamic memory allocation. I think a static variable of 20 bytes is enough to store the name of the key and the oxr string2048 or string4096 to store the provided int/string argument. This would waste a few Kb of memory. If this is ok then I'll just use static variable. command line keys like `-start` isn't an issue since this is a separate case and needs another approach like passing a string with new keys to the command_line_key parser once again - we should discuss this case more.

1. Let's continue this discussion in the comment above :)
2. It is possible to remove memory allocations and use `pcstr`. Please, use `pcstr` without memory allocations.

src/xr_3da/entry_point.cpp
Xottab-DUTY: xrCore.h is already included in stdafx.h, which is included to these files. There shouldn't be any problems with this...

jjdredd:
>     2\. It is possible to remove memory allocations and use `pcstr`. Please, use `pcstr` without memory allocations.

Wait, I don't get it. I thought I'm supposed to use a string stored in the command_line_key class like string2048?
If it's just a pointer, then I  will have to to allocate memory for that string.
Keep in mind that I'll need to modify the provided string (option name) to prepend a `-`.

As for the key value (if it requires a string value) I can just copy the pointer to the string from `char **argv`. I'm  not sure
if windows provides null-terminated strings so that could be a problem.

src/xrCDB/xrCDB.cpp
jjdredd: ok, so this line should actually look like `static command_line_key<bool> mt_cdb("mt_cdb", "mt_cdb", false);`?
This should already be possible with current implementation (see constructor).
There are options in the engine that actually do not start with a `-` (i think mostly for debug stuff).

src/xr_3da/entry_point.cpp
jjdredd: In that case I'll just include it in xrCore.h.
Will it also work for linux?

Xottab-DUTY:
> > ```
> > 2\. It is possible to remove memory allocations and use `pcstr`. Please, use `pcstr` without memory allocations.
> > ```
> 
> Wait, I don't get it. I thought I'm supposed to use a string stored in the command_line_key class like string2048?
> If it's just a pointer, then I will have to to allocate memory for that string.
> Keep in mind that I'll need to modify the provided string (option name) to prepend a `-`.
> 
> As for the key value (if it requires a string value) I can just copy the pointer to the string from `char **argv`. I'm not sure
> if windows provides null-terminated strings so that could be a problem.

Ah, yea. You need to store the value... For now, let's try to use `string1024`.

> Keep in mind that I'll need to modify the provided string (option name) to prepend a -.

To avoid modification of the original string, you can use temporary string on the stack in the initialization function.

jjdredd:
> To avoid modification of the original string, you can use temporary string on the stack in the initialization function.

Hm, I could just store the key name without the `-` and imply that it's there when parsing the command line, but it's still
simpler and more readable, explicit for the programmers to store the key name as a `string32` in the command_line_key class.
It's not that much resources, although could be redundant. Either way is possible.

I also did not want to assume that arguments to the `command_line_key` constructor have static lifetime.
This could be useful if you're adding a key and then reparsing a new command line again. I think there are cases when
they dynamically add keys to command line in the engine or utils, the current architecture (old Core.Params) allow this.

Also: modifying the command line key name and storing it in the class as `string32` is more convenient to account
for both cases:
when we call the constructor with a key name *with* a `-` and *without* a `-`, so that
for example both cases:
`static command_line_key<bool> mt_cdb("mt_cdb", "mt_cdb", false);`
and
`static command_line_key<bool> mt_cdb("-mt_cdb", "mt_cdb", false);`
are valid. This will reduce the possiblity of mistakes later on.

Xottab-DUTY:
All keys should have static lifetime and no reparsing is needed.

Key name should be pcstr.

jjdredd:
ok, it's not hard to change then. we'll require the programmer to provide the key name with static lifetime and with no `-`
in the key name.

jjdredd:
Для хранения переданных строк с ключом тоже использую указатель просто копируя соответствующий из argv.
Сквашну естественно после апрува.

russkel:
Excuse the ignorance, but wouldn't it be better adopting existing C++ libraries for this and reducing the amount of code in the engine itself? The more custom code that can be "outsourced" to other libraries means less maintenance work and less learning how this particular engine does it differently.

jjdredd:
> Excuse the ignorance, but wouldn't it be better adopting existing C++ libraries

That has been discussed.

russkel:
> > Excuse the ignorance, but wouldn't it be better adopting existing C++ libraries
> 
> That has been discussed.

Could you please link me to this discussion?

Xottab-DUTY:
> > > Excuse the ignorance, but wouldn't it be better adopting existing C++ libraries
> > 
> > 
> > That has been discussed.
> 
> Could you please link me to this discussion?

Oh, it was somewhere in our Discord server and in Russian... Shortly, I didn't find any 3rd party library with the required design.
The requirement was locality, the ability to declare a key as a single entity right in the file where this key is used.

russkel:
> > > > Excuse the ignorance, but wouldn't it be better adopting existing C++ libraries
> > > 
> > > 
> > > That has been discussed.
> > 
> > 
> > Could you please link me to this discussion?
> 
> Oh, it was somewhere in our Discord server and in Russian... Shortly, I didn't find any 3rd party library with the required design. The requirement was locality, the ability to declare a key as a single entity right in the file where this key is used.

Thanks. That makes sense!

jjdredd:
After a year and a half I think it's safe to say that the project is dead

Xottab-DUTY:
I remember about you, sorry for silence.

PR #859, opened 2021-07-29
Discord Rich Presence adaptation

doliinyk:
At the moment shows in your Discord game status engine name, location name or not if you in menu, spent time


src/xrGame/MainMenu.cpp
```diff
@@ -235,7 +237,9 @@ void CMainMenu::Activate(bool bActivate)
             }
             Device.seqRender.Remove(g_pGameLevel);
             CCameraManager::ResetPP();
-        };
+        }
+        else
+            g_DiscordRPC->Update("Главное меню");
```
collectioner: I think it'll be better to move this constant string to config file, since we support not only Russian language.
cc: @Xottab-DUTY what do you think?

src/xrGame/GamePersistent.cpp
```diff
@@ -612,6 +613,8 @@ extern CUISequencer* g_tutorial2;
 
 void CGamePersistent::OnFrame()
 {
+    
+
```
collectioner: Please, remove these empty lines.

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,35 @@
...
+DiscordRPC::~DiscordRPC()
+{
+    Discord_Shutdown();
+    Discord_ClearPresence();
+}
+
+void DiscordRPC::Update(const char* level_name)
+{
+    DiscordRichPresence discordPresence;
+    memset(&discordPresence, 0, sizeof(discordPresence));
```
collectioner: Not sure, but shouldn't we clear the memory after this variable usage?

src/xrGame/GamePersistent.cpp
```diff
@@ -938,6 +941,22 @@ void CGamePersistent::LoadTitle(bool change_tip, shared_str map_name)
...
+                temp_map_name = "Затон";
+            else if (map_name == "jupiter")
+                temp_map_name = "Окрестности «Юпитера»";
+            else if (map_name == "jupiter_underground")
+                temp_map_name = "Путепровод «Припять-1»";
+            else if (map_name == "pripyat")
+                temp_map_name = "Припять";
+            else if (map_name == "labx8")
+                temp_map_name = "Лаборатория X8";
+            g_DiscordRPC->Update(temp_map_name);
```
collectioner: Again, that's fine for the first version. But I'd like to see these things in config files with localization. I'll leave priority of this for @Xottab-DUTY .

src/xrGame/MainMenu.cpp
```diff
@@ -203,6 +204,7 @@ void CMainMenu::Activate(bool bActivate)
 
     if (bActivate)
     {
+        g_DiscordRPC->Init();
```
collectioner: Does it mean we initialize RPC every time we open main menu (well, when menu is active)?

sdk/include/DiscordRPC/discord_register.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
```
collectioner: @DenysOliinyk am I right that these 2 files (`discord_register.h` and `discord_rpc.`) are from Discord itself?

collectioner (changes requested):

sdk/include/DiscordRPC/discord_register.h
Xottab-DUTY: @collectioner, yes, they are.

doliinyk: @collectioner yup, from their official repo

collectioner: Thanks you both, I resolved the comment.

src/xrGame/MainMenu.cpp
Xottab-DUTY: Yes. And it is totally wrong. Please, call this function in the constructor of CGamePersistent class, right before Ansel initialization.

src/xrGame/MainMenu.cpp
Xottab-DUTY: The game may even have that string translated (we need to find if that's true), so we can use it.

Generally, yes, we need to use translation, not just string hardcoded in the engine :)

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,35 @@
+#include "stdafx.h"
+#include "DiscordRPC.h"
+#include <ctime>
+
+DiscordRPC* g_DiscordRPC;
+
+void DiscordRPC::Init()
+{
+    DiscordEventHandlers Handle;
+    memset(&Handle, 0, sizeof(Handle));
```
Xottab-DUTY: 1. X-Ray uses ZeroMemory function for zeroing memory instead of memset and we use it too.
2. We don't even need manual memory zeroing. Just uses braces initializer;
3. First letter for local variable name should be lowercase. (we use `camelCase` for local variables. `snake_case` is allowed sometimes too)

```suggestion
    DiscordEventHandlers handle{};
```

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,35 @@
...
+    Discord_Initialize("869824972464488510", &Handle, 1, NULL);
```
Xottab-DUTY: 1. Using NULL is outdated, it should be nullptr.
2. It's better to make a constant on top of the file (after includes) instead of hardcoded string with discord app ID.

src/xrGame/DiscordRPC.cpp
Xottab-DUTY: > Not sure, but shouldn't we clear the memory after this variable usage?

memset doesn't allocate memory.

Xottab-DUTY: Again, no need for memset.
```suggestion
    DiscordRichPresence discordPresence{};
```

src/xrGame/DiscordRPC.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <DiscordRPC/discord_register.h>
+#include <DiscordRPC/discord_rpc.h>
+#include <Windows.h>
+
```
Xottab-DUTY: Don't pay for what you don't use.
No need to include files which are not used. (include them in `.cpp`, except `windows.h`, it's already included in precompiled headers, globally)

And you should be especially careful with `windows.h` and not including it everywhere.
```suggestion
```

src/xrGame/DiscordRPC.h
```diff
@@ -0,0 +1,17 @@
...
+class DiscordRPC
+{
+public:
+    DiscordRPC() = default;
+    ~DiscordRPC();
+
+    void Init();
+    void Update(const char* level_name);
```
Xottab-DUTY: We use `pcstr` alias for `const char*`.
```suggestion
    void Update(pcstr level_name);
```

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,35 @@
...
+}
+
+DiscordRPC::~DiscordRPC()
+{
+    Discord_Shutdown();
+    Discord_ClearPresence();
```
Xottab-DUTY: Please, make the code symmetric: move deinitialization code to `Deinit` function.

src/xrGame/GametaskManager.cpp
```diff
@@ -238,7 +239,6 @@ CGameTask* CGameTaskManager::ActiveTask(ETaskType type)
         return nullptr;
 
     TASK_ID& t_id = g_active_task_id[type];
-
```
Xottab-DUTY: Return deleted string

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,53 @@
+#include "stdafx.h"
+#include "DiscordRPC.h"
+#include <DiscordRPC/discord_register.h>
+#include <DiscordRPC/discord_rpc.h>
+#include <xrCore/Text/StringConversion.hpp>
+
+DiscordRPC g_DiscordRPC;
+const pcstr g_discordAppID = "869824972464488510";
```
Xottab-DUTY, line 8:
```suggestion
constexpr pcstr DISCORD_APP_ID = "869824972464488510";
```

Xottab-DUTY:
```suggestion
constexpr pcstr DISCORD_APP_ID = "869824972464488510";

DiscordRPC g_DiscordRPC;
```

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+void DiscordRPC::Init()
+{
+    DiscordEventHandlers handle{};
+    Discord_Initialize(g_discordAppID, &handle, 1, nullptr);
```
Xottab-DUTY:
```suggestion
    Discord_Initialize(DISCORD_APP_ID, &handle, 1, nullptr);
```

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    xr_sprintf(build_name, "build %u", Core.GetBuildId());
```
Xottab-DUTY, line 15: Move it to the constructor

src/xrGame/DiscordRPC.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+class DiscordRPC
+{
+private:
+    int64_t     start_time          {};
+    string64    build_name          {};
+    string64    current_level_name  {};
+    pcstr       current_task_name   {};
```
Xottab-DUTY, line 9: I don't see any purpose for `current_level_name` to be a member variable.
```suggestion
    s64 start_time{};
    string64 build_name{};
    pcstr current_task_name{};
```

src/xrGame/DiscordRPC.h
```diff
@@ -0,0 +1,21 @@
...
+
+public:
+    DiscordRPC      (){ start_time = time(nullptr); };
+    ~DiscordRPC     (){};
```
Xottab-DUTY: Move constructor to .cpp
```suggestion
    DiscordRPC() { start_time = time(nullptr); }
    ~DiscordRPC();
```

src/xrGame/DiscordRPC.h
```diff
@@ -0,0 +1,21 @@
...
+
+    void Init       ();
+    void Update     (pcstr  level_name = nullptr);
+    void SetTask    (pcstr  task_name);
+    void Deinit     ();
```
Xottab-DUTY:
```suggestion
    void Init();
    void Deinit();

    void Update(pcstr level_name = nullptr);
    void SetTask(pcstr task_name);
```

src/xrGame/GamePersistent.cpp
```diff
@@ -938,6 +941,9 @@ void CGamePersistent::LoadTitle(bool change_tip, shared_str map_name)
 
         pApp->LoadTitleInt(
             StringTable().translate("ls_header").c_str(), tmp.c_str(), StringTable().translate(buff).c_str());
+
+        if (map_name.c_str())
```
Xottab-DUTY:
```suggestion
        if (!map_name.empty())
```

src/xrGame/DiscordRPC.h
Xottab-DUTY: Currently, we can make `current_level_name` to be local variable instead of member variable. :)
```suggestion
    int64_t start_time{};
    string64 build_name{};
    pcstr current_task_name{};
```

src/xrGame/DiscordRPC.h
doliinyk: So I have to move Init() functions to constructor, may I delete function? There's no need of it else

src/xrGame/GamePersistent.cpp
doliinyk: image
There's no empty() method for map_name

src/xrGame/DiscordRPC.h
Xottab-DUTY: > So I have to move Init() functions to constructor

No

Xottab-DUTY: Both constructor and Init stay the same, except there's just one line of code that should be moved from `Init` to constructor

src/xrGame/DiscordRPC.cpp
Xottab-DUTY: I mean, move ONLY THIS LINE

src/xrGame/GametaskManager.cpp
```diff
@@ -269,6 +270,7 @@ void CGameTaskManager::SetActiveTask(CGameTask* task)
         g_active_task_id[type] = task->m_ID;
         m_flags.set(eChanged, TRUE);
         task->m_read = true;
+        g_DiscordRPC.SetTask(StringTable().translate(task ? task->m_Title : "st_no_active_task").c_str());
```
collectioner: I think, I'd like to move `StringTable().translate()` call inside `SetTask` and `Update` methods. My point is that if anything will be changed in the logic of getting localization, then we'll fix it only inside `DiscordRPC` class.
@Xottab-DUTY what do you think?

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,53 @@
...
+    if (level_name)
+    {
+        std::locale myLocale("");
+        xr_string temp = StringToUTF8(level_name, myLocale);
+        xr_sprintf(current_level_name, "%s", temp);
+    }
+
+    discordPresence.state               = current_level_name;
+    discordPresence.largeImageText      = current_level_name;
+    discordPresence.details             = current_task_name;
```
collectioner: @Xottab-DUTY should we leave such `column` style for assignment operators?

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,53 @@
+#include "stdafx.h"
+#include "DiscordRPC.h"
+#include <DiscordRPC/discord_register.h>
+#include <DiscordRPC/discord_rpc.h>
+#include <xrCore/Text/StringConversion.hpp>
```
Xottab-DUTY: Using `<>` on own files is wrong, we should use quotes.
We also include third-party files after our own.
```suggestion
#include "DiscordRPC.h"
#include "xrCore/Text/StringConversion.hpp"

#include <DiscordRPC/discord_register.h>
#include <DiscordRPC/discord_rpc.h>
```

src/xrGame/GamePersistent.cpp
Xottab-DUTY: Whoops, use this then:
```suggestion
        if (map_name.size())
```

src/xrGame/DiscordRPC.h
collectioner: I agree. We can set all properties related to level inside `if` statement, when new level is passed.

collectioner (changes requested):

src/xrGame/GametaskManager.cpp
Xottab-DUTY: I totally disagree.

Xottab-DUTY: We should keep Discord integration abstract from localization system.

src/xrGame/DiscordRPC.h
Xottab-DUTY: And make destructor default. There's no point in making it empty.

doliinyk: Made `~DiscordRPC() = default;` 'cause compiler complains on `~DiscordRPC();`

src/xrGame/DiscordRPC.h
doliinyk: Did not make variable local, because then current_task_name overwrites location in status.

doliinyk:
Pay attention that status is still broking, when player changes location/loading save.

src/xrGame/DiscordRPC.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+class DiscordRPC
+{
+private:
```
Xottab-DUTY: Not needed, default access specifier is private already.
```suggestion
```

src/xrGame/GamePersistent.cpp
```diff
@@ -38,6 +38,7 @@
 #include "xrCore/Threading/TaskManager.hpp"
 
 #include "xrPhysics/IPHWorld.h"
+#include "DiscordRPC.h"
```
Xottab-DUTY: Please, remove `AnselManager.h` inclusion on line 37 and make it like that:
```suggestion
#include "xrPhysics/IPHWorld.h"

#include "AnselManager.h"
#include "DiscordRPC.h"
```

src/xrGame/DiscordRPC.cpp
Xottab-DUTY: Let's leave it as is, yes.

Xottab-DUTY:
> Pay attention that status is still broking, when player changes location/loading save.

Mm??

collectioner (approved):

src/xrGame/GamePersistent.cpp
```diff
@@ -34,11 +34,12 @@
 #include "xrEngine/xr_input.h"
 #include "xrEngine/x_ray.h"
 #include "ui/UILoadingScreen.h"
-#include "AnselManager.h"
-#include "xrCore/Threading/TaskManager.hpp"
```
Xottab-DUTY:
```suggestion
```

src/xrGame/DiscordRPC.h
```diff
@@ -0,0 +1,21 @@
...
+{
+    int64_t start_time{};
+    string64 build_name{};
+    string64 current_level_name{};
+    pcstr current_task_name{};
+
+public:
+    DiscordRPC();
+    ~DiscordRPC() = default;
+
```
Xottab-DUTY:
```suggestion
```

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,60 @@
...
+{
+    xr_sprintf(build_name, "build %u", Core.GetBuildId());
+    start_time = time(nullptr);
+}
+
+void DiscordRPC::Init()
+{
+    DiscordEventHandlers handle{};
+    Discord_Initialize(DISCORD_APP_ID, &handle, 1, nullptr);
+}
```
Xottab-DUTY:
```suggestion
void DiscordRPC::Init()
{
    xr_sprintf(build_name, "build %u", Core.GetBuildId());
    start_time = time(nullptr);
    DiscordEventHandlers handle{};
    Discord_Initialize(DISCORD_APP_ID, &handle, 1, nullptr);
}
```

Xottab-DUTY (commented):
I just realized that `Core.GetBuildId()` can return 0 (at the point of executing constructor code), so it's better to move the code to `Init` function.

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,50 @@
...
+    DiscordRichPresence discordPresence{};
+    static std::locale locale("");
+    xr_string temp = StringToUTF8(data, locale);
+
+    discordPresence.startTimestamp  = start_time;
+    discordPresence.smallImageText  = build_name;
+    discordPresence.largeImageKey   = "current_level_name";
+    discordPresence.smallImageKey   = "build_name";
+
+    //  type = TRUE  -> setting location
```
collectioner: Please, remove comments.

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,50 @@
...
+    //  type = FALSE -> setting task
+    if (type)
+        xr_strcpy(current_level_name, temp.c_str());
+    else
+        xr_strcpy(current_task_name, temp.c_str());
```
collectioner: Looks like it can be merged into conditional operator, like this:
`xr_strcpy(type ? current_level_name : current_task_name, temp.c_str());`

But I'd leave @Xottab-DUTY to decide If we need this adjustment.

src/xrGame/DiscordRPC.h
```diff
@@ -0,0 +1,17 @@
...
+    int64_t start_time{};
+    string64 build_name{};
+    string256 current_level_name{};
+    string256 current_task_name{};
+
+public:
+    void Init();
+    void Deinit();
+
+    void Update(bool type, pcstr data);
```
collectioner: I don't like the name of the first argument. It's not clear what type or how should we use it.
According to the usage of this argument, I'd suggest to rename it to smth like `isLevelUpdate` or similar.

cc: @Xottab-DUTY

src/xrGame/GametaskManager.cpp
```diff
@@ -85,7 +86,11 @@ CGameTask* CGameTaskManager::HasGameTask(const TASK_ID& id, bool only_inprocess)
     FindTaskByID key(id, only_inprocess);
     auto it = std::find_if(GetGameTasks().begin(), GetGameTasks().end(), key);
     if (it != GetGameTasks().end())
+    {
+        g_DiscordRPC.Update(
+            FALSE, StringTable().translate((*it).game_task ? (*it).game_task->m_Title : "st_no_active_task").c_str());
```
collectioner: Since you declared first argument as bool, why do you use defines instead of standard bool variables, like `false` and `true`?

src/xrGame/DiscordRPC.cpp
Xottab-DUTY: Why?

collectioner: Well, yep, I see why do we need them now. But we can rename type to something more clear and then remove comments :)

src/xrGame/DiscordRPC.h
collectioner: Ok, let's update this as we discussed offline. I'll mark other related comments as resolved for now.

src/xrGame/DiscordRPC.h
```diff
@@ -0,0 +1,23 @@
...
+    int64_t start_time{};
+    string64 build_name{};
+    string256 current_level_name{};
+    string256 current_task_name{};
+
+public:
+    enum DiscordStatusType
+    {
+        UPDATE_LEVEL,
+        UPDATE_TASK
```
Xottab-DUTY:
```suggestion
        UpdateLevel,
        UpdateTask,
```

src/xrGame/Level_start.cpp
```diff
@@ -154,6 +155,9 @@ bool CLevel::net_start2()
         }
         Server->SLS_Default();
         map_data.m_name = Server->level_name(m_caServerOptions);
+        if (map_data.m_name.size())
+            g_DiscordRPC.Update(
+                DiscordRPC::DiscordStatusType::UPDATE_LEVEL, StringTable().translate(map_data.m_name).c_str());
```
Xottab-DUTY:
```suggestion
            g_DiscordRPC.Update(DiscordRPC::DiscordStatusType::UPDATE_LEVEL, StringTable().translate(map_data.m_name).c_str());
```

src/xrGame/DiscordRPC.cpp
```diff
@@ -0,0 +1,47 @@
...
+void DiscordRPC::Update(DiscordStatusType updateType, pcstr updateData)
+{
+    DiscordRichPresence discordPresence{};
+    static std::locale locale("");
+    xr_string tempData = StringToUTF8(updateData, locale);
+
+    if (updateType == DiscordRPC::DiscordStatusType::UPDATE_LEVEL)
+        xr_strcpy(current_level_name, tempData.c_str());
+    else if (updateType == DiscordRPC::DiscordStatusType::UPDATE_TASK)
+        xr_strcpy(current_task_name, tempData.c_str());
```
Xottab-DUTY:
```suggestion
    if (updateType == DiscordStatusType::UpdateLevel)
        xr_strcpy(current_level_name, tempData.c_str());
    else if (updateType == DiscordStatusType::UpdateTask)
        xr_strcpy(current_task_name, tempData.c_str());
```

src/xrGame/DiscordRPC.cpp
Xottab-DUTY: Using ternary operator is no longer relevant, since now we have enumeration instead of bool :)

Xottab-DUTY (approved):

src/xrGame/GametaskManager.cpp
```diff
@@ -85,7 +86,11 @@ CGameTask* CGameTaskManager::HasGameTask(const TASK_ID& id, bool only_inprocess)
     FindTaskByID key(id, only_inprocess);
     auto it = std::find_if(GetGameTasks().begin(), GetGameTasks().end(), key);
     if (it != GetGameTasks().end())
+    {
+        g_DiscordRPC.Update(DiscordRPC::DiscordStatusType::UpdateTask,
+            StringTable().translate((*it).game_task ? (*it).game_task->m_Title : "st_no_active_task").c_str());
```
OPNA2608, line 86: Needs Windows-specific `#ifdef` blocks like in src/xrGame/GamePersistent.cpp:
```
CMakeFiles/xrGame.dir/GametaskManager.cpp.o:GametaskManager.cpp:function CGameTaskManager::HasGameTask(shared_str const&, bool): error: undefined reference to 'g_DiscordRPC'
CMakeFiles/xrGame.dir/GametaskManager.cpp.o:GametaskManager.cpp:function CGameTaskManager::HasGameTask(shared_str const&, bool): error: undefined reference to 'DiscordRPC::Update(DiscordRPC::DiscordStatusType, char const*)'
```

src/xrGame/Level_start.cpp
```diff
@@ -154,6 +155,8 @@ bool CLevel::net_start2()
         }
         Server->SLS_Default();
         map_data.m_name = Server->level_name(m_caServerOptions);
+        if (map_data.m_name.size())
+            g_DiscordRPC.Update(DiscordRPC::DiscordStatusType::UpdateLevel, StringTable().translate(map_data.m_name).c_str());
```
OPNA2608, line 158: Needs Windows-specific `#ifdef` blocks like in src/xrGame/GamePersistent.cpp:
```
CMakeFiles/xrGame.dir/Level_start.cpp.o:Level_start.cpp:function CLevel::net_start2(): error: undefined reference to 'g_DiscordRPC'
CMakeFiles/xrGame.dir/Level_start.cpp.o:Level_start.cpp:function CLevel::net_start2(): error: undefined reference to 'DiscordRPC::Update(DiscordRPC::DiscordStatusType, char const*)'
```

OPNA2608 (changes requested):
Breaks Linux builds in its current state.

jjdredd:
I don't think that it should be available for linux yet. It's not included in cmakelists afaik

src/xrGame/GametaskManager.cpp
Xottab-DUTY: No, no. No ifdefs, this can work on Linux, just needs to be fixed for it :)

src/xrGame/Level_start.cpp
Xottab-DUTY: No, no. No ifdefs, this can work on Linux, just needs to be fixed for it :)

russkel:
Should the binaries be directly added to the repo? Maybe it would be better as an `External`?

PR #868, opened 2021-08-03
Show cursor and give it back so it's possible to use mouse while debugging

Masterkatze:


Xottab-DUTY:
Noo, let's do it another way: implement this in window handler implementation (i.e. CRenderDevice)

Xottab-DUTY:
GiveBackCursor is not needed and, moreover, it is being called in a wrong place (should be called in xrDebug::Fail as well as possible TakeCursor function)

Please, use existing functionality of xrInput in `CRenderDevice::DisableFullscreen` and `CRenderDevice::EnableFullscreen`. 🙂

Xottab-DUTY:
Originally, I suggested to call `GrabInput` from `IWindowHandler` interface, but eventually I've found that it is not needed, there is a duplicated and not used functionality... So, I decided to do a little refactoring myself.

I hope you won't be mad at me and I hope that my refactoring fixes the issue :D

PR #894, opened 2021-10-09
Fix links to Travis CI

MerijnHendriks:
With this, the build status for Travis-CI will show up correctly again


Xottab-DUTY:
Thank you! I forgot to changed it from removed `xd_dev`...

Please, set it to travis-ci.**org**. We don't have travis-ci.**com** configured and it doesn't have any build history at all...

Xottab-DUTY:
Ok, I've found a way to migrate to travis-ci.com.

PR #913, opened 2021-11-23
"snd_precache_all" commad fix

VaIeroK:
Fixed sounds loading if some sound is broken


Xottab-DUTY:
We figured out that everything is ok with the current engine code.

PR #924, opened 2021-12-17
Added emplace method to AssociativeVector.

### LINKED ISSUE #256: Add `emplace` to AssociativeVector
and use it in `game_state_accumulator_state_register.cpp`

dino8890:
Fix for the issue #256.

### DIFF
--- a/src/xrCore/Containers/AssociativeVector.hpp
+++ b/src/xrCore/Containers/AssociativeVector.hpp
@@ -63,6 +63,7 @@ public:
     IC iterator insert(iterator where, const value_type& value);
     template <class TIterator>
     IC void insert(TIterator first, TIterator last);
+    IC insert_result emplace(key_type key, mapped_type value);
     IC void erase(iterator element);
     IC void erase(iterator first, iterator last);
     IC size_type erase(const key_type& key);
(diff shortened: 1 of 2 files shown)


src/xrCore/Containers/AssociativeVector.hpp
```diff
@@ -302,6 +303,11 @@ IC void _associative_vector::insert(TIterator first, TIterator last)
     std::sort(begin(), end(), static_cast<TComparer&>(*this));
 }
 
+TEMPLATE_SPECIALIZATION
+IC typename _associative_vector::insert_result _associative_vector::emplace(key_type key, mapped_type value) {
+    return insert(std::make_pair(key, value));
```
Xottab-DUTY: It should not call own `insert` function. It should call `emplace` from the underlying `xr_vector`.

dino8890: Hey, thanks for the feedback!

Figured that might be the case, although it didn't seem prudent at first considering the vector elements can't ever be anything else than an `std::pair`. I will get back on it, and fix the issue.

Xottab-DUTY:
Thank you for all the changes! Though, they are needed, they are not related to this pull request.

Please, move other changes into new pull request and leave in this pull request only changes that are related to AssociativeVector.

dino8890:
Will do, cheers!

dino8890:
This should be it, let me know if you need additional changes.

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you!

Sorry for the long review, I was sick and then was a bit busy with other stuff.. image

PR #937, opened 2022-01-17
Fix LUA_STUDIO debugger ifdefs

russkel:
I have been trying to compile a Debug build of xray-16 on my computer and I was unable to because of a few things in this commit.

I think maybe it was because of a dirty cmake state, but I don't see how `ScheduledBase` could ever compile with `MASTER_GOLD` being set, because `Registered` is never defined.

Is a `Debug` and `MASTER_GOLD` combination an expected combination?

### DIFF
--- a/src/xrScriptEngine/ScriptEngineScript.cpp
+++ b/src/xrScriptEngine/ScriptEngineScript.cpp
@@ -32,7 +32,7 @@ void ErrorLog(pcstr caMessage)
     if (GEnv.ScriptEngine->debugger())
         GEnv.ScriptEngine->debugger()->Write(caMessage);
 #endif
-#ifdef DEBUG
+#if defined(USE_DEBUGGER) && defined(USE_LUA_STUDIO)
     bool lua_studio_connected = !!GEnv.ScriptEngine->debugger();
     if (!lua_studio_connected)
         R_ASSERT2(0, caMessage);
--- a/src/xrScriptEngine/script_engine.cpp
+++ b/src/xrScriptEngine/script_engine.cpp
@@ -549,7 +549,7 @@ struct raii_guard : private Noncopyable
 
     ~raii_guard()
     {
-#ifdef DEBUG
+#if defined(USE_DEBUGGER) && defined(USE_LUA_STUDIO)
         const bool lua_studio_connected = !!m_script_engine->debugger();
         if (!lua_studio_connected)
 #endif


src/xrScriptEngine/ScriptEngineScript.cpp
```diff
@@ -32,7 +32,7 @@ void ErrorLog(pcstr caMessage)
     if (GEnv.ScriptEngine->debugger())
         GEnv.ScriptEngine->debugger()->Write(caMessage);
 #endif
-#ifdef DEBUG
+#ifdef USE_DEBUGGER
```
Xottab-DUTY:
```suggestion
#ifdef USE_LUA_STUDIO
```

src/xrScriptEngine/script_engine.cpp
```diff
@@ -549,7 +549,7 @@ struct raii_guard : private Noncopyable
 
     ~raii_guard()
     {
-#ifdef DEBUG
+#ifdef USE_DEBUGGER
```
Xottab-DUTY:
```suggestion
#ifdef USE_LUA_STUDIO
```

Xottab-DUTY:
Master Gold is not intended to be enabled in Debug, it's only for Release.

Xottab-DUTY (commented):
> Master Gold is not intended to be enabled in Debug, it's only for Release.

But! Changes in xrScriptEngine are useful! Though, `USE_LUA_STUDIO` should be used there.
Can you submit new PR or reopen this with only changes to xrScriptEngine? :)

src/xrScriptEngine/ScriptEngineScript.cpp
russkel: > > Master Gold is not intended to be enabled in Debug, it's only for Release.
> 
> But! Changes in xrScriptEngine are useful! Though, `USE_LUA_STUDIO` should be used there. Can you submit new PR or reopen this with only changes to xrScriptEngine? :)

Can do. Re-examining the file it looks like `#if defined(USE_DEBUGGER) && defined(USE_LUA_STUDIO)` is used. I will keep in line with that.

Can you reopen the PR and I will push fixed commits?

Xottab-DUTY: @russkel, done!

russkel: Thanks @Xottab-DUTY. Should be good to go now.

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you!

Xottab-DUTY:
Thank you!

PR #941, opened 2022-01-23
Added bash script for building by non-prepared users.

007exe:


russkel:
This seems more complicated than the actual commands required to build and install. If users want a simple way to build it then they should use their distro's packaging system, no?

```
cmake .
make
sudo make install
```
Should all be that's required, and if it isn't then the cmakefiles should be fixed accordingly.

I like the unpacking/copying/symlinking game data commands, though.

make_portable.sh
```diff
@@ -0,0 +1,156 @@
+#! /bin/bash
+
+out=~/OpenXRay
+def_copy_path=~
+
+#=================================== Функция справки
```
q4a: Please, translate russian comments in english

make_portable.sh
```diff
@@ -0,0 +1,156 @@
...
+Attention!!! On some distributions, packages may be prefixed with -dev e.g. libglvnd-dev libjpeg6-turbo-dev " 14 130)  then
+
+
+#   rm -f -R bin
+   mkdir -p bin
+   cd bin
+   cmake .. -DCMAKE_BUILD_TYPE=Release \
+   -DCMAKE_INSTALL_PREFIX=/usr \
+   -DCMAKE_INSTALL_LIBDIR=lib
+   make -j5
```
q4a: `make -j5` - looks strange for me.
I would suggest to use:
`make -j$(nproc)` or `make -j$(nproc || echo 4)`
(`nproc` - is part of coreutils like `ls` or `mkdir`, so I thinks that not need to check if it exists)

q4a (changes requested):
Thanks. Looks good for me except 2 things (check comment in code).

make_portable.sh
007exe: Thanks, I'll fix that.

make_portable.sh
007exe: Over time, I will revisit or delete them altogether, it’s easier for me to navigate the code

007exe:
I haven’t implemented everything that I wanted and what they ask, at least you need to add functionality for unpacking the game distribution using innoextract and Sultan, just add auto installation of dependencies for the build, there are ideas, but there is no inspiration and time yet. It is also desirable to think about how to disable symlinks in ~/.local/share/GSK ....

PR #946, opened 2022-01-28
Fix CMake generator expressions comparing options

garyttierney:
Hi, just a simple build system bugfix for an issue I ran into while seeing if I could easily
run the CMake build with Clang on Windows.

There are a few instances of CMake generator expressions that attempt to
check a configuration time option and convert it to a boolean, but use
the literal string name of the option instead of expanding it to the
value. The result is that every expression evaluates to `true`, per: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#logical-operators

I've opened a separate PR against the luabind fork for the same thing: https://github.com/OpenXRay/luabind-deboostified/pull/21


garyttierney:
Oops, seems I missed a few across the NVTT buildsystem:

- https://github.com/OpenXRay/xray-16/blob/b8ce8d735cd4b581f7300c90ead6b1d06bd95d49/Externals/NVTT/src/nvtt/CMakeLists.txt#L74
- https://github.com/OpenXRay/xray-16/blob/b8ce8d735cd4b581f7300c90ead6b1d06bd95d49/Externals/NVTT/src/nvtt/squish/CMakeLists.txt#L33
- https://github.com/OpenXRay/xray-16/blob/b8ce8d735cd4b581f7300c90ead6b1d06bd95d49/Externals/NVTT/src/nvmath/CMakeLists.txt#L48

I'll fix these and rebuild on my Linux system then update the PR.

Xottab-DUTY:
> Hi, just a simple build system bugfix for an issue I ran into while seeing if I could easily
> run the CMake build with Clang on Windows.

Hi, thank you!
Currently, Windows builds are properly configured only for MSBuild (Visual Studio) and building on Windows using CMake is just unsupported.
But we are open for efforts to configure CMake for Windows! :) Maybe even drop MSBuild and move to CMake for all platforms. (I have some concerns about compilation speed with CMake, though. Our Linux builds compilation speed is slower than Windows builds🙁)

garyttierney:
Hey,

Yeah, I appreciate CMake is unsupported on Windows and don't expect any dev support for it. I'm happy to keep the parts for the things I break :). I think I'll keep tinkering to see if I can get it working but if I make significant progress I'll open an issue for discussion first before I try sending any large changes upstream.

> Our Linux builds compilation speed is slower than Windows builds🙁)

Ah, that's actually the opposite of why I had a look at getting CMake working on Windows. For me MSBuild only uses 1 of my cores and takes upwards of about 10 minutes for a fresh build, but maybe that's down to my inexperience with Microsoft's C++ toolchain and there's a parallelism option in there somewhere I can turn on. On my Linux machine I get much faster builds when running across all my cores.

Masterkatze:
Hey, thanks for this pull-request! It was overlooked for sure

Masterkatze (approved):

PR #950, opened 2022-02-06
rgl: fix for stretched noise in radia hazard zones

vTurbine:


Xottab-DUTY (approved):

src/Layers/xrRender_R2/r2_rendertarget_phase_PP.cpp
```diff
@@ -2,7 +2,18 @@
 
 void CRenderTarget::u_calc_tc_noise(Fvector2& p0, Fvector2& p1)
 {
-    CTexture* T = RCache.get_ActiveTexture(2);
+#if defined(USE_OGL)
+    // NOTE: OGL uniforms need to be separated by types. Otherwise, accessing them
```
Xottab-DUTY: We need to mark it more explicitly as `XXX` 
image

PR #957, opened 2022-02-19
xrRender_R4: fix for unbound sampler in UI draw

vTurbine:
D3D11 WARNING: ID3D11DeviceContext::Draw: The Pixel Shader unit expects a Sampler to be set at Slot 0, but none is bound. This is perfectly valid, as a NULL Sampler maps to default Sampler state. However, the developer may not want to rely on the defaults.  [ EXECUTION WARNING #352: DEVICE_DRAW_SAMPLER_NOT_SET]


Xottab-DUTY:
Marked this as bug because, occasionally, this isn't a vanilla flaw, but it was introduced by us in 030fdd1b17aa0ccb96873436ce0edfba0130c2f8

PR #997, opened 2022-04-11
Add configs for debugging and tasks for VSCode

ChugunovRoman:


.vscode/settings.json
```diff
@@ -0,0 +1,56 @@
...
+  "files.associations": {
+    "*.tpl": "html",
+    "*.svetle": "html",
+    "*.mustache": "html",
+    "*.js": "javascript",
+    "*.ts": "typescript",
+    "*.toml": "properties",
+    "*.service": "properties",
+    "*.prettierrc": "json",
+    "*.qc": "c",
```
Xottab-DUTY: We don't have any TypeScript, JavaScript or SQL and etc...
Is it needed here for some reason?

ChugunovRoman: Yes, this is unnecessary. I removed it

PR #1016, opened 2022-04-30
Add the ResetBoneProtections method

ChugunovRoman:


src/xrGame/script_game_object2.cpp
```diff
@@ -561,3 +561,12 @@ bool CScriptGameObject::is_there_items_to_pickup() const
     }
     return (!!stalker->memory().item().selected());
 }
+
+void CScriptGameObject::ResetBoneProtections(pcstr imm_sect, pcstr bone_sect)
+{
+    CAI_Stalker* stalker = smart_cast<CAI_Stalker*>(&object());
+    if (!stalker)
+        return;
```
Xottab-DUTY, line 572: Maybe we need to add error log here?
Or this function is used also on, e.g., monsters?

ChugunovRoman: Added the log

Xottab-DUTY (approved):

PR #1027, opened 2022-05-10
Disable mirroring attempts on forks

q4a:


Xottab-DUTY:
@q4a, can you also disable mirroring attempts on pull requests?

PR #1032, opened 2022-05-13
Extend level script

ChugunovRoman:


src/xrGame/level_script.cpp
```diff
@@ -573,6 +578,19 @@ void stop_tutorial()
         g_tutorial->Stop();
 }
 
+LPCSTR tutorial_name()
```
Xottab-DUTY: Never use `LPCSTR` in the new code, use `pcstr`.

src/xrGame/ui/UIGameTutorial.h
```diff
@@ -24,6 +24,7 @@ class CUISequencer : public pureFrame, public pureRender, public IInputReceiver
     shared_str m_stop_lua_function;
 
 public:
+    LPCSTR m_name;
```
Xottab-DUTY:
```suggestion
    pcstr m_name;
```

Xottab-DUTY: Please, make it NOT public

src/xrGame/level_script.cpp
```diff
@@ -685,6 +708,8 @@ IC static void CLevel_Export(lua_State* luaState)
         def("spawn_item", &spawn_section),
         def("get_active_cam", &get_active_cam),
         def("set_active_cam", &set_active_cam),
+        def("get_start_time", &get_start_time),
```
Xottab-DUTY: Let's use lambdas for all new exported functions, like this:
```suggestion
        def("get_start_time", +[]() { return xrTime(Level().GetStartGameTime()); }),
```

Xottab-DUTY (changes requested):

src/xrGame/ui/UIGameTutorial.h
ChugunovRoman: If it will made private then we can not got this in script.

screenshot_2022-05-14T12:11:49

src/xrGame/level_script.cpp
ChugunovRoman: Fixed

src/xrGame/level_script.cpp
ChugunovRoman: Fixed

src/xrGame/ui/UIGameTutorial.h
ChugunovRoman: I made this field as private, and replaced it on a method-getter for this filed

src/xrGame/level_script.cpp
```diff
@@ -877,7 +882,16 @@ IC static void CLevel_Export(lua_State* luaState)
         def("start_tutorial", &start_tutorial),
         def("stop_tutorial", &stop_tutorial),
         def("has_active_tutorial", &has_active_tutotial),
-        def("translate_string", &translate_string)
+	    def("active_tutorial_name", +[]()
+        {
+            if (g_tutorial)
+                return g_tutorial->GetTutorName();
+
+            return "invalid";
```
Xottab-DUTY: Maybe it's better to use `nullptr` and check for `nil` in Lua?

ChugunovRoman: Exactly. Updated

ChugunovRoman: The build fails if return `nullptr`... I removed `return`, but now this fail the build only on Windows. How to return `nullptr` here?

src/xrGame/ui/UIGameTutorial.cpp
```diff
@@ -97,6 +97,7 @@ CUISequencer::CUISequencer()
 {
     m_UIWindow = nullptr;
     m_pStoredInputReceiver = nullptr;
+    m_name = "invalid";
```
Xottab-DUTY: It's better to keep it `nullptr`. What do you think?

src/xrGame/ui/UIGameTutorial.h
```diff
@@ -67,6 +68,9 @@ class CUISequencer : public pureFrame, public pureRender, public IInputReceiver
         etsOverMainMenu = (1 << 6),
     };
     Flags32 m_flags;
+
+private:
+    pcstr m_name;
```
Xottab-DUTY: It's better to place it after `m_sequencer_items`

src/xrGame/ui/UIGameTutorial.cpp
```diff
@@ -118,6 +119,7 @@ bool CUISequencer::Start(LPCSTR tutor_name)
 
     Device.seqFrame.Add(this, REG_PRIORITY_LOW - 10000);
 
+    m_name = tutor_name;
```
Xottab-DUTY, line 122: And we also should set it to the default name when tutorial ends\stops.

src/xrGame/ui/UIGameTutorial.cpp
ChugunovRoman: I just fogot change this on nullptr ;)
Updated

src/xrGame/ui/UIGameTutorial.cpp
ChugunovRoman: Updated

src/xrGame/ui/UIGameTutorial.h
ChugunovRoman: Updated

src/xrGame/ui/UIGameTutorial.cpp
ChugunovRoman: Did you mean set to nullptr when tutorial ends\stops?

Xottab-DUTY: @ChugunovRoman, yes

ChugunovRoman: Fixed, but I'm not sure...

src/xrGame/ui/UIGameTutorial.cpp
```diff
@@ -193,6 +195,9 @@ bool CUISequencer::Start(LPCSTR tutor_name)
 
     if (m_start_lua_function.size())
         CallFunction(m_start_lua_function);
+
+    m_name = nullptr;
+
```
Xottab-DUTY, line 200:
```suggestion
```

Xottab-DUTY (approved):

PR #1031, opened 2022-05-13
Feature: Allow to create custom PDA tabs

ChugunovRoman:

PR #1037, opened 2022-05-13
Fixed crash if parent dialog holder is not set

ChugunovRoman:


Xottab-DUTY:
Parent dialog holder always should be set! I don't think this fix should be here.

ChugunovRoman:
> Parent dialog holder always should be set! I don't think this fix should be here.

Sometimes I got segmentation fault on this line, but I have not yet figured out the true reason.

ChugunovRoman:
@Xottab-DUTY When I'll reproduce the bug I'll let you know

ChugunovRoman:
Related #1031

PR #1049, opened 2022-06-11
Reject item in active slot on death in TDM

unnamedguy:
There is a mistake in commit 7c99022f47d32a1b6216466815551f1cfe88f700 in game_sv_teamdeathmatch.cpp file. When you die in TDM, your item in the active slot is destroyed, but it shouldn't be, according to Call of Pripyat multiplayer.


Xottab-DUTY:
Thank you!

PR #1052, opened 2022-06-14
[xrRender] Fixed game crash when rendering wallmarks 

eagleivg:
from an underbarrel grenade launcher (out-of-bounds vertex array)

Detected with adress saitizer. If you iterate over vertices, you go to the next vertex without checking its existence (just using a increment of a vertex pointer). This worked fine for bullet hits, but caused a crash when using a grenade launcher (maybe more bones are affected for rendering?). In any case, the blood spatters in bullet hits are drawn, and in case of grenade hits there is no point in even drawing them - they are completely overlapped by the effects of the explosion.


eagleivg:
As far as I could figure it out, the streaming that is supposed to render at least 1024 * 3 vertices at a time doesn't work correctly. If the number of vertices exceeds this constant, it is likely to crash. Increased to 16k - to overlap the number of vertices in the burer's wallmark, probably the highest polygonal monster in the game.

PR #1091, opened 2022-09-21
Fixed: using %c[color] tag with multibyte font causing unexpected line ending

DPurple:
here's something useful :
https://github.com/themrdemonized/STALKER-Anomaly-modded-exes/issues/6
hope helps!

Before the patch:
image

After the patch:
image

Compare with English:
image


DPurple:
there's still something wrong is this code, I'll check them now

DPurple:
I found it was a bug in language texts, not here...

Xottab-DUTY:
Thank you! Sorry for a long review, I wanted to research how the system works and why the bug is appeared, but decided to just merge now and research later.

PR #1096, opened 2022-10-04
Get all active object by CTelekinesis::get_objects_total_count instea…

### LINKED ISSUE #886: Incorrect work of burer's telekinesis
CStateBurerAttackTele class uses CTelekinesis::get_objects_count method  to iterate over all active objects. It's incorrect - this method does not return the total number of objects, but the number of objects in the TS_Keep and TS_Raise states. To get the total number of active objects we need to use CTelekinesis::get_objects_total_count.
So, we need to review every call of get_objects_count  in CStateBurerAttackTele, replace the incorrect calls with get_objects_total_count and check the logic of the surrounding code.

Kiloyd:
…d of get_objects_count
Closes #886


Plotja:
It looks like you changed ALL the calls of get_objects_count to get_objects_total_count. Are you sure that we need to change all of them?

Kiloyd:
Is there any spec or doc that describe the AI action? Take `CStateBurerAttackTele<Object>::FireAllToEnemy()`  for example, it originally use `get_objects_count()` to get object in `TS_Keep` or `TS_Raise` state only, but I think it should loop all the objects no matter the state. The other function I'm not quite sure if it's ok to use `get_objects_total_count` directly.

Plotja:
Good example of function. I think that burer can fire only those objects that he has already raised. So we need to iterate only through raised objects here.

Kiloyd:
Nice and thanks for comment. I'll fix the `FireAllToEnemy()` in the next commit. If there's more info that I can get and check the spec, please let me know. Thanks again!

Kiloyd:
I noticed that poltergeist also use telekinesis way to attack. `CPolterTele::tele_fire_objects` has the condition to check the object is in `TS_Raise` or `TS_Keep` state then fire the object. So previous commit fix the burer telekinesis `FireAllToEnemy` function to only count object in above 2 states. I might cross-verification the way that burer and poltergeist use telekinesis to fix this issue.

src/xrGame/ai/monsters/poltergeist/poltergeist_telekinesis.cpp
```diff
@@ -84,6 +84,7 @@ void CPolterTele::update_schedule()
         m_time = time();
         m_time_next = 0;
         m_state = eFireObjects;
+        break;
```
FreeZoneMods: Are you sure you want to wait for the next update iteration before executing fire?

KevinNobollel: Thanks for the description. I seldom change state and execute it immediately so I add `break` in here. I agree with you that once the object is raised then fire it, will recover it in next commit. Thanks again!

Kiloyd: Thanks for the description. I seldom change state and execute it immediately so I add `break` in here. I agree with you that once the object is raised then fire it, will recover it in next commit. Thanks again!

Xottab-DUTY:
```suggestion
```

PR #1100, opened 2022-10-09
Fix incorrect focus on input fields. (#844)

### LINKED ISSUE #844: Input field is only in focus (and you only type into it) when mouse in hovered on it
In vanilla, input field didn't lose focus when you move mouse out.

Bug appeared after reimplementing input using SDL text input.

Plotja:
Resolving of #844 issue

- Fix Incorrect focus capture when hovering the cursor over the edit box and uncapture when the cursor moved out.
- Add capturing of focus by appropriate actions. 
- Correct return of focus to input field after hiding console.
- Fix "ё" error


src/xrUICore/EditBox/UICustomEdit.cpp
```diff
@@ -313,3 +314,4 @@ void CUICustomEdit::CaptureFocus(bool bCapture)
 
     m_bInputFocus = bCapture;
 }
+
```
Xottab-DUTY:
```suggestion
```

src/xrUICore/UIMessages.h
```diff
@@ -134,5 +135,5 @@ enum EUIMessages
     MAP_CHANGE_SPOT_HINT_ACT,
     MAP_REMOVE_SPOT_ACT,
 
-    MAIN_MENU_RELOADED,
+    MAIN_MENU_RELOADED
```
Xottab-DUTY:
```suggestion
    MAIN_MENU_RELOADED,
```

src/xrUICore/EditBox/UICustomEdit.cpp
```diff
@@ -76,8 +76,8 @@ void CUICustomEdit::Init(u32 max_char_count, bool number_only_mode, bool read_mo
 
     Register_callbacks();
     ClearText();
-
-    m_bInputFocus = false;
+
```
Xottab-DUTY:
```suggestion

```

Xottab-DUTY:
Thanks!

I found another solution, that works in all cases I know! 🙂

PR #923, opened 2021-12-17
Macos feature support

vertver:
Being tested on Mac M1

PR #1105, opened 2022-10-13
A set of small fixes to be able to run the game on macOS (fixes #1104)

### LINKED ISSUE #1104: Run the game on macOS
Although the engine can be built successfully on macOS, the game won't run.

Lnd-stoL:
These fixes are essential for running the game on macOS.
With this set of fixes I was able to successfully run the game on M1 Pro Mac without severe glitches (though there are some rendering related error messages in logs).

### DIFF
--- a/src/Layers/xrRender/r__dsgraph_render.cpp
+++ b/src/Layers/xrRender/r__dsgraph_render.cpp
@@ -26,8 +26,9 @@ bool cmp_ssa(const T &lhs, const T &rhs)
 template <typename T>
 bool cmp_pass(const T& left, const T& right)
 {
-    return left->first->equal(*right->first)
-        || left->second.ssa > right->second.ssa;
+    if (left->first->equal(*right->first))
+        return false;
+    return left->second.ssa >= right->second.ssa;
 }
 
 void R_dsgraph_structure::r_dsgraph_render_graph(u32 _priority)
@@ -46,7 +47,7 @@ void R_dsgraph_structure::r_dsgraph_render_graph(u32 _priority)
         auto& map = mapNormalPasses[_priority][iPass];
 
         map.get_any_p(nrmPasses);
-        std::sort(nrmPasses.begin(), nrmPasses.end(), cmp_pass<mapNormal_T::value_type*>);
+        std::stable_sort(nrmPasses.begin(), nrmPasses.end(), cmp_pass<mapNormal_T::value_type*>);
         for (const auto& it : nrmPasses)
         {
             RCache.set_Pass(it->first);
@@ -83,7 +84,7 @@ void R_dsgraph_structure::r_dsgraph_render_graph(u32 _priority)
         auto& map = mapMatrixPasses[_priority][iPass];
 
         map.get_any_p(matPasses);
-        std::sort(matPasses.begin(), matPasses.end(), cmp_pass<mapMatrix_T::value_type*>);
+        std::stable_sort(matPasses.begin(), matPasses.end(), cmp_pass<mapMatrix_T::value_type*>);
         for (const auto& it : matPasses)
         {
             RCache.set_Pass(it->first);
--- a/src/Layers/xrRenderGL/glSH_RT.cpp
+++ b/src/Layers/xrRenderGL/glSH_RT.cpp
@@ -52,15 +52,20 @@ void CRT::create(LPCSTR Name, u32 w, u32 h, D3DFORMAT f, u32 SampleCount /*= 1*/
     fmt = f;
     sampleCount = SampleCount;
 
-    // Get caps
-    GLint max_width, max_height;
-    CHK_GL(glGetIntegerv(GL_MAX_FRAMEBUFFER_WIDTH, &max_width));
-    CHK_GL(glGetIntegerv(GL_MAX_FRAMEBUFFER_HEIGHT, &max_height));
+    if (GLEW_VERSION_4_3) 
+    {
+        // Get caps
+        GLint max_width, max_height;
+        CHK_GL(glGetIntegerv(GL_MAX_FRAMEBUFFER_WIDTH, &max_width));
+        CHK_GL(glGetIntegerv(GL_MAX_FRAMEBUFFER_HEIGHT, &max_height));
 
-    // Check width-and-height of render target surface
-    // XXX: While seemingly silly, assert w/h are positive?
-    if (w > u32(max_width)) return;
-    if (h > u32(max_height)) return;
+        // Check width-and-height of render target surface
+        // XXX: While seemingly silly, assert w/h are positive?
+        if (w > u32(max_width))
+            return;
+        if (h > u32(max_height))
+            return;
+    }
 
     RImplementation.Resources->Evict();
 
--- a/src/Layers/xrRenderPC_GL/rgl_shaders.cpp
+++ b/src/Layers/xrRenderPC_GL/rgl_shaders.cpp
@@ -411,7 +411,12 @@ HRESULT CRender::shader_compile(pcstr name, IReader* fs, pcstr pFunctionName,
     appendShaderOption(o.gbuffer_opt, "GBUFFER_OPTIMIZATION", "1");
 
     // Shader Model 4.1
+#ifndef __APPLE__
     appendShaderOption(o.dx11_sm4_1, "SM_4_1", "1");
+    // Despite the fact that glsl 4.1 is claimed to be supported on macOS,
+    // the issue is that gatherTextureOffset requires compile-time constant offset argument.
+    // So it is more handy to disable its use for mac as for now.
+#endif
 
     // Minmax SM
     appendShaderOption(o.minmax_sm, "USE_MINMAX_SM", "1");
--- a/src/xrCore/xrMemory.cpp
+++ b/src/xrCore/xrMemory.cpp
@@ -112,7 +112,7 @@ size_t xrMemory::mem_usage()
         CloseHandle(h);
     }
     return pmc.PagefileUsage;
-#elif defined(XR_PLATFORM_LINUX)
+#elif defined(XR_PLATFORM_LINUX) || defined(XR_PLATFORM_APPLE)
     struct rusage ru;
     getrusage(RUSAGE_SELF, &ru);
     return (size_t)ru.ru_maxrss;
--- a/src/xrGame/reward_event_handler.h
+++ b/src/xrGame/reward_event_handler.h
@@ -11,7 +11,7 @@ class reward_event_handler
 {
 public:
     reward_event_handler(game_state_accumulator* pstate) : m_player_state_accum(pstate) {}
-    ~reward_event_handler(){};
+    virtual ~reward_event_handler(){};
     virtual bool OnWeapon_Fire(u16 sender, u16 sender_weapon_id) = 0;
     virtual bool OnBullet_Fire(u16 sender, u16 sender_weapon_id, const Fvector& position, const Fvector& direction) = 0;
     virtual bool OnBullet_Hit(
--- a/src/xrGame/ui/UIGameTutorial.h
+++ b/src/xrGame/ui/UIGameTutorial.h
@@ -26,6 +26,7 @@ protected:
 
 public:
     IInputReceiver* m_pStoredInputReceiver;
+    virtual ~CUISequencer() = default;
     CUISequencer();
     bool Start(LPCSTR tutor_name);
     void Stop();
@@ -129,6 +130,8 @@ class CUISequenceSimpleItem : public CUISequenceItem
 
         virtual void Start();
         virtual void Stop();
+
+        virtual ~SSubItem() = default;
     };
     using SubItemVec = xr_vector<SSubItem>;
     SubItemVec m_subitems;
--- a/src/xrGame/ui/UIGameTutorial.h
+++ b/src/xrGame/ui/UIGameTutorial.h
@@ -26,8 +26,10 @@ protected:
 
 public:
     IInputReceiver* m_pStoredInputReceiver;
-    virtual ~CUISequencer() = default;
+
     CUISequencer();
+    virtual ~CUISequencer() = default;
+
     bool Start(LPCSTR tutor_name);
     void Stop();
     void Next();
--- a/src/xrServerEntities/object_item_abstract.h
+++ b/src/xrServerEntities/object_item_abstract.h
@@ -21,8 +21,9 @@ protected:
     shared_str m_script_clsid;
 
 public:
-    virtual ~CObjectItemAbstract() = default;
     IC CObjectItemAbstract(const CLASS_ID& clsid, LPCSTR script_clsid);
+    virtual ~CObjectItemAbstract() = default;
+
     IC const CLASS_ID& clsid() const;
     IC shared_str script_clsid() const;
 #ifndef NO_XR_GAME
--- a/src/xrServerEntities/object_item_abstract.h
+++ b/src/xrServerEntities/object_item_abstract.h
@@ -21,6 +21,7 @@ protected:
     shared_str m_script_clsid;
 
 public:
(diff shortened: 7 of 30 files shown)


Xottab-DUTY:
Wow maan, many thanks! Especially for the first two commits! These `&*` dereferences and nitpicky problems like missing virtual destructors were pain in the ***!

src/xrGame/ui/UIGameTutorial.h
```diff
@@ -26,6 +26,7 @@ class CUISequencer : public pureFrame, public pureRender, public IInputReceiver
 
 public:
     IInputReceiver* m_pStoredInputReceiver;
+    virtual ~CUISequencer() = default;
     CUISequencer();
```
Xottab-DUTY: For the consistency, it's better to place the destructor after the constructor.
```suggestion

    CUISequencer();
    virtual ~CUISequencer() = default;

```

src/xrServerEntities/object_item_abstract.h
```diff
@@ -21,6 +21,7 @@ class CObjectItemAbstract
     shared_str m_script_clsid;
 
 public:
+    virtual ~CObjectItemAbstract() = default;
     IC CObjectItemAbstract(const CLASS_ID& clsid, LPCSTR script_clsid);
```
Xottab-DUTY: For the consistency, it's better to place the destructor after the constructor.
```suggestion
    IC CObjectItemAbstract(const CLASS_ID& clsid, LPCSTR script_clsid);
    virtual ~CObjectItemAbstract() = default;

```

src/Layers/xrRenderGL/glSH_RT.cpp
```diff
@@ -52,15 +52,18 @@ void CRT::create(LPCSTR Name, u32 w, u32 h, D3DFORMAT f, u32 SampleCount /*= 1*/
...
+    {
+        // Get caps
+        GLint max_width, max_height;
+        CHK_GL(glGetIntegerv(GL_MAX_FRAMEBUFFER_WIDTH, &max_width));
+        CHK_GL(glGetIntegerv(GL_MAX_FRAMEBUFFER_HEIGHT, &max_height));
+
+        // Check width-and-height of render target surface
+        // XXX: While seemingly silly, assert w/h are positive?
+        if (w > u32(max_width)) return;
+        if (h > u32(max_height)) return;
```
Xottab-DUTY: Let's fix the code style by the way :)
```suggestion
        if (w > u32(max_width))
            return;
        if (h > u32(max_height))
            return;
```

src/Layers/xrRenderPC_GL/rgl_shaders.cpp
```diff
@@ -411,7 +411,12 @@ HRESULT CRender::shader_compile(pcstr name, IReader* fs, pcstr pFunctionName,
     appendShaderOption(o.gbuffer_opt, "GBUFFER_OPTIMIZATION", "1");
 
     // Shader Model 4.1
+#ifndef __APPLE__
     appendShaderOption(o.dx11_sm4_1, "SM_4_1", "1");
+    // Despite the fact that glsl 4.1 is claimed to be supported on macOS,
+    // the issue is that gatherTextureOffset requires compile-time constant offset argument.
+    // So it is more handy to disable its use for mac as for now.
```
Xottab-DUTY, line 418: Maybe, we can provide compile-time offset later? (after some investigation, if this is possible)
Maybe we can leave a TODO here?
```suggestion
    // So it is more handy to disable its use for mac as for now.
    // XXX: investigate if we can provide compile-time constant offset
```

Xottab-DUTY (changes requested):
Thanks again!
There are few small remarks and compilation fails on Windows x64, though.

Xottab-DUTY:
@Lnd-stoL, you also maybe want to look at #923. Maybe, you'll find something useful there.

src/Layers/xrRenderPC_GL/rgl_shaders.cpp
Lnd-stoL: Unfortunately, its not so easy to provide compile-time constant, cause the typical use case of textureGatherOffset is in for loops (using loop counter as offset parameter). 
The solution could be loop unrolling, but it is not supported by OpenGL4.1.

So, the only option I see is "manual" loop unrolling producing code like

```
  float4 s_dmap_gpp[16];
  s_dmap_gpp[0] = textureGatherOffset( s_dmap, tc.xy, int2( -3, -3 ) );
  s_dmap_gpp[1] = textureGatherOffset( s_dmap, tc.xy, int2( -3, -1 ) );
  s_dmap_gpp[2] = textureGatherOffset( s_dmap, tc.xy, int2( -3,  1 ) );
  s_dmap_gpp[3] = textureGatherOffset( s_dmap, tc.xy, int2( -3,  3 ) );

  s_dmap_gpp[4] = textureGatherOffset( s_dmap, tc.xy, int2( -1, -3 ) );
  s_dmap_gpp[5] = textureGatherOffset( s_dmap, tc.xy, int2( -1, -1 ) );
  s_dmap_gpp[6] = textureGatherOffset( s_dmap, tc.xy, int2( -1,  1 ) );
  s_dmap_gpp[7] = textureGatherOffset( s_dmap, tc.xy, int2( -1,  3 ) );

  s_dmap_gpp[8] = textureGatherOffset( s_dmap, tc.xy, int2( 1, -3 ) );
  s_dmap_gpp[9] = textureGatherOffset( s_dmap, tc.xy, int2( 1, -1 ) );
  s_dmap_gpp[10] = textureGatherOffset( s_dmap, tc.xy, int2( 1,  1 ) );
  s_dmap_gpp[11] = textureGatherOffset( s_dmap, tc.xy, int2( 1,  3 ) );          

  s_dmap_gpp[12] = textureGatherOffset( s_dmap, tc.xy, int2( 3, -3 ) );
  s_dmap_gpp[13] = textureGatherOffset( s_dmap, tc.xy, int2( 3, -1 ) );
  s_dmap_gpp[14] = textureGatherOffset( s_dmap, tc.xy, int2( 3,  1 ) );
  s_dmap_gpp[15] = textureGatherOffset( s_dmap, tc.xy, int2( 3,  3 ) );   
```

This boilerplate code can be generated in C++, by the way.
Don't know if the approach is acceptable by the community.

Lnd-stoL:
> There are few small remarks and compilation fails on Windows x64, though.

They seem to be fixes by now.

src/Layers/xrRender/r__dsgraph_render.cpp
```diff
@@ -46,7 +47,7 @@ void R_dsgraph_structure::r_dsgraph_render_graph(u32 _priority)
         auto& map = mapNormalPasses[_priority][iPass];
 
         map.get_any_p(nrmPasses);
-        std::sort(nrmPasses.begin(), nrmPasses.end(), cmp_pass<mapNormal_T::value_type*>);
+        std::stable_sort(nrmPasses.begin(), nrmPasses.end(), cmp_pass<mapNormal_T::value_type*>);
```
Xottab-DUTY: Why sort algorithm and comparator was changed? What are the results of these changes?

src/xrGame/reward_event_handler.h
```diff
@@ -11,7 +11,7 @@ class reward_event_handler
 {
 public:
     reward_event_handler(game_state_accumulator* pstate) : m_player_state_accum(pstate) {}
-    ~reward_event_handler(){};
+    virtual ~reward_event_handler(){};
```
Xottab-DUTY, line 14: Let's let the compiler to generate everything else.
```suggestion
    virtual ~reward_event_handler() = default;
```

src/xrCore/xrMemory.cpp
```diff
@@ -112,7 +112,7 @@ size_t xrMemory::mem_usage()
         CloseHandle(h);
     }
     return pmc.PagefileUsage;
-#elif defined(XR_PLATFORM_LINUX)
+#elif defined(XR_PLATFORM_LINUX) || defined(XR_PLATFORM_APPLE)
     struct rusage ru;
     getrusage(RUSAGE_SELF, &ru);
     return (size_t)ru.ru_maxrss;
```
Xottab-DUTY, line 118: We can just by the way prevent missing return from function for possible future platforms.
```suggestion
    return (size_t)ru.ru_maxrss;
#else
#   error Select or add an implementation for your platform
```

src/Layers/xrRender/r__dsgraph_render.cpp
Lnd-stoL: As far as I understand, the comparator contained logical error leading to UB in resulting sort algorithm, causing crashes (somehow) on my Mac machine.
The error is that it returns true for two equal arguments, that means cmp(a, b) == cmp(b, a) if a.equal(b) or, in the other words, a < b && b < a is true. Such comparator behavior doesn't meet the C++ standard criteria https://en.cppreference.com/w/cpp/named_req/Compare and doesn't make much sense (how can we order two items if one orders before another, but another is also before the one).
Looking at the code I made a guess that the intent was actually to sort SPass'es based on ssa, but doesn't touch equal ones (though I didn't figure out, why there can be equal SPass'es and can they have different ssa). Fell free to correct me if I got it wrong.
As for stable sort, I just followed the intuition that it's always better not to mess up rendering order from frame to frame (which may cause kind of flickering). So, knowing there can be equal items in the nrmPasses I decided to elide possible sorting instability factor.

src/xrGame/reward_event_handler.h
Lnd-stoL: Ok, I just tried to limit the changes to the required minimum.

src/Layers/xrRenderPC_GL/rgl_shaders.cpp
Lnd-stoL: Let me know if it is acceptable approach or maybe there are some more experienced OpenGL people here, who can give a piece of advice.

src/Layers/xrRender/r__dsgraph_render.cpp
Xottab-DUTY: Maybe it's better to move this particular change to another pull request... Or, @vTurbine, can you tell if this change in comparator and stable sort is ok or even better than what we have currently?

Original engine code was sorting the scene graph in a special way to minimize state changes. I simplified the code to make SPass look like PSO (so we can add DX12/Vulkan in future), but this optimization was lost... but I tried to make comparator sort like in the original code.
Originally, sorting was performed with two criterias instead of just one: state change and SSA.
You can look at the original sorting code here: 5aeb5a64022628626688758d8a613d095511bddd. Maybe, you will be able to

StathisKap:
hey @Lnd-stoL I've been trying to compile your fork of the engine and I haven't been able too. I get this error
```
-- CMAKE_SYSTEM_PROCESSOR: arm64
-- MASTER_GOLD: ON
-- STATIC_BUILD: ON
Using mimalloc memory allocator
CMake build Release, CFLAGS=  -mfpu=neon, CXXFLAGS= -fdelayed-template-parsing -Wno-unused-command-line-argument -Wno-inconsistent-missing-override -stdlib=libc++ -Wno-attributes -fvisibility=hidden -Wl,-undefined,error  -mfpu=neon
git commit: 5d1a79146dc1a229880af0c6981a6292eb54dc15
git branch: dev
CMake Error at Externals/LuaJIT-proj/luajit.cmake:169 (message):
  Unsupported luajit target architecture
Call Stack (most recent call first):
  Externals/LuaJIT-proj/CMakeLists.txt:1 (include)


-- Configuring incomplete, errors occurred!
See also "/Users/stathis/Code/xray-16/build/CMakeFiles/CMakeOutput.log".
See also "/Users/stathis/Code/xray-16/build/CMakeFiles/CMakeError.log".
```

Do you know what I need to change?

pkolchanov:
@StathisKap try to clone with submodules  

`git clone https://github.com/Lnd-stoL/xray-16.git --recurse-submodules`

Lnd-stoL:
> hey @Lnd-stoL I've been trying to compile your fork of the engine and I haven't been able too. I get this error
> 
> ```
> -- CMAKE_SYSTEM_PROCESSOR: arm64
> -- MASTER_GOLD: ON
> -- STATIC_BUILD: ON
> Using mimalloc memory allocator
> CMake build Release, CFLAGS=  -mfpu=neon, CXXFLAGS= -fdelayed-template-parsing -Wno-unused-command-line-argument -Wno-inconsistent-missing-override -stdlib=libc++ -Wno-attributes -fvisibility=hidden -Wl,-undefined,error  -mfpu=neon
> git commit: 5d1a79146dc1a229880af0c6981a6292eb54dc15
> git branch: dev
> CMake Error at Externals/LuaJIT-proj/luajit.cmake:169 (message):
>   Unsupported luajit target architecture
> Call Stack (most recent call first):
>   Externals/LuaJIT-proj/CMakeLists.txt:1 (include)
> 
> 
> -- Configuring incomplete, errors occurred!
> See also "/Users/stathis/Code/xray-16/build/CMakeFiles/CMakeOutput.log".
> See also "/Users/stathis/Code/xray-16/build/CMakeFiles/CMakeError.log".
> ```
> 
> Do you know what I need to change?

@StathisKap, I faced this issue too and managed to get around it doing some dirty hacks, and basically, disabling jit  in luajit via cmake option. I can’t publish these changes via PR cause they are hacks, but can share the diff which helped me to get luajit building.
As far as I know, latest luajit actually has support for Mac M1, so it maybe worth trying the latest version instead of the x-ray repository one. If this doesn’t help, the issue should be addressed to luajit developers.

Xottab-DUTY:
Upstream LuaJIT won't work correctly with OpenXRay, we have applied custom patches to make it work...

src/Layers/xrRender/r__dsgraph_render.cpp
Lnd-stoL: The comment seems to be truncated, what was the question "Maybe, you will be able to")?

Lnd-stoL: > Maybe it's better to move this particular change to another pull request

Sure, it's possible. But this change is absolutely required to run the game on macOS: without it the game crashes right at the beginning, rendering the first frame.

Lnd-stoL: >  but I tried to make comparator sort like in the original code

As far as I can see the comparator is not like in the original code, but rather nonsense because of 

>  it returns true for two equal arguments, that means cmp(a, b) == cmp(b, a) if a.equal(b) or, in the other words, a < b && b < a is true

I'm pretty much surprised how it even works on windows and what are the resulting order of the items.

Lnd-stoL:
> Upstream LuaJIT won't work correctly with OpenXRay, we have applied custom patches to make it work...

Hmm, so it must be possible to apply OpenXRay patches on top of the latest upstream LuaJIT?
Where are these patches? What is the reason for these patches?

Xottab-DUTY:
> Where are these patches? What is the reason for these patches?

They are right in the fork as commits.
Main reason: original scripts were written for LuaJIT 1.1.x, while latest LuaJIT 2.1 has breaking behaviour changes and we have a restriction to not change game resources at all..

> Hmm, so it must be possible to apply OpenXRay patches on top of the latest upstream LuaJIT?

Maybe, I haven't work with git patches. 🙂

src/Layers/xrRender/r__dsgraph_render.cpp
Xottab-DUTY: @Lnd-stoL, so, with @vTurbine we have figured that we don't need stable sort here and it's better to return generic sort. Comparator change is ok and enough! 🙂

Lnd-stoL: Ok, thanks for figuring that out.

Xottab-DUTY (approved):

Xottab-DUTY:
Yay! Thank you!

PR #1139, opened 2022-11-25
Use default symbols visibility for macOS build.

Lnd-stoL:
This change is essential for game to run successfully on macOS.
Otherwise, there are errors like
`One or more of the following type_info's has hidden visibility or is defined in more than one translation unit. They should all have public visibility`
which eventually lead to crash.


Xottab-DUTY:
Thank you!

PR #1150, opened 2022-12-03
Added console command: spawn to inventory

xrSimplDimp:
Let's use the console command to add a section to the inventory


src/xrGame/console_commands.cpp
```diff
@@ -181,6 +181,33 @@ class CCC_MemStats : public IConsole_Command
 };
 
 // console commands
+class CCC_Spawn_to_inv : public IConsole_Command
+{
+public:
+    CCC_Spawn_to_inv(LPCSTR N) : IConsole_Command(N){};
+    virtual void Execute(LPCSTR args)
```
375gnu: better: `void Execute(LPCSTR args) override`

src/xrGame/console_commands.cpp
```diff
@@ -181,6 +181,33 @@ class CCC_MemStats : public IConsole_Command
...
+    {
+        if (!g_pGameLevel)
+        {
+            Log("Error: No game level!");
+            return;
+        }
+
+        if (!pSettings->section_exist(args))
+        {
+            Msg("! Section [%s] isn`t exist...", args);
```
375gnu: doesn't exist

src/xrGame/console_commands.cpp
```diff
@@ -181,6 +181,33 @@ class CCC_MemStats : public IConsole_Command
...
+            return;
+        }
+
+        char Name[128];
+        Name[0] = 0;
+        sscanf(args, "%s", Name);
```
375gnu: may i ask why do you need a local copy? and scanf is unsafe

src/xrGame/console_commands.cpp
```diff
@@ -181,6 +181,33 @@ class CCC_MemStats : public IConsole_Command
...
+
+        Level().spawn_item(Name, Actor()->Position(), false, Actor()->ID());
+    }
+    virtual void Info(TInfo& I) { strcpy(I, "name,team,squad,group"); }
```
375gnu: `void Info(TInfo& I) override ...`
xr_strcpy

src/xrGame/console_commands.cpp
xrSimplDimp: Well, I'll take this into account and change it to new methods

src/xrGame/console_commands.cpp
xrSimplDimp: > не существует
I'll redo it

xrSimplDimp: > не существует

I'll redo it

PR #1154, opened 2022-12-15
GitHub Actions: Build ARM/ARM64

Xottab-DUTY:


Xottab-DUTY:
@q4a just look at this! Look at the build times!

Xottab-DUTY:
ARM/ARM64 via QEMU on Alpine Linux under chroot
image

PR #1156, opened 2022-12-15
Github Actions: switch Alpine builds to Clang

Xottab-DUTY:
To speed up the build times.


Xottab-DUTY:
I don't know how to fix compilation errors, maybe, some package is missing...
It compiles on my local Alpine under WSL...

PR #1161, opened 2022-12-17
Replaced `push_back(std::make_pair(...))` to `emplace_back(...)`

GermanAizek:
@Xottab-DUTY,
Optimize inserts with `std::vector` if it was possible.


Xottab-DUTY (approved):

Xottab-DUTY:
Thanks!

PR #1167, opened 2022-12-18
Add time dilation when opening inventory and PDA

### LINKED ISSUE #1132: Add time dilation and DOF when opening inventory and PDA
Complementary feature for #943. Might go on par with #1373, because Anomaly has the blur effect we need in the inventory/dialogs.

- [x] Time dilation (#1167)
  - [x] Configurable: what time factor will be set, can be fully disabled or enabled only for inventory or only for PDA, etc...
- [ ] DOF: blur out everything except inventory, PDA UI
  - [ ] Configurable: how strong blur is, enabled at all or not.

Plotja:
#1132 

Add time dilation when opening inventory and PDA. 
Inventory sounds and ambient music is not dilated.

It can be configured through console commands:
"ui_time_factor" - time factor when opening inventory or PDA (values: 0.01 - 1.0)
"time_dilation_inventory" - is time dilation enabled in inventory (on/off)
"time_dilation_pda" - is time dilation enabled in pda (on/off)


src/xrGame/UIGameSP.h
```diff
@@ -2,6 +2,7 @@
 #include "UIGameCustom.h"
 #include "ui/UIDialogWnd.h"
 #include "xrAICore/Navigation/game_graph_space.h"
+#include <UITimeDilator.h>
```
Xottab-DUTY:
```suggestion
#include "UITimeDilator.h"
```

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you, this is awesome!

PR #1169, opened 2022-12-19
C++17 composition optimization and const param

GermanAizek:
@Xottab-DUTY, render_main R2 renderer is affected.


src/Layers/xrRender_R2/r2_R_render.cpp
```diff
@@ -76,11 +76,12 @@ void CRender::render_main(Fmatrix& m_ViewProjection, bool _fportals)
             {
                 ISpatial* spatial = lstRenderables[o_it];
                 spatial->spatial_updatesector();
-                CSector* sector = (CSector*)spatial->GetSpatialData().sector;
+                const auto& data = spatial->GetSpatialData();
+                const auto& [type, sphere, sector] = std::tuple(data.type, data.sphere, (CSector*)data.sector);
```
Xottab-DUTY, line 80: std::tuple here looks like using C++17 just for C++17.
More C++17 for the God of C++17 :)

Xottab-DUTY: This makes the code more complex for the compiler, there are more chances to be affected by the compiler bugs.
But, let's merge this and see how it goes.

PR #1171, opened 2022-12-19
Removed excess fields that are shadowing parent class fields

GermanAizek:
Fixed common typos when writing classes and inheriting them


src/xrGame/sight_manager.h
```diff
@@ -44,9 +44,8 @@ class CSightManager : public CSetupManager<CSightControlAction, CAI_Stalker, u32
         Fmatrix m_rotation;
     }; // struct bone_parameters_base
 
-    struct bone_parameters : bone_parameters_base
+    struct bone_parameters : public bone_parameters_base
```
Xottab-DUTY:
```suggestion
    struct bone_parameters : bone_parameters_base
```

PR #1176, opened 2022-12-21
Replaced own clamp funcs to native C++17 std::clamp and add noexcept

GermanAizek:
@Xottab-DUTY,
Possible optimization and improvement code security, as well as following C++17 standard.

### DIFF
--- a/src/xrCommon/math_funcs_inline.h
+++ b/src/xrCommon/math_funcs_inline.h
@@ -37,7 +37,14 @@ constexpr void clamp(T& val, const T& _low, const T& _high) noexcept
 
 // XXX: Check usages and provide overloads for native types where arguments are NOT references.
 template <class T>
-constexpr T clampr(const T& val, const T& _low, const T& _high) noexcept { return std::clamp(val, _low, _high); }
+constexpr T clampr(const T& val, const T& _low, const T& _high) noexcept
+{
+    if (val < _low)
+        return _low;
+    if (val > _high)
+        return _high;
+    return val;
+}
 
 inline float snapto(float value, float snap)
 {
--- a/src/xrCommon/math_funcs_inline.h
+++ b/src/xrCommon/math_funcs_inline.h
@@ -1,6 +1,5 @@
 #pragma once
-#include <cmath>
-#include <algorithm>
+#include <math.h>
 #include "xrCore/math_constants.h"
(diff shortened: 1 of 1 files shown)


src/xrCommon/math_funcs_inline.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
-#include <math.h>
+#include <cmath>
+#include <algorithm>
```
Xottab-DUTY:
```suggestion
#include <math.h>
```

src/xrCommon/math_funcs_inline.h
```diff
@@ -27,24 +28,11 @@ constexpr double rad2deg(double val) noexcept { return val*180 / M_PI;}
...
 // clamping/snapping
 template <class T>
-constexpr void clamp(T& val, const T& _low, const T& _high)
-{
-    if (val<_low)
-        val = _low;
-    else if (val>_high)
-        val = _high;
-}
+constexpr void clamp(T& val, const T& _low, const T& _high) noexcept { val = std::clamp(val, _low, _high); }
```
Xottab-DUTY:
```suggestion
constexpr void clamp(T& val, const T& _low, const T& _high) noexcept
{
    if (val<_low)
        val = _low;
    else if (val>_high)
        val = _high;
}
```

Xottab-DUTY (commented):
`algorithm` is a heavy header to include which may slow down the compilation. We don't want that just for the sake of std or new C++ features.

src/xrCommon/math_funcs_inline.h
```diff
@@ -37,14 +37,7 @@ constexpr void clamp(T& val, const T& _low, const T& _high)
...
 template <class T>
-constexpr T clampr(const T& val, const T& _low, const T& _high)
-{
-    if (val < _low)
-        return _low;
-    if (val > _high)
-        return _high;
-    return val;
-}
+constexpr T clampr(const T& val, const T& _low, const T& _high) noexcept { return std::clamp(val, _low, _high); }
```
Xottab-DUTY:
```suggestion
constexpr T clampr(const T& val, const T& _low, const T& _high) noexcept
{
    if (val < _low)
        return _low;
    if (val > _high)
        return _high;
    return val;
}
```

PR #1191, opened 2022-12-25
Layers/xrRender/ParticleGroup.h: use C++11 to initialize elements in the constructor

GermanAizek:
image

Difference constructor classes with CParticleEffect

image

### DIFF
--- a/src/Layers/xrRender/ParticleGroup.h
+++ b/src/Layers/xrRender/ParticleGroup.h
@@ -73,8 +73,8 @@ public:
 class ECORE_API CParticleGroup : public dxParticleCustom
 {
     const CPGDef* m_Def{};
-    float m_CurrentTime;
-    Fvector m_InitialPosition;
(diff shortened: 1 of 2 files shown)


src/Layers/xrRender/ParticleGroup.h
```diff
@@ -72,7 +72,7 @@ class ECORE_API CPGDef
 
 class ECORE_API CParticleGroup : public dxParticleCustom
 {
-    const CPGDef* m_Def;
+    CPGDef* m_Def;
```
Xottab-DUTY: Why remove const?
And we can just initialize the variable in the class declaration:
```suggestion
    const CPGDef* m_Def{};
```

GermanAizek: It's even better and more convenient this way

Xottab-DUTY (approved):

PR #1198, opened 2022-12-27
xrGame/ai_stalker_alife.cpp: use range-based for and emplace_back

GermanAizek:


src/xrGame/ai_stalker_alife.cpp
```diff
@@ -311,11 +302,9 @@ bool CAI_Stalker::enough_ammo(const CWeapon* new_weapon) const
 {
     int ammo_box_count = 0;
 
-    TIItemContainer::const_iterator I = inventory().m_all.begin();
-    TIItemContainer::const_iterator E = inventory().m_all.end();
-    for (; I != E; ++I)
+    for (const auto& item : inventory().m_all)
```
Xottab-DUTY, line 305: ~~It's better to extract `inventory().m_all` into new variable.~~
Not needed.

TmLev:
```push_back(Type(args...))``` is equivalent to  ```emplace_back(args...)```

File: src/xrGame/ai_stalker_alife.cpp (hunk 1 above)
Comment (Xottab-DUTY) [reply src/xrGame/ai_stalker_alife.cpp:305]: @GermanAizek

File: src/xrGame/ai_stalker_alife.cpp [hunk 2]
```diff
@@ -366,10 +355,8 @@ bool CAI_Stalker::can_take(CInventoryItem const* item)
     bool new_weapon_enough_ammo = enough_ammo(new_weapon);
     u32 new_weapon_rank = get_rank(new_weapon->cNameSect());
 
-    TIItemContainer::iterator I = inventory().m_all.begin();
-    TIItemContainer::iterator E = inventory().m_all.end();
-    for (; I != E; ++I)
-        if (conflicted(*I, new_weapon, new_weapon_enough_ammo, new_weapon_rank))
+    for (const auto& item : inventory().m_all)
```
Xottab-DUTY: ~~@GermanAizek, it's better to extract inventory().m_all into new variable.~~
Not needed.

Xottab-DUTY: @GermanAizek

Xottab-DUTY (approved):

PR #1204, opened 2023-01-03
xrGame: fix ability to get a negative money balance after paid service

### LINKED ISSUE #1173: Negative money balance after paid service
**Describe the bug**
Negative balance after a paid service if the player has less than the required amount.

**To Reproduce**
Steps to reproduce the behavior in Call of Pripyat:
1. Watch video

**Screenshots/Video**
<video src='https://user-images.githubusercontent.com/21138600/208969883-3ca459ad-c4be-4f44-9fcb-43e929ade094.mp4'/>

**Desktop:**
 - OS: any
 - OpenXRay and original Clear Sky and Call of Pripyat

Vlad2001MFS:
Why:
Money was checked when player press to a some question in dialog. If player by buy something at this moment, then your money state will be changed, but a condition for a service dialog wil not be rechecked.

How:
Just reset a talk dialog state to initial when player press to a trade button.

Issue: #1173


Xottab-DUTY:
From what I can guess, this can potentially cause more problems with broken dialog lines: you selected some phrase which triggered a dialog line, then switched to trade – now that line is broken: not all phrases finished, but you can't restart it again.
That's a corner case probably, but still can be a case.

We need to recheck if this approach can cause such problems or existing dialog system won't allow that and we can merge this.

For now, I think we need to improve money system itself – make it more hardened, prevent situations when money balance can turn negative.

PR #1209, opened 2023-01-07
Small fixes while inspecting the compiler warnings.

Lnd-stoL:


src/xrEngine/xr_input.h
```diff
@@ -150,7 +150,7 @@ class ENGINE_API CInput
     void RemoveKeyMapChangeWatcher(pureKeyMapChanged* watcher);
 
     CInput(const bool exclusive = true);
-    ~CInput();
+    virtual ~CInput();
```
Xottab-DUTY: Not needed. It's better to disable warning for this particular class.

src/xrSound/SoundRender_Emitter.h
```diff
@@ -110,5 +110,5 @@ class CSoundRender_Emitter : public CSound_emitter
     u32 play_time() override;
 
     CSoundRender_Emitter();
-    ~CSoundRender_Emitter();
+    virtual ~CSoundRender_Emitter();
```
Xottab-DUTY: Here, virtual destructor should be in the base class.

src/xrSound/SoundRender_Source.h
```diff
@@ -29,7 +29,7 @@ class XRSOUND_API CSoundRender_Source : public CSound_source
 
 public:
     CSoundRender_Source();
-    ~CSoundRender_Source();
+    virtual ~CSoundRender_Source();
```
Xottab-DUTY: Here, virtual destructor should be in the base class.

src/xrEngine/xr_input.h
Xottab-DUTY: Please, remove this change.

Xottab-DUTY (changes requested):
Please, use `u32(-1)`, `u16(-1)`, `u8(-1)` forms instead of `~0`. We don't want to go away from engine code style.

Xottab-DUTY (approved):

PR #1213, opened 2023-01-09
Increased node position size for large ai-maps

SkyLoaderr:
The current size is not enough to pack position, so it has been increased to 6 bytes.

### DIFF
--- a/src/Common/LevelStructure.hpp
+++ b/src/Common/LevelStructure.hpp
@@ -353,11 +353,11 @@ private:
     }
 
 public:
-    NodeCover5 high;
-    NodeCover5 low;
-    u16 plane;
-    NodePosition4 p;
-    // 13 + 2 + 2 + 2 + 5 = 24 bytes
+    NodeCover5 high;  // 2 bytes
+    NodeCover5 low;   // 2 bytes
+    u16 plane;        // 2 bytes
+    NodePosition12 p; // 6 bytes
+    // 13 + 2 + 2 + 2 + 6 = 25 bytes
 
     [[nodiscard]]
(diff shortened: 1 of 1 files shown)


src/Common/LevelStructure.hpp
```diff
@@ -353,11 +353,11 @@ struct NodeCompressed11
...
-    NodeCover5 high;
-    NodeCover5 low;
-    u16 plane;
-    NodePosition4 p;
-    // 13 + 2 + 2 + 2 + 5 = 24 bytes
+    NodeCover5 high;  // 2 bytes
+    NodeCover5 low;   // 2 bytes
+    u16 plane;        // 2 bytes
+    NodePosition12 p; // 6 bytes
+    // 13 + 2 + 2 + 2 + 6 = 25 bytes
```
Xottab-DUTY, line 360: NodeCompressed12 already has NodePosition12 included.

The versioning was taken from Living Zone fork: https://github.com/D1m0n93/xray-16/commit/251dd29c4eb1f0acba5cd2e579b0883a3249ae0d

So, in that commit it has xrAI version 12 and 11 is skipped. Why?
Because 11 is for abramcumner's AI-map: https://github.com/abramcumner/xray15/commit/459da9c8fb213e33bde273c57fe9dbf98467e05d

I decided not to change that and added just support both for 12th and 11th versions.

Xottab-DUTY:
So, instead of this change, you just need to use NodeCompressed12 and xrAI version 12 :)

... If it's not used already by your custom level.ai format, I hope.. trollsad

Xottab-DUTY:
So, let's decide on 11th and 12th formats in Discord...

PR #1218, opened 2023-01-11
xrGame/UIActorMenuInitialize.cpp: update conditions for initialize slot for backpacks

ChugunovRoman:
I tested this changes on CoP and work fine. There were no crashes

### DIFF
--- a/src/xrGame/ui/UIActorMenuInitialize.cpp
+++ b/src/xrGame/ui/UIActorMenuInitialize.cpp
@@ -334,7 +334,7 @@ void CUIActorMenu::InitializeInventoryMode(CUIXml& uiXml)
     };
     for (auto [id, section, parent] : inventory_lists)
     {
-        if (!m_pLists[id])
+        if (!uiXml.NavigateToNode(section))
             continue;
         if (id != eInventoryOutfitList)
             m_pLists[id] = UIHelper::CreateDragDropListEx(uiXml, section, parent);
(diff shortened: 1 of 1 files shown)


src/xrGame/ui/UIActorMenuInitialize.cpp
```diff
@@ -327,13 +327,15 @@ void CUIActorMenu::InitializeInventoryMode(CUIXml& uiXml)
...
-        //{ eInventoryBackpackList,  "dragdrop_backpack",  m_pInventoryWnd },
+        { eInventoryBackpackList,  "dragdrop_backpack",  m_pInventoryWnd },
         { eInventoryOutfitList,    "dragdrop_outfit",    m_pInventoryWnd },
         { eInventoryBeltList,      "dragdrop_belt",      m_pInventoryWnd },
         { eInventoryBagList,       "dragdrop_bag",       m_ActorWeightBar->m_BagWnd },
     };
     for (auto [id, section, parent] : inventory_lists)
     {
+        if (!m_pLists[id])
+            continue;
```
Xottab-DUTY: This code is wrong, because that `if` is always true.

ChugunovRoman: Need compare with nullptr?

ChugunovRoman: Ah ok, I understood. I replaced on `!uiXml.NavigateToNode(section)` this should work

Xottab-DUTY (approved):

PR #1222, opened 2023-01-13
Fix 647 revised

johncurley:
This code adds error checking for glTexStorage functions that were causing false positives on the error stack further down the code path on Linux. The errors are now outputted to the game log, allowing the error queue to be flushed and the theora video code to be loaded without any issues.

Note: There is still an uncaught error on MacOS however both green screen issues were not related to each other. This will be fixed in a further update once the error is correctly located and dealt with.


johncurley:
Sorry I accidentally included the theora fix into this code. You can merge the other pull request first or I can delete it if necessary.

PR #1229, opened 2023-01-19
Efx environment

johncurley:
EFX environment code port. There is currently an error flush routine for OpenAL as else the sound cuts out. I'll attempt to find that error and clean the code up properly at some stage but it would be good to test this on other operating systems as EFX is software accelerated. Seems to function well on pulse audio.


.gitignore
```diff
@@ -27,3 +27,6 @@ CMakeFiles/
 
 # exclude doxygen output
 doc/doxygen/html/*
+
+# exclude project files
+*.kdev4
```
Xottab-DUTY: This should go as another pull request

Xottab-DUTY:
Please, don't remove EAX code

johncurley:
I'll rewrite the environment code to have parallel code paths for both EFX and EAX support with the old code.

johncurley:
I'll see if I can find the source of the error causing the Linux sound to stop using EFX.

johncurley:
I'm going to fix all this up and repull.

PR #1231, opened 2023-01-23
Improved EFX support

johncurley:
Effects appear to be fully working. Please test on Windows but otherwise should be ready to use. :)


johncurley:
I might go through and review this code as well and make sure that all the right effects are in the right place after the last commit. It would be nice to test this on Windows as well since I've only tested on Linux. EFX should hopefully make all previous iterations of EAX unnecessary as it's EAX 5.0 equivalent.

johncurley:
On another note, I'm not an audio engineer so right now we're stuck with the OpenALSoft presets for environment files. We could further enhance this by getting someone who understands sound engineering to code effects into audio maps in the editor. This isn't necessary to fix the sound bugs though and it would just be a nice enhancement for modders or anyone skilled with sound engineering.

https://github.com/kcat/openal-soft/issues/61

Xottab-DUTY:
Is this ready to merge? :)

johncurley:
> Is this ready to merge? :)

It should be. I can't think of anything that is broken in it.

johncurley:
There are more effects but they're not present in the original games as they use later versions of EAX/EFX. However I can add them in another commit after they've been tested.

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you!

PR #1232, opened 2023-01-25
Added error check for error queue flushing for green video fix

johncurley:
The green video error on Mac was different to what was causing the issue on Linux. I've caught the error and flushed the queue in the correct place but it'll be worth looking into what is causing the array to not draw correctly.


Xottab-DUTY:
Is it possible to determine the source of the error?

johncurley:
> Is it possible to determine the source of the error?

I found some clues here:

https://stackoverflow.com/questions/21619917/gldrawelements-fails-with-error-gl-invalid-operation

I'll look into fixing the actual issue if possible rather than just catch the error.

Xottab-DUTY:
I believe this should not should go into production, because `CBackend::Render` functions are used to render everything in the game. But the solution here could help in backtracing the roots of the issue.

johncurley:
> I believe this should not should go into production, because `CBackend::Render` functions are used to render everything in the game. But the solution here could help in backtracing the roots of the issue.

To be honest this should probably be flagged as it doesn't solve the root of the issue. I'm still not sure what is causing the glDrawElementsBaseVertex calls to fail as they work sometimes and not others.

PR #1235, opened 2023-01-27
Dynamic byte alignment for cross architectural builds

johncurley:
Right now using the environmental variable MACOSX_DEPLOYMENT_TARGET set to 13.1 causes the debug build of Open X-Ray to fail on MacOSX aarch64 due to pointer size being out of alignment. This patch uses a macro from stack overflow to dynamically calculate the native pointer size of an unsigned int, allowing the code to compile without error for the two pointers.

https://stackoverflow.com/questions/51616057/how-to-determine-pointer-size-preprocessor-c


Xottab-DUTY:
No, just no!

1. The code you have provided fix for is correct and compliant C++ code.
2. We have ALIGN macro that should be used instead of GCC/Clang specific __attribute__
3. We have C++11 and it has `alignas` attribute which is preferred over ALIGN.

johncurley:
> 

Alright, will I have to use alignas manually and set the value for each architecture? Or do I use a sizeof?

Xottab-DUTY:
We should identify why Clang throws warnings/errors and fix them.... I just don't believe we should provide any explicit alignment to the structures. I see it as a compiler problem or a problem somewhere else in the code.


Can you show the compiler warnings/errors?

johncurley:
Will do. I'll recompile.

johncurley:
> We should identify why Clang throws warnings/errors and fix them.... I just don't believe we should provide any explicit alignment to the structures. I see it as a compiler problem or a problem somewhere else in the code.
> 
> Can you show the compiler warnings/errors?

`ld: warning: pointer not aligned at address 0x14EC167 ('_g_GameModes' + 36 from CMakeFiles/xrGame.dir/ui/UIOptConCom.cpp.o)
ld: warning: pointer not aligned at address 0x14EC15B ('_g_GameModes' + 24 from CMakeFiles/xrGame.dir/ui/UIOptConCom.cpp.o)
ld: warning: pointer not aligned at address 0x14EC14F ('_g_GameModes' + 12 from CMakeFiles/xrGame.dir/ui/UIOptConCom.cpp.o)
ld: warning: pointer not aligned at address 0x14EC143 ('_g_GameModes' + 0 from CMakeFiles/xrGame.dir/ui/UIOptConCom.cpp.o)
ld: warning: pointer not aligned at address 0x14EC51F ('_po_types' + 36 from CMakeFiles/xrGame.dir/__/xrServerEntities/xrServer_Objects_ALife.cpp.o)
ld: warning: pointer not aligned at address 0x14EC513 ('_po_types' + 24 from CMakeFiles/xrGame.dir/__/xrServerEntities/xrServer_Objects_ALife.cpp.o)
ld: warning: pointer not aligned at address 0x14EC507 ('_po_types' + 12 from CMakeFiles/xrGame.dir/__/xrServerEntities/xrServer_Objects_ALife.cpp.o)
ld: warning: pointer not aligned at address 0x14EC4FB ('_po_types' + 0 from CMakeFiles/xrGame.dir/__/xrServerEntities/xrServer_Objects_ALife.cpp.o)
ld: unaligned pointer(s) for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [/Users/johncurley/Projects/xray-16/bin/x64/Debug/xrGame.dylib] Error 1
make[1]: *** [src/xrGame/CMakeFiles/xrGame.dir/all] Error 2
make: *** [all] Error 2
20:09:52: The process "/opt/local/bin/cmake" exited with code 2.
Error while building/deploying project OpenXRay (kit: Desktop Qt 5.15.8 from MacPorts arm64)
When executing step "Build"`

src/xrCore/xr_token.h
```diff
@@ -1,15 +1,15 @@
...
 
     pcstr name;
     int id;
 };
-#pragma pack(pop)
+static_assert(sizeof(xr_token) == sizeof(pcstr) * 2, "xr_token should be aligned, otherwise it may have problems on RISC (e.g. ARM) architectures, which require aligned pointers.");
 
 XRCORE_API pcstr get_token_name(const xr_token* tokens, int key);
 XRCORE_API int get_token_id(const xr_token* tokens, pcstr key);
+
```
Xottab-DUTY:
```suggestion
```

Xottab-DUTY:
Thank you for the effort!

PR #1251, opened 2023-02-16
changed some shader preprocessor includes to allow macos compiling

johncurley:
Some preprocessor macros and functions weren't working correctly for the shadow code preventing it being compiled on MacOS and possibly Linux. This didn't fix the Mac green video issue though but it does allow the shadows to work.


johncurley:
This breaks some things on Linux so I'll see what I can do to fix that before I recommit.

PR #1252, opened 2023-02-16
Fix shadows on macOS

johncurley:
Shadow shader was not compiling on MacOS. Player shader is still not working but now is not giving syntax errors and environmental shadows are appearing correctly on MacOS.


Xottab-DUTY (changes requested):
Please, remove non-functional changes like grammatical mistakes fixes, whitespace changes, etc.
we need these shaders look like original R3 shaders to be able to use diff tools with reduced differences count.

res/gamedata/shaders/gl/shadow.h
```diff
@@ -155,35 +156,36 @@ float shadow_extreme_quality( float3 tc )
     {
         for( int col = -FS2; col <= FS2; col +=2 )
         {
-            float4 d4;
-            float  b;
+          float4 d4;
```
Xottab-DUTY: You are using incorrect tab-size of 2, it should be 4.
And the body of this cycle still has the extra diff because original shaders use mixed tabs and spaces, and probably your code formatter replaced tabs with spaces.

johncurley:
I was working on xcode and probably set my tabs to 2 spaces. I'll grab the original shader code and just add the changes to  that as I think reformatting by hand might be too complicated.

johncurley:
I'm going to test Klaxon's water shader file on Windows, but please let me know if anything is still broken and I'll attempt to fix it before this gets approved for a merge.

johncurley:
So it turns out I only had to modify gather.ps and the rest of the shaders are working fine. I'm not sure why it was having issues with #endif statements and macros. I'll look into that further but the commit should be ready as it's giving no errors on MacOS.

johncurley:
From MacOS log running the original shaders:

! shader compilation failed: gl\accum_sun_near_nomsaa_nominmax.ps\204811110011000000000000100000131111332110000000001
! error:  ERROR: 0:4460: Argument 4 to function 'textureLodOffset' must be a constant expression
ERROR: 0:4461: Argument 4 to function 'textureLodOffset' must be a constant expression
ERROR: 0:4462: Argument 4 to function 'textureLodOffset' must be a constant expression
ERROR: 0:4463: Argument 4 to function 'textureLodOffset' must be a constant expression
ERROR: 0:4465: Use of undeclared identifier 's0'
ERROR: 0:4465: Use of undeclared identifier 's1'
ERROR: 0:4465: Use of undeclared identifier 's2'
ERROR: 0:4465: Use of undeclared identifier 's3'
ERROR: 0:4479: Argument 4 to function 'textureLodOffset' must be a constant expression
ERROR: 0:4480: Argument 4 to function 'textureLodOffset' must be a constant expression
ERROR: 0:4481: Argument 4 to function 'textureLodOffset' must be a constant expression
ERROR: 0:4482: Argument 4 to function 'textureLodOffset' must be a constant expression
ERROR: 0:4484: Use of undeclared identifier 's0'
ERROR: 0:4484: Use of undeclared identifier 's1'
ERROR: 0:4484: Use of undeclared identifier 's2'
ERROR: 0:4484: Use of undeclared identifier 's3'

vTurbine:
This won't fly: `textureLod` expected texture coordinates in normalized range of floats, while `textureLodOffset` applies offset provided in texels. To make a similar thing, you'll need to normalize offset, dividing it by texture dimensions.

johncurley:
> This won't fly: `textureLod` expected texture coordinates in normalized range of floats, while `textureLodOffset` applies offset provided in texels. To make a similar thing, you'll need to normalize offset, dividing it by texture dimensions.

Ah I see. Both values are two different data types. I'll see what I can do to fix this up.

johncurley:
LVutner's solution worked and offset is now scaled and added to the tc value instead of the const int2 for the last argument. Shadows are correctly working on MacOS GL builds.

Xottab-DUTY:
Why did you closed it? 😅

johncurley:
> Why did you closed it? 😅

Green video fix? Because we weren't going to use it. I'll make another pull request for that as I don't have a solution at this point. This branch isn't closed is it?

Xottab-DUTY:
Oh! Oops, I just mixed these two different PRs...

johncurley:
> Oh! Oops, I just mixed these two different PRs...

It's my fault as I commented on the wrong branch for this comment. I can post some screenshots of the shadows for MacOS if anyone wants to double check.

Xottab-DUTY:
Yay!

PR #1257, opened 2023-02-20
Replaced C format concat to modern C++ std::to_string()

GermanAizek:
@Xottab-DUTY, Implementation `std::to_string()` works faster with different types than `sprintf()`
Also, C++ standard requires use of it, readability and fixing code also becomes easier.

References:
https://www.zverovich.net/2020/06/13/fast-int-to-string-revisited.html


Xottab-DUTY:
Using std::string means uncontrollable memory allocation, which we should not allow ideally.

We should not use std:: classes when we have xr_ classes

PR #1287, opened 2023-04-03
Rewrite old functor predicate to lambda expressions

GermanAizek:


src/Layers/xrRender/ResourceManager.h
Xottab-DUTY, line 1: Please, revert changes to the file.
Using std and C++17 just for the sake of std and C++17 WON'T give any advantages and, even worse, can introduce overhead in compile and RUN time!

Xottab-DUTY (changes requested):

src/xrGame/game_sv_item_respawner.cpp
```diff
@@ -279,8 +271,10 @@ void item_respawn_manager::add_new_rpoint(shared_str profile_sect, RPoint const&
 
 void item_respawn_manager::check_to_delete(u16 item_id)
 {
-    respawn_iter temp_iter =
-        std::find_if(m_respawns.begin(), m_respawns.end(), std::bind(search_by_id_predicate(), std::placeholders::_1, item_id));
+    auto temp_iter = std::find_if(m_respawns.begin(), m_respawns.end(), [&item_id](const spawn_item& respawn_item)
```
Xottab-DUTY:
```suggestion
    auto temp_iter = std::find_if(m_respawns.begin(), m_respawns.end(), [item_id](const spawn_item& respawn_item)
```

src/xrGame/ai/monsters/monster_event_manager.cpp
```diff
@@ -40,7 +40,10 @@ void CMonsterEventManager::raise(EEventType event, IEventData* data)
             (I->delegate)(data);
     }
 
-    auto it_del = std::remove_if(it->second.begin(), it->second.end(), pred_remove());
+    auto it_del = std::remove_if(it->second.begin(), it->second.end(), [](const event_struc& p)
+    {
+        return p.need_remove;
```
Xottab-DUTY:
```suggestion
    auto it_del = std::remove_if(it->second.begin(), it->second.end(), [](const event_struc& event)
    {
        return event.need_remove;
```

Xottab-DUTY (approved):

src/Layers/xrRender/ResourceManager.h
Xottab-DUTY: I did it myself :)

PR #1292, opened 2023-04-06
[Feature] optional console command to normalize mouse sens for weapons and upgrades

yohjimane:
**g_normalize_mouse_sens (0, 1)**
If set to 1, all weapons will have the same mouse sensitivity; upgrades will still modify sensitivity (scaled down to avoid negative sensitivity).

**g_normalize_upgrade_mouse_sens (0, 1)**
If set to 1, upgrades will not effect mouse sensitivity.

### DIFF
--- a/src/xrGame/inventory_item.cpp
+++ b/src/xrGame/inventory_item.cpp
@@ -31,6 +31,7 @@
 constexpr pcstr INV_NAME_KEY = "inv_name";
 constexpr pcstr INV_NAME_SHORT_KEY = "inv_name_short";
 constexpr pcstr DESCRIPTION_KEY = "description";
+extern int g_normalize_mouse_sens;
 
 net_updateInvData* CInventoryItem::NetSync()
 {
@@ -86,8 +87,6 @@ CInventoryItem::~CInventoryItem()
     }
 #endif // #ifndef MASTER_GOLD
 }
-
-extern int g_normalize_mouse_sens;
 
 void CInventoryItem::Load(LPCSTR section)
 {
--- a/src/xrGame/inventory_item.cpp
+++ b/src/xrGame/inventory_item.cpp
@@ -87,6 +87,8 @@ CInventoryItem::~CInventoryItem()
 #endif // #ifndef MASTER_GOLD
 }
 
+extern int g_normalize_mouse_sens;
+
 void CInventoryItem::Load(LPCSTR section)
 {
     CHitImmunity::LoadImmunities(pSettings->r_string(section, "immunities_sect"), pSettings);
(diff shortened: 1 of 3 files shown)


src/xrGame/inventory_item.cpp
```diff
@@ -87,6 +87,8 @@ CInventoryItem::~CInventoryItem()
 #endif // #ifndef MASTER_GOLD
 }
 
+extern int g_normalize_mouse_sens;
```
Xottab-DUTY: Please, place it to the beginning of the file: after `DESCRIPTION_KEY` and before `CInventoryItem::NetSync`

src/xrGame/inventory_item_upgrade.cpp
```diff
@@ -179,8 +182,27 @@ bool CInventoryItem::install_upgrade_impl(LPCSTR section, bool test)
         }
         result |= result2;
 
-        result |=
-            process_if_exists(section, "control_inertion_factor", &CInifile::r_float, m_fControlInertionFactor, test);
+        if (!g_normalize_upgrade_mouse_sens)
+            result |=
+                process_if_exists(section, "control_inertion_factor", &CInifile::r_float, m_fControlInertionFactor, test);
```
Xottab-DUTY:
```suggestion
        if (!g_normalize_upgrade_mouse_sens)
        {
            result |= process_if_exists(section, "control_inertion_factor", &CInifile::r_float, m_fControlInertionFactor, test);
        }
```

src/xrGame/inventory_item_upgrade.cpp
```diff
@@ -179,8 +182,27 @@ bool CInventoryItem::install_upgrade_impl(LPCSTR section, bool test)
...
+
+        bool needsNormalizedUpgradeSens = g_normalize_mouse_sens && !g_normalize_upgrade_mouse_sens;
```
Xottab-DUTY:
```suggestion
        const bool needsNormalizedUpgradeSens = g_normalize_mouse_sens && !g_normalize_upgrade_mouse_sens;
```

Xottab-DUTY (changes requested):
There's a question: if both console commands are disabled, the behaviour stays vanilla, right?
Like, mouse sens will be less than usual if you have PKM in your hands? :)

yohjimane:
> There's a question: if both console commands are disabled, the behaviour stays vanilla, right? Like, mouse sens will be less than usual if you have PKM in your hands? :)

Yes that is right, if both options are disabled nothing is changed from vanilla behavior.

Xottab-DUTY (approved):

PR #1294, opened 2023-04-08
Added  -DMEMORY_ALLOCATOR=standard flag

fractalshine:
Due to failing mimalloc let cmake use  -DMEMORY_ALLOCATOR=standard

### DIFF
--- a/ArchLinux/PKGBUILD
+++ b/ArchLinux/PKGBUILD
@@ -27,9 +27,11 @@ build() {
 #  rm -f -R bin
    mkdir -p bin
    cd bin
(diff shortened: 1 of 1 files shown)


Xottab-DUTY:
It would be better to just set system allocator as default in the root CMakeLists.txt, because crash with mimalloc is not Arch specific, but Linux-specific overall.

For now I'll just close this, since @Masterkatze is currently working on a fix, and hopefully it will be available some time soon.

Xottab-DUTY:
Thanks for the pull request!

Issue #1278, opened 2023-03-29
Linux - now that EAX work it doesn't seem to be applied correclty


BielBdeLuna:
**Describe the bug**
the main menu, and the opening videos, are played with EAX effects, now this shouldn't be the case. it could be said that the whole game seems to be playing inside a big tunnel. so it doesn't seem to have changes in the EAX profiles during the gameplay ( I don't know how well is EAX implemented in the original Windows game to begin with ).

also some of the sounds of the UI ( like having a new mission, or arguably the radio messages (are those supposed to be played in the PDA or in a headgear set?) ) or sounds like the music, or the ambient sound beds, those shouldn't have EAX applied to them.

**To Reproduce**
just by opening the engine, all sounds will have already a fixed EAX profile running. so the menus and the opening videos should sound already wrong.

**Expected behavior**
use EAX profiles smarter

**Screenshots**
nope

**BugTrap error report**
none

**Desktop (please complete the following information):**
 - OS: [Ubuntu 22.10]
 - OpenXRay build version [3ee814fa0b1855e368d036dae600d6655565db64]

**Additional context**
is EAX in windows working? how well it works?

Xottab-DUTY:
Meh! We just came to the start when reverb is being applied to all sounds! Whatta heck, yay!

johncurley:
It's EFX replacements for EAX, one thing that might need to be adjusted is the effects gain as it's current set to maximum(1, though really it's a 10 as it can be broken down into fractional increments).

Xottab-DUTY:
Gain won't really fix the problem of effects being applied to the sounds that should not be affected...

johncurley:
> Gain won't really fix the problem of effects being applied to the sounds that should not be affected...

I don't really notice that myself. Only some of the sounds have the effects applied.

BielBdeLuna:
At the moment all sounds seems to use the same profile EFX even when they shouldn't use any. The gain of the EFX profile should descent and ascent naturally with the distance to the sound source like they actually do.

johncurley:
It does run through an update function but it is very loud and might overpower some other sounds. In terms of the scripts and what they have applied for the original game. Except sounds like footprints and ambient noises are not accelerated, they sound a lot quieter so maybe if we turn the effects gain down it might help the non EFX sounds be heard? As for having EFX in the menu, etc, I don't know. I don't have that issue.

BielBdeLuna:
[...]

PR #1297, opened 2023-04-15
Removed inappropriate effects for EFX on default sounds.

johncurley:
This is a temporary patch that lowers the default gain for EFX in line with EAX and also reverts AL_EFFECTSLOT_AUXILIARY_SEND_AUTO to false so it does not play on sound channels that it is not triggered for. This is a temporary patch until an actual working 3D sound implementation can be engineered as the original games did not have zone files and therefore no real way to zone audio effects.


johncurley:
This patch should close #1278 until further improvements can be designed and implemented. For now it will cancel the reverb but it is not a true 3D sound environment due to the inability to trigger effects in the environment maps.

src/xrSound/SoundRender_Environment.cpp
```diff
@@ -169,6 +175,7 @@ bool CSoundRender_Environment::load(IReader* fs)
 
         Room = fs->r_float();
         RoomHF = fs->r_float();
+        RoomLF = fs->r_float();
```
Xottab-DUTY: This is binary file loading, it should not change!

Xottab-DUTY: File is versioned, so order of loading values should not change.

Xottab-DUTY (changes requested):
Please, revert changes to `CSoundRender_Environment::load`.

Xottab-DUTY (approved):

PR #1305, opened 2023-04-19
Convert path separators when saving imgui config

Masterkatze:


src/Common/PlatformLinux.inl
```diff
@@ -1100,6 +1100,11 @@ inline void restore_path_separators(char * path)
     while (char* sep = strchr(path, '/')) *sep = '\\'; //
 }
 
+inline void platform_path_separators(char * path)
```
Xottab-DUTY: This function is a duplicate of `convert_path_separators` function 🙃

Masterkatze: Fixed

PR #1312, opened 2023-04-23
Re-fix resource linking

### LINKED ISSUE #1298: CLocatorAPI::setup_fs_path symlinking ignores CMAKE_INSTALL_DATAROOTDIR
**Describe the bug**
Introduced by https://github.com/OpenXRay/xray-16/pull/739, https://github.com/OpenXRay/xray-16/commit/c076c7a5780acdf1c95e5cb51982ff1ba7bb0d82:
https://github.com/OpenXRay/xray-16/blob/c393d2d8530f8f6795852affc56b33d9a3d6dd61/src/xrCore/LocatorAPI.cpp#L873-L887

This assumes that `fsgame.ltx` and `gamedata/shaders/gl` are always installed to `/usr/share/openxray`, which is not the case:
https://github.com/OpenXRay/xray-16/blob/83c0397f2b51e12f1507697c1f2a8e1095a62f5a/res/CMakeLists.txt#L3-L5

Building with a different [`CMAKE_INSTALL_PREFIX` or `CMAKE_INSTALL_DATAROOTDIR`](https://cmake.org/cmake/help/v3.26/module/GNUInstallDirs.html?highlight=gnuinstalldirs#result-variables) breaks this assumption and produces bogus symlinks.

**To Reproduce**
1. Build with `cmake -DCMAKE_INSTALL_PREFIX=/usr/local` and install
2. Run `xr_3da` in an environment where the pref path in the user's home doesn't exist
3. Observe that it produces bogus symlinks to `/usr/share/openxray`, while the files were actually installed to `/usr/local/share/openxray`

**Expected behavior**
The CMake script should make the value of `CMAKE_INSTALL_FULL_DATAROOTDIR` available to this code via a define (either under this or a less CMake-specific name) and the symlinking calls should be
```c++
symlink(CMAKE_INSTALL_FULL_DATAROOTDIR "/openxray/...", tmp);
```

**Screenshots**
image

**BugTrap error report**
n/a

 - OS: NixOS unstable
 - OpenXRay build version 83c0397f2b51e12f1507697c1f2a8e1095a62f5a

**Additional context**
Should be replicable on any Linux system.

OPNA2608:
Some corrections to #739:

1. `CMAKE_INSTALL_RPATH` is concatenated from `CMAKE_INSTALL_PREFIX` variable and `/lib`. This should instead use the `CMAKE_INSTALL_FULL_LIBDIR` variable.
2. Re-apply the resource linking fixes from #798, which seem to have been unintentionally dropped in a merge conflict resolve.
  (recap: Use `CMAKE_INSTALL_FULL_DATAROOTDIR` instead of hardcoding `/usr/share`, detect & replace bad symlinks)

1\. is untested but seems correct to me. 2. was previously merged, I've just re-applied the changes with some minimal changes.

Fixes #1298.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ set(CMAKE_PDB_OUTPUT_DIRECTORY ${COMPILE_OUTPUT_FOLDER})
 set(CMAKE_COMPILE_PDB_OUTPUT_DIRECTORY ${COMPILE_OUTPUT_FOLDER})
 
 # Set the shared object search path for the installation
-set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_PREFIX}/lib)
+set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR})
 
 # Set the shared object search path for the build tree
 # On Linux/BSD we can use $ORIGIN to search for modules next to the executable
--- a/src/xrCore/LocatorAPI.cpp
+++ b/src/xrCore/LocatorAPI.cpp
@@ -853,24 +853,45 @@ void CLocatorAPI::setup_fs_path(pcstr fs_name)
                 pref_path = SDL_GetPrefPath("GSC Game World", "S.T.A.L.K.E.R. - Call of Pripyat");
 
             /* A final decision must be made regarding the changed resources. Since only OpenGL shaders remain mandatory for Linux for the entire trilogy,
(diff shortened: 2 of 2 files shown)


Xottab-DUTY (approved):
I've found [a PR](https://github.com/abseil/abseil-cpp/pull/963) related to Item 1.
It seems correct to me too. LGTM!

Xottab-DUTY:
Thank you!!!

PR #1317, opened 2023-04-28
xrRender: optimizations and speedup (1/3)

vTurbine:
This splits renderer into subsystems and adds parallel scene traversal for each of them:
IMG_3199

Note: shadow casting lights are batched by 3 and re-use contexts from cascades calculation. In case if more contexts needed, one can increase `R__NUM_AUX_CONTEXTS`.
Some of locks can be avoided if gameplay code will be untangled.

Next steps are:
- parallel command lists recording (green events on the picture)
- faster scene/lights occlusion + some opts.


Xottab-DUTY (approved):

Xottab-DUTY:
Hurray!

PR #1318, opened 2023-04-29
Added method reserve faces in xrCDB and code refactor

GermanAizek:


src/Layers/xrRender/NvTriStripObjects.cpp
```diff
@@ -652,20 +652,22 @@ void NvStripifier::RemoveSmallStrips(NvStripInfoVec& allStrips, NvStripInfoVec&
...
-        if (allStrips[i]->m_faces.size() < static_cast<size_t>(minStripLength))
+        const size_t lenList = strip->m_faces.size();
+        if (lenList < static_cast<size_t>(minStripLength))
         {
             // strip is too small, add faces to faceList
-            for (size_t j = 0; j < allStrips[i]->m_faces.size(); j++)
-                tempFaceList.push_back(allStrips[i]->m_faces[j]);
+            tempFaceList.reserve(lenList);
+            for (const auto& m_face : strip->m_faces)
+                tempFaceList.push_back(m_face);
```
Xottab-DUTY:
```suggestion
            for (const auto& face : strip->m_faces)
                tempFaceList.push_back(face);
```

src/Layers/xrRender/HOM.cpp
```diff
@@ -75,6 +75,7 @@ void CHOM::Load()
         IReader* S = fs->open_chunk(1);
         const auto begin = static_cast<HOM_poly*>(S->pointer());
         const auto end   = static_cast<HOM_poly*>(S->end());
+        CL.reserve_faces(S->length());
```
Xottab-DUTY: Using `S->length()` is wrong here.
There's so much pull requests from you with many mistakes like this or similar ones, so I won't tell you why this code is wrong. Learn it yourself, please, and raise the quality of your pull requests.

src/Layers/xrRenderPC_R1/FStaticRender_Loader.cpp
```diff
@@ -425,6 +425,7 @@ void CRender::LoadSectors(IReader* fs)
 
             if (do_rebuild)
             {
+                CL.reserve_faces(P.vertices.size() - 2);
```
Xottab-DUTY, line 441: I have a small suspect, that using `size() - 2` is wrong too, but not sure about this. Will check it a bit later..

src/Layers/xrRender/HOM.cpp
Xottab-DUTY: This may sound bitchy :) but wasting time and energy on reviewing continuous mistakes is not good too.

PR #1330, opened 2023-05-07
Restore original ffp ppe implementation 

yohjimane:
Referencing commit from SoC historical repo:
https://github.com/OpenXRay/xray-soc-history/commit/2dc9530fa6b11b270780d35d1f5d4e87836df686
https://github.com/OpenXRay/xray-soc-history/commit/998b717185d278f48fb4feb2f2ba3c02aecd4e46
https://github.com/OpenXRay/xray-soc-history/commit/9480c4414b8221d2f1773fb04064b4b9f5c9dcd5

Discord discussion: https://discord.com/channels/410170555619082240/489453264157409281/1104616744913489980

Pic is from ffp build, I also looked at this spot with rtx remix enabled and didn't get any weird rasterization issue like we typically would when ppe is applied.
image

It doesn't apply hue though when compared to vanilla. The noise effect is also a little different.
image

### DIFF
--- a/src/Layers/xrRenderPC_R1/FStaticRender_RenderTarget.cpp
+++ b/src/Layers/xrRenderPC_R1/FStaticRender_RenderTarget.cpp
@@ -2,8 +2,11 @@
 #include "FStaticRender_RenderTarget.h"
 #include "xrEngine/IGame_Persistent.h"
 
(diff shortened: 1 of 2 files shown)


Xottab-DUTY:
Truly, hue isn't applied, but that can be addressed later, probably.

PR #1331, opened 2023-05-07
xrRender: optimizations and speedup (2/3)

vTurbine:
This adds support for parallel cascades command lists recording (DX11 only):

выява
There is still room for improvement:
(1) - currently, render task waits for all cascades to be calculated before being kicked-off. This can be optimized by spawning a rendering task for each cascade separately
(2) - lights aren't rendered in parallel, but they can. I didn't add it because of two reasons: in CSM caching (which I plan to add soon) we will have only one spare smap slice; some of the tasks are just few draw calls, and it will be inefficient to parallelize them without any balancing. I will think about lights scheduling a bit later.

NOTE:
- GL render is quite broken, so this is a draft PR until I will stabilize it
- Backend class requires an urgent refactoring. At least, RT management should be moved into `CBackend`. I didn't do it because it is not in the scope of this PR


vTurbine:
I've added `set_pass_targets` to the `CBackend` class as the only method which should be used instead of explicit RT/ZB setters and that bunch of ugly `u_setrt` functions.

Xottab-DUTY (approved):

Xottab-DUTY:
Ура!

PR #1336, opened 2023-05-11
xrRender: optimize using reserve and set render_sun as const method

GermanAizek:


vTurbine:
Move these vectors out as the member of `render_sun_old` class and do the reserve once at construction time

GermanAizek:
@vTurbine, wow good idea. I'll do that.

Xottab-DUTY:
> @vTurbine, wow good idea. I'll do that.

@GermanAizek, so, when? :)

PR #1337, opened 2023-05-17
[Feature]: #1992 Add top, right, bottom, left alignment for CUIStatic

### LINKED ISSUE #1192: All alignment modes for CUIStatic.
I would like to use align the CUIStatic by left-right-top-bottom like a text alignment.

In vanilla we have only Center-mode for CUIStatic's alignment. By default, it reads only Center mode. 
```c++
bool CUIXmlInitBase::InitAlignment(CUIXml& xml_doc, const char* path, int index, float& x, float& y, CUIWindow* pWnd)
{
    xr_string wnd_alignment = xml_doc.ReadAttrib(path, index, "alignment", "");

    if (strchr(wnd_alignment.c_str(), 'c'))
        pWnd->SetAlignment(waCenter)
```
Also, if we will try to add this - it will be crash, because in `uiabstract.h` we have that code:
```c++
IC void GetWndRect(Frect& res) const
    {
        switch (m_alignment)
        {
        case waNone: res.set(m_wndPos.x, m_wndPos.y, m_wndPos.x + m_wndSize.x, m_wndPos.y + m_wndSize.y); break;
        case waCenter:
        {
            float half_w = m_wndSize.x / 2.0f;
            float half_h = m_wndSize.y / 2.0f;
            res.set(m_wndPos.x - half_w, m_wndPos.y - half_h, m_wndPos.x + half_w, m_wndPos.y + half_h);
        }
        break;
        default: NODEFAULT;
        };
    }
```
So, it crashes by NODEFAULT and it needs to add the other modes here. But i'm bad in math for this :hidethepain:.

yohjimane:
Resolves https://github.com/OpenXRay/xray-16/issues/1192

Example:

xrEngine_2023_05_17_12_16_05_514

Example code in ui_custom_msgs.xml:
```xml
  <test x="0" y="0" width="50" height="50" alignment="l">
          <text font="graffiti32"  r="240" g="217" b="182" a="255">Left</text>
  </test>
  <test2 x="0" y="0" width="50" height="50" alignment="r">
          <text font="graffiti32"  r="240" g="217" b="182" a="255">Right</text>
  </test2>
  <test3 x="500" y="0" width="50" height="50" alignment="t">
          <text font="graffiti32"  r="240" g="217" b="182" a="255">Top</text>
  </test3>
  <test4 x="500" y="0" width="50" height="50" alignment="b">
          <text font="graffiti32"  r="240" g="217" b="182" a="255">Bottom</text>
  </test4>
```

### DIFF
--- a/src/xrUICore/XML/UIXmlInitBase.cpp
+++ b/src/xrUICore/XML/UIXmlInitBase.cpp
@@ -1058,30 +1058,39 @@ void CUIXmlInitBase::ApplyAlign(float& x, float& y, u32 align)
 
 //////////////////////////////////////////////////////////////////////////
 
+// source: https://stackoverflow.com/questions/650162/why-cant-the-switch-statement-be-applied-to-strings
+constexpr uint32_t hash(const std::string_view data) noexcept 
+{ 
+    uint32_t hash = 5385;    
+    for (const auto& e : data)     
+        hash = ((hash << 5) + hash) + e;    
+    return hash; 
+}
+
 bool CUIXmlInitBase::InitAlignment(CUIXml& xml_doc, const char* path, int index, float& x, float& y, CUIWindow* pWnd)
 {
     // Alignment: top: "t", right: "r", bottom: "b", left: "l", center: "c"
     xr_string wnd_alignment = xml_doc.ReadAttrib(path, index, "alignment", "");
 
-    if (strchr(wnd_alignment.c_str(), 'r'))
+    switch (hash(wnd_alignment.c_str()))
     {
+    case hash("r"):
         pWnd->SetAlignment(waRight);
-    }
-    if (strchr(wnd_alignment.c_str(), 'l'))
-    {
+        break;
+    case hash("l"):
         pWnd->SetAlignment(waLeft);
-    }
-    if (strchr(wnd_alignment.c_str(), 't'))
-    {
+        break;
+    case hash("t"):
         pWnd->SetAlignment(waTop);
-    }
-    if (strchr(wnd_alignment.c_str(), 'b'))
-    {
+        break;
+    case hash("b"):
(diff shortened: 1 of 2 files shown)


src/xrUICore/XML/UIXmlInitBase.cpp
```diff
@@ -1058,32 +1058,62 @@ void CUIXmlInitBase::ApplyAlign(float& x, float& y, u32 align)
 
 //////////////////////////////////////////////////////////////////////////
 
+// source: https://stackoverflow.com/questions/650162/why-cant-the-switch-statement-be-applied-to-strings
+constexpr uint32_t hash(const std::string_view data) noexcept
```
yohjimane, line 1062: @Xottab-DUTY can you advise where to put this function? It can be used to refactor other large if statement blocks with strings -> switch cases.

Xottab-DUTY (approved):

src/xrUICore/XML/UIXmlInitBase.cpp
Xottab-DUTY: I didn't merged this PR in time only because I was thinking where to put this function, and then decided to think a bit later, and...

Just screw this "thinking". 😆 I'll just merge this, it's ok to keep this function here for now..

Xottab-DUTY:
@Hrusteckiy, hurray! Ura!!!

I really like this PR has 1337th number, hehe.

src/xrUICore/XML/UIXmlInitBase.cpp
Xottab-DUTY: Typically, this should be put somewhere in xrCore.

Hrusteckiy:
I'm glad, will check.

PR #1341, opened 2023-05-21
xrRender, xrCDB: use reserve to optimize inserts into containers

GermanAizek:


src/Layers/xrRender/ResourceManager_Reset.cpp
```diff
@@ -69,6 +69,7 @@ void CResourceManager::reset_end()
         // RT
         const size_t buffer_size = m_rtargets.size() * sizeof(CRT*);
         buffer_vector<CRT*> sorted_rts(xr_alloca(buffer_size), m_rtargets.size());
+        sorted_rts.reserve(m_rtargets.size());
```
Xottab-DUTY: This does nothing, buffer_vector already contains required amount of memory.
```suggestion
```

src/Layers/xrRender/r__dsgraph_render_lods.cpp
```diff
@@ -43,6 +43,7 @@ void R_dsgraph_structure::render_lods(bool _setup_zb, bool _clear)
     //Msg("dbg_lods: shader[%X]",u32((void*)firstV->shader._get()));
     //Msg("dbg_lods: shader_E[%X]",u32((void*)cur_S._get()));
 
+    lstLODgroups.reserve(lstLODs.size());
```
Xottab-DUTY: We don't need to reserve exactly `lstLODs.size()`, the actual amount is lower. And this array is being used every frame, so after few frames it stabilizes, so we don't need this.
```suggestion
```

PR #1344, opened 2023-05-23
If it is not static big array, better to use static constexpr

GermanAizek:
More info:
 - https://stackoverflow.com/questions/70711580/static-constexpr-vs-constexpr-in-function-body#answer-70765552
 - https://stackoverflow.com/questions/13865842/does-static-constexpr-variable-inside-a-function-make-sense#answer-66509089


Xottab-DUTY:
`static` specifier is not needed for global variables. Could you reduce the code length by removing `static` keyword from global variables?

GermanAizek:
@Xottab-DUTY, Yes you're right I'll fix it now

PR #1345, opened 2023-05-24
If it was possible rewritten lambdas to constexpr lambdas С++17

GermanAizek:


Xottab-DUTY:
1. Marking runtime-only functions as constexpr won't change anything.
2. `R_ASSERT` contains static variable for a reason – to make it possible to skip the assert just once and forever. `R_ASSERT2_NOSTATIC` will be triggered everytime, even if one clicked on a "Skip" button.

Issue #1024, opened 2022-05-07
Changed graphics settings can't be reverted automatically


OPNA2608:
**Describe the bug**
When you change some graphics settings (for example, resolution), the game will do some short reloading with the new settings and show a prompt with a timeout asking if you would like to keep these settings or revert them. Choosing to revert (or letting the prompt time out) will cause another reload, but it doesn't actually revert the settings.

**To Reproduce**
1. Go to 'Options'
2. Change the resolution
3. Hit 'Apply'
4. On the prompt, hit 'No' or let the timer run out
5. Observe that the settings haven't been reverted

**Expected behavior**
The game should do what it says it'll do: Revert the settings.

**Screenshots**

https://user-images.githubusercontent.com/23431373/167248418-6a912bf9-f493-4779-901a-4dda71f2c999.mp4

(The game window size & graphical artifacts after resizing is not a problem, just a curiosity of the recording program, but it helps with visualizing that the resolution doesn't in fact get reverted)

**BugTrap error report**
n/a

**Desktop (please complete the following information):**
 - OS: NixOS 21.11 (Linux)
 - OpenXRay build version: commit 78dd002c8c29c5eede7ff18eb5949e6c08576759

**Additional context**
n/a

Xottab-DUTY:
If this issue is only about the resolution settings, then it should have been fixed with #1383.
@OPNA2608, can you check it?

Xottab-DUTY:
So, I believe this issue is related to resolution only, other options should be reversible as usual.
Please, reopen if I'm incorrect.

PR #1350, opened 2023-05-24
Add customizable framerate cap (closes #1349)

### LINKED ISSUE #1349: Add command to customize menu framerate cap
Add something like `g_menu_fps_limit` for users with >60hz monitors.

yohjimane:
This solution is only approximate since updateDelta is an int.


yohjimane:
Working on menu UI for this.

yohjimane:
This feature is done, but needs doublechecking of Polish, Russian and German translations.

vTurbine:
Imo we should have menu capped to 30 FPS in all cases. There is nothing to show there at full speed. UX also won't suffer

yohjimane:
> Imo we should have menu capped to 30 FPS in all cases. There is nothing to show there at full speed. UX also won't suffer

The menu feels relatively sluggish when playing at high frame rate, attached is screenshot of game after opening menu. This doesn't negatively effect anything, and is also a feature found in many modern games.

image

Xottab-DUTY (approved):

Xottab-DUTY:
There's an issue – we can't use `depend="runtime"` because engine just sets FPS limit to a minimal value, when clamping it:
CUITrackBar isn't initialized yet and contains value 0, this value gets clamped to 30 and `rs_fps_limit 30` immediately gets called, which is not suitable totally.

We need to fix initialization orders it seems.

This might be related with #1024.

PR #1354, opened 2023-05-27
[Vanilla Bugfix]: Fix reversed input assignment, and dependent logic

yohjimane:
Bug: Mousewheel up is assigned to `kWPN_ZOOM_DEC`, mousewheel down is assigned to `kWPN_ZOOM_INC`.

Binocular also has reversed logic in `ZoomInc`/`ZoomDec` functions due to this.

Fix: Assign `kWPN_ZOOM_INC` to mousewheel up, and `kWPN_ZOOM_DEC` to mousewheel down, fix logic in binoculars.cpp.

### DIFF
--- a/src/xrGame/WeaponBinoculars.cpp
+++ b/src/xrGame/WeaponBinoculars.cpp
@@ -114,7 +114,7 @@ void CWeaponBinoculars::ZoomInc()
     float delta, min_zoom_factor;
     GetZoomData(m_zoom_params.m_fScopeZoomFactor, delta, min_zoom_factor);
 
-    float f = GetZoomFactor() - delta;
+    float f = GetZoomFactor() + delta;
     clamp(f, m_zoom_params.m_fScopeZoomFactor, min_zoom_factor);
     SetZoomFactor(f);
 }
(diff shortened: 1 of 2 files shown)


src/xrGame/WeaponBinoculars.cpp
```diff
@@ -124,7 +124,7 @@ void CWeaponBinoculars::ZoomDec()
     float delta, min_zoom_factor;
     GetZoomData(m_zoom_params.m_fScopeZoomFactor, delta, min_zoom_factor);
 
-    float f = GetZoomFactor() + delta;
+    float f = GetZoomFactor() - delta;
```
Xottab-DUTY, line 127: This is gameplay change and current behaviour could be done intentionally by GSC

yohjimane: The final result that the end-user sees is unaffected, so this does not change gameplay at all. Just makes more logical sense in code (imo).

If i revert the changes in `WeaponBinoculars.cpp`, then gameplay behavior would be changed.

Xottab-DUTY (approved):

PR #1357, opened 2023-05-28
Using constexpr strlen and fixed use defined strings

GermanAizek:


src/xrCore/xr_ini.cpp
```diff
@@ -736,7 +736,7 @@ CInifile::Sect& CInifile::r_section(pcstr S) const
     return **I;
 }
 
-pcstr CInifile::r_string(pcstr S, pcstr L) const
+constexpr pcstr CInifile::r_string(pcstr S, pcstr L) const
```
Xottab-DUTY, line 739: constexpr functions should be visible – thus, they should be in a header file.

This is not good for this function, and it won't help, since r_string only used in run-time.

Xottab-DUTY:
These xr_strlen usage take 0.00000001 of performance, and `std::char_traits<char>::length` uses a template which increases complexity of the code for the compiler and makes code a bit harder to read.
So, these changes are not helpful. I moved a few to a separate commit, though.

Thank you for the effort!

PR #1358, opened 2023-05-30
[Bugfix]: Save/load camera data with packet, fix incorrect camera yaw/pitch/roll on load

yohjimane:
### Description:
Camera yaw/pitch/roll can be incorrect when loading a save.

### Bug repro steps:
1. Load new game
2. Look all the way up
3. Save game
4. check `cam_Active()->pitch`  evaluates to -1.5 (which is within limits. 1.5 = looking down, -1.5 = looking up)
5. Load game from step 3
6. Observe with breakpoint in CameraBase.h line 79: `cam_Active()->pitch` is set to value of `o_torso` (evaluates to ~4.865). This value is both incorrect, and invalid for pitch, as pitch values must be between -1.5 to 1.5.

### Notes:
This bug is present, but does not cause any direct problems (that i can notice) in latest OpenXray dev build ([429519f](https://github.com/OpenXRay/xray-16/commit/429519f25ccf21f4f6dab5ede6ca9e4c19018599)). 

I found this bug while working on some hud inertia logic for my subproject, where some logic dependent on camera yaw value would get messed up after reloading a save while looking upwards.

### Solution:
The bug is fixed by using packet to save and load camera yaw/pitch/roll. Now, we only initialize yaw/pitch/roll values from `CSE_ALifeCreatureAbstract::o_torso` when packet data is missing.

Initializing values from `CSE_ALifeCreatureAbstract::o_torso` is necessary for correct camera orientation when loading into new levels.

Unfortunately this approach will make save files incompatible.


Xottab-DUTY:
Can we fix it without adding data to net packets? Doing so will cause previous game saves incompatibility

yohjimane:
@Xottab-DUTY i am not sure 😢 

I was thinking maybe we could just clamp the pitch/yaw values from `o_torso`, but the values are too different. For example, when camera pitch = -1.5 then o_torso pitch = 4.5, clamping that to proper limits would give inverted value of 1.5.

PR #1372, opened 2023-06-17
[Bugfix]: fix race condition preventing OnAnimationEnd call for shoot animation (closes #1371)

### LINKED ISSUE #1371: OnAnimationEnd callback unreliable for shoot animation
**Describe the bug**
OnAnimationEnd can be forced to get called by setting breakpoints, but without breakpoints it does not work for certain animations.

**To Reproduce**
1. Using mixed build, start and load into new game, attach debugger
2. Set breakpoint on [line 640 of WeaponMagazined.cpp](https://github.com/OpenXRay/xray-16/blob/dev/src/xrGame/WeaponMagazined.cpp#L640)
3. Shoot weapon
4. Observe breakpoint does not get hit
5. Set breakpoint on [lines 207](https://github.com/OpenXRay/xray-16/blob/dev/src/xrGame/HudItem.cpp#L207) and [214 ](https://github.com/OpenXRay/xray-16/blob/dev/src/xrGame/HudItem.cpp#L214) of HudItem.cpp
6. Shoot weapon
7. Breakpoint will get hit in this order: line 207 2x, line 214, line 640 of WeaponMagazined.cpp

**Expected behavior**
OnAnimationEnd should be called reliably on animation end.

**System info:**
 - Windows 10
 - Latest openxray dev: [663d2ba](https://github.com/OpenXRay/xray-16/commit/663d2ba6b9eb7a8018fb999e1982bc12f84d6719)

yohjimane:
Added a check to see if the shoot animation is complete before we call `StopShooting`

The bug was caused because `StopShooting` causes immediate state switch to `eIdle`. In this case, it is called just before `m_dwMotionCurrTm > m_dwMotionEndTm` can evaluate to true for shooting anim in `CHudItem::UpdateCL`, preventing `OnAnimationEnd` from getting called.

### DIFF
--- a/src/xrGame/WeaponFire.cpp
+++ b/src/xrGame/WeaponFire.cpp
@@ -132,8 +132,6 @@ void CWeapon::StopShooting()
     if (m_pFlameParticles && m_pFlameParticles->IsLooped())
         StopFlameParticles();
 
-    SwitchState(eIdle);
-
     bWorking = false;
 }
 
--- a/src/xrGame/WeaponMagazined.cpp
+++ b/src/xrGame/WeaponMagazined.cpp
@@ -588,7 +588,8 @@ void CWeaponMagazined::state_Fire(float dt)
         if (iAmmoElapsed == 0)
             OnMagazineEmpty();
 
-        StopShooting();
+        if (m_dwMotionCurrTm >= m_dwMotionEndTm)
+            StopShooting();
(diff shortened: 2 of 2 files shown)


Xottab-DUTY:
LGTM.

@yohjimane, hopefully, this should not cause any possible bugs? :D

PR #1375, opened 2023-06-25
[Bugfix]: #1374 - fix incorrect initialization CHUDState by setting eHidden as first value of EHudStates

### LINKED ISSUE #1374: CHUDState m_hud_item_state, m_nextState incorrect initialization on new game
**Describe the bug**
m_hud_item_state and m_nextState initialize with incorrect value `eIdle` on new game. The value they should initialize with is `eHidden`.

**To Reproduce**
1. put breakpoint in `CWeapon::OnActiveItem` function, on the first line
2. Load into new game with debugger enabled
3. On initial load when breakpoint is hit, observe CHUDState m_hud_item_state, m_nextState = 0 `eIdle`
4. Switch to a different weapon, observe CHUDState m_hud_item_state, m_nextState = 0 `eIdle`
5. Switch to initial weapon again, observe observe CHUDState m_hud_item_state, m_nextState = 3 `eHidden`

**Expected behavior**
On initial load, items should have eHidden state. It looks like [this logic should work](https://github.com/OpenXRay/xray-16/blob/dev/src/xrGame/HudItem.h#L38), but it does not.

**Screenshots**
Initial load state:
image

State after switching to different weapon, and then switching back
image

yohjimane:
avoids editing packets - yay!

### DIFF
--- a/src/xrGame/HudItem.h
+++ b/src/xrGame/HudItem.h
@@ -18,10 +18,10 @@ class CHUDState
 public:
     enum EHudStates
     {
-        eIdle = 0,
+        eHidden = 0,
+        eIdle,
         eShowing,
         eHiding,
-        eHidden,
         eBore,
         eLastBaseState = eBore,
     };


Xottab-DUTY (approved):

Xottab-DUTY:
Thanks!

PR #1390, opened 2023-07-02
[Bugfix]: Fix player_hud_tune item position/rotation adjustment

yohjimane:
adjusting item pos/rotation now works correctly :)

### DIFF
--- a/src/xrGame/player_hud_tune.cpp
+++ b/src/xrGame/player_hud_tune.cpp
@@ -349,32 +349,32 @@ void hud_adjust_mode_keyb(int dik)
 {
     if (pInput->iGetAsyncKeyState(SDL_SCANCODE_LSHIFT))
     {
-        if (dik == SDL_SCANCODE_KP_0)
+        if (dik == SDL_SCANCODE_KP_0 || dik == SDL_SCANCODE_0)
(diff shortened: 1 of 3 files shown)


src/xrGame/player_hud_tune.cpp
```diff
@@ -349,32 +349,32 @@ void hud_adjust_mode_keyb(int dik)
 {
     if (pInput->iGetAsyncKeyState(SDL_SCANCODE_LSHIFT))
     {
-        if (dik == SDL_SCANCODE_KP_0)
+        if (dik == SDL_SCANCODE_KP_0 || dik == SDL_SCANCODE_0)
```
yohjimane, line 352: @Xottab-DUTY is it okay to keep this change in? If you use keyboard without numpad (for instance, most 60%/66% layout keyboards or smaller), player_hud_tune is completely unusable unless your specific keyboard supports faking numpad inputs.

Xottab-DUTY (approved):

src/xrGame/player_hud_tune.cpp
Xottab-DUTY, line 352: Yes, it's ok

Xottab-DUTY:
Thank you!

PR #1396, opened 2023-07-12
WIP: [Feature] First Person Body

yohjimane:
See #1421 for list of bugs, which should be resolved before this is taken out of WIP status.


src/Layers/xrRender/SkeletonCustom.cpp
```diff
@@ -495,11 +495,28 @@ void CKinematics::LL_SetBoneVisible(u16 bone_id, BOOL val, BOOL bRecursive)
     visimask.set(mask, val);
     if (!visimask.is(mask))
     {
-        bone_instances[bone_id].mTransform.scale(0.f, 0.f, 0.f);
-    }
-    else
-    {
-        CalculateBones_Invalidate();
+        VERIFY(bone_id < LL_BoneCount());
```
yohjimane: This block of code is from Anomaly - fixes some issues with LL_SetBoneVisible.

Issue #1373, opened 2023-06-18
Support for Anomaly 


ghost:
**Is your feature request related to a problem? Please describe.**
I would like to create a new branch and adapt this engine for Stalker Anomaly, since its own fork is only contains game-logic changes but the performance is awful.

**Describe the solution you'd like**
Create new branch `/anomaly` and apply/refactor changes from here: https://bitbucket.org/anomalymod/xray-monolith/commits/ to make this fork compatible with **Anomaly** mod.

**Additional context**
I believe this will benefit both the playerbase and the team of this fork, due to the increased popularity.

Xottab-DUTY:
Do you need any help with that?

@ChugunovRoman was working on merging Call of Chernobyl to the latest OpenXRay. Anomaly is based on CoC, so Roman's work can help really much!

ghost:
@Xottab-DUTY Yes, I could use some help. 
I plan to investigate conflicting parts of the code between these two to see where code cannot be simply copy-pasted from anomaly sources. There will be needed merging and refactoring.
I think at first it would be correct to transfer only the mod logic, without touching parts related to the core engine's functionality.

LVutner:
I suggest to compare Anomaly code against: https://bitbucket.org/SeargeDP/coc-1.4-xray-64/src
Then, you can see what was added/removed/changed. Should be easier to port Anomaly functionality onto OpenXRay engine.

LVutner:
Most of the render code (excluding features required/accessed by LUA) can be ignored.

Xottab-DUTY:
@ChugunovRoman, could you give a tip about branches in your fork?
https://github.com/ChugunovRoman/xray-16

I see there are `coc` and `future_changes_for_coc` branches. Which should be used and what changes they contain?

ChugunovRoman:
> I see there are `coc` and `future_changes_for_coc` branches. Which should be used and what changes they contain?

@Xottab-DUTY The `future_changes_for_coc` branch include all latest changes which I moved from the CoC Engine

yohjimane:
This may have some overlap with the work I'm doing for #1398, as I have to port some anomaly functionality to support the shaders

mbehm:
> I suggest to compare Anomaly code against: https://bitbucket.org/SeargeDP/coc-1.4-xray-64/src Then, you can see what was added/removed/changed. Should be easier to port Anomaly functionality onto OpenXRay engine.

[...]

PR #1401, opened 2023-07-17
[Feature] Adding ini_file luabind constructor override.

Neloreck:
Adding additional constructor override for ini_file luabind export class.
As result, modders will be able to create ini files placed not only in gamedata/configs folder.

In my case it may be useful for `extensions` feature, where separate extensions can be packed as script+ltx files without file system conflicts.

## Changes:
- Additional ini_file constructor override


Xottab-DUTY (approved):
This looks similar feature to Anomaly, although, it's a bit different.
Could we make it the same as in Anomaly?

Neloreck:
@Xottab-DUTY 
Checked anomaly variant recently

1) For scripts they are using simple files list for-each and call `on_game_start` callback of every file if it is defined 
2) For ltx files anomaly modders are typically using dltx-modded exe (https://igigog.github.io/anomaly-modding-book/addons/dltx.html), which is not part of anomaly itself as I understand. It introduces new syntax for ini files (somewhere here https://github.com/themrdemonized/STALKER-Anomaly-modded-exes/blob/main/diff%20files%20for%20devs/dltx-bugfix.diff#L623)

In my variant I just simply add one more constructor for luabind ini file to open files not only from default configs folder.

Regarding monolith engine:

https://bitbucket.org/anomalymod/xray-monolith/src/d369db0ce75b8101c8d0fe1b2fa7a84b04910a35/src/xrServerEntities/script_ini_file.cpp#lines-20
https://bitbucket.org/anomalymod/xray-monolith/src/d369db0ce75b8101c8d0fe1b2fa7a84b04910a35/src/xrServerEntities/script_ini_file.h#lines-21

I believe similar approach is used here, but in anomaly variant override adds one optional parameter in the list end. If it is provided, it simply ignores first parameter with ternary operator.

---

What I would do as next steps:
- Still leave this new override since it follows general pattern where we can provide path base like $game_data$, $scripts$ and not simply ignore it
- I would think whether we need additional signature with same approach as anomaly engine is using. If it is needed, I can simply add it here
- I would think whether we need DLTX as part of openxray. Generally it looks nice, but changing all the implementation of ini parser is not what I can do and it can cause major regression since we do not have any unit tests as part of ini implementation. Adding it as R&D task/future thing is nice to have

Neloreck:
There is also thing called DXML: https://github.com/themrdemonized/STALKER-Anomaly-modded-exes/blob/main/DXML.md
People are creating interceptors and patch XML files in some order.

Did anyone ever proposed DLTX / DXML integration?

Xottab-DUTY:
> I would think whether we need additional signature with same approach as anomaly engine is using. If it is needed, I can simply add it here

It would be helpful, if we want to make life life easier for modders, that are used to Anomaly features. So, my initial suggestion had only a small scope of this: only adding function override like in Anomaly :)
I didn't even knew about DLTX/DXML! And that's awesome that you've mentioned it! I totally agree with creating an R&D task 💚 

> If it is needed, I can simply add it here

I fully support this, especially since we have #1373 :)

Neloreck:
I will add one more override to match anomaly as soon, as I have time then

src/xrServerEntities/script_ini_file.cpp
```diff
@@ -14,14 +14,18 @@
...
 CScriptIniFile::CScriptIniFile(IReader* F, LPCSTR path) : inherited(F, path) {}
-CScriptIniFile::CScriptIniFile(LPCSTR szFileName, BOOL ReadOnly, BOOL bLoadAtStart, BOOL SaveAtEnd)
-    : inherited(update(szFileName), ReadOnly, bLoadAtStart, SaveAtEnd)
+CScriptIniFile::CScriptIniFile(LPCSTR szFileName, BOOL ReadOnly, BOOL bLoadAtStart, BOOL SaveAtEnd, LPCSTR path)
+    : inherited(path ? path : update("$game_config$", szFileName), ReadOnly, bLoadAtStart, SaveAtEnd)
 {}
+CScriptIniFile::CScriptIniFile(LPCSTR initial, LPCSTR szFileName, BOOL ReadOnly, BOOL bLoadAtStart, BOOL SaveAtEnd)
+    : inherited(update(initial, szFileName), ReadOnly, bLoadAtStart, SaveAtEnd)
+{
+}
```
Xottab-DUTY:
```suggestion
CScriptIniFile::CScriptIniFile(LPCSTR szFileName, BOOL ReadOnly, BOOL bLoadAtStart, BOOL SaveAtEnd)
    : inherited(update("$game_config$", szFileName), ReadOnly, bLoadAtStart, SaveAtEnd) {}
CScriptIniFile::CScriptIniFile(LPCSTR initial, LPCSTR szFileName, BOOL ReadOnly, BOOL bLoadAtStart, BOOL SaveAtEnd)
    : inherited(update(initial, szFileName), ReadOnly, bLoadAtStart, SaveAtEnd) {}
```

src/xrServerEntities/script_ini_file.h
```diff
@@ -17,11 +17,12 @@ class CScriptIniFile : public CInifile
 
 public:
     CScriptIniFile(IReader* F, LPCSTR path = nullptr);
-    CScriptIniFile(LPCSTR szFileName, BOOL ReadOnly = TRUE, BOOL bLoadAtStart = TRUE, BOOL SaveAtEnd = TRUE);
+    CScriptIniFile(LPCSTR szFileName, BOOL ReadOnly = TRUE, BOOL bLoadAtStart = TRUE, BOOL SaveAtEnd = TRUE, LPCSTR path = NULL);
```
Xottab-DUTY:
```suggestion
    CScriptIniFile(LPCSTR szFileName, BOOL ReadOnly = TRUE, BOOL bLoadAtStart = TRUE, BOOL SaveAtEnd = TRUE);
```

Xottab-DUTY (approved):

Xottab-DUTY:
Anomaly matching constructor will be merged as a part of the Call of Chernobyl code merge, so I reverted it here to help merging process a bit.

Thank you!

Xottab-DUTY:
Call of Chernobyl support can be tracked in #1529.

DLTX and DXML support should be further tracked and discussed in #1373

PR #1404, opened 2023-07-23
Prevent weaponhide/kill/input disable on blowout end if godmode is on

yohjimane:
Prevents actor from dying and input from disabling after blowout end if you have godmode on


Xottab-DUTY:
I didn't merged that immediately in July, because I was thinking about cases where it could introduce problems... But let it be. God mode is not intended in normal player playthrough anyway, and we can fix any found problems later.

PR #1405, opened 2023-07-24
[FIX] Custom drag drop script export fix.

Neloreck:
## Changes:
- Fixing drag drop callback logic according to (https://github.com/Roman-n/xray-monolith__/blob/d404b95d92947e83b283d912111fe0957fb5fbb6/src/xrGame/ui/UIActorMenu_action.cpp#L99)

Without this fix bugs happened (game drag drops are just not usable when users have actor_menu_inventory.CUIActorMenu_OnItemDropped callback defined):
 - Items drop crashed game
 - You could attach bread etc to weapon
 - You could place weapon as artefact / armor etc
 - And many others

I think it happened because when code was moved from anomaly, it was not obvious that XRay implementation and Monolith engines have reversed conditions.

Checked in game drop events -> callback only firing when dropping item on item and not firing when moving/equiping/dropping items. 

---

Also I have noticed that we do not have check for `actor_menu_inventory` file existence before calling `CUIActorMenu_OnItemDropped` callback it it. As result, every drag-drop prints stack trace that file does not exist in game

### DIFF
--- a/src/xrGame/ui/UIActorMenu_action.cpp
+++ b/src/xrGame/ui/UIActorMenu_action.cpp
@@ -92,8 +92,12 @@ bool CUIActorMenu::DropItemOnAnotherItem(EDDListType t_old, EDDListType t_new, C
 
         const PIItem _iitem = _citem ? static_cast<PIItem>(_citem->m_pData) : nullptr;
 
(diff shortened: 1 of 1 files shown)


Xottab-DUTY (approved):
Thank you!

PR #1408, opened 2023-07-24
Common/PlatformApple.inl: Fix compilation with GCC

OPNA2608:
Get `locale_t` via `xlocale.h`:
image

Define `__fastcall`:
image


Xottab-DUTY:
Thanks!

PR #1420, opened 2023-08-16
More robust CUISequenceSimpleItem loading to prevent crashes when loading game tutorials 

sobkas:
Added NULL-checks, skipping not working subitems, increased verbosity of messages
Starting New Game Clear Sky with this change works
Need to also test COP

### DIFF
--- a/src/xrGame/ui/UIGameTutorialSimpleItem.cpp
+++ b/src/xrGame/ui/UIGameTutorialSimpleItem.cpp
@@ -109,7 +109,7 @@ void CUISequenceSimpleItem::Load(CUIXml* xml, int idx)
 
     // initialize auto_static
     int cnt = xml->GetNodesNum("main_wnd", 0, "auto_static");
-    m_subitems.resize(cnt);
+    m_subitems.reserve(cnt);
     string64 sname;
     for (int i = 0; i < cnt; ++i)
     {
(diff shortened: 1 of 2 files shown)


src/xrGame/ui/UIGameTutorial.cpp
Xottab-DUTY, line 1: Changes to this file totally not needed.
If there's no `game_tutorials.xml`, an assert will be triggered in `CUIXml::Load`. Other changes that add additional logging actually pollute the log.

src/xrGame/ui/UIGameTutorialSimpleItem.cpp
```diff
@@ -42,8 +42,12 @@ CUIWindow* find_child_window(CUIWindow* parent, const shared_str& _name)
     auto _I = wl.begin();
     auto _E = wl.end();
     for (; _I != _E; ++_I)
-        if ((*_I)->WindowName() == _name)
+    {
+        if ((*_I)->WindowName().equal(_name))
+        {
             return (*_I);
+        }
+    }
```
Xottab-DUTY: This is not needed too

Xottab-DUTY (changes requested):

src/xrGame/ui/UIGameTutorialSimpleItem.cpp
sobkas: Fixed

sobkas:
Fixed coding style problems

src/xrGame/ui/UIGameTutorial.cpp
```diff
@@ -114,6 +114,10 @@ bool CUISequencer::Start(LPCSTR tutor_name)
         Msg("! can't find tutorial [%s]", tutor_name);
         return false;
     }
+    else
+    {
+        Msg("found tutorial [%s]", tutor_name);
+    }
```
Xottab-DUTY: This is not needed too.
```suggestion
```

sobkas: Fixed

Xottab-DUTY:
So, I've found the real culprit of this crash. But we still need some changes to `CUISequenceSimpleItem` to make it more robust!
Let me do them in this PR 🙂

Xottab-DUTY:
Thank you!

PR #1422, opened 2023-08-19
[Linux]: Update debian/ folder

sobkas:
Move to new source format, make files more explicit, use quilt, add symbol file.
Depends on #1423


debian/control
```diff
@@ -2,11 +2,48 @@ Source: openxray
...
+
+Package: openxray-data
+Architecture: all
+Multi-Arch: foreign
+Depends:
+ ${misc:Depends},
+Description: X-Ray Engine data files (modules, templates and documentation)
+ ${source:Extended-Description}
+ .
+ This package provides openxray architecture independent data files (modules, templates, documentation etc.). Unless you have openxray installed, you probably do not need this package.
```
Kaffeine: Please end the file with a new line (https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file)

debian/openxray-data.install
```diff
@@ -0,0 +1,775 @@
...
+usr/share/icons/hicolor/48x48/apps/openxray_soc.png
+usr/share/icons/hicolor/64x64/apps/openxray_cop.png
+usr/share/icons/hicolor/64x64/apps/openxray_cs.png
+usr/share/icons/hicolor/64x64/apps/openxray_soc.png
+usr/share/icons/hicolor/32x32/apps/openxray_cop.png
+usr/share/icons/hicolor/32x32/apps/openxray_cs.png
+usr/share/icons/hicolor/32x32/apps/openxray_soc.png
+usr/share/pixmaps/openxray_cop.png
+usr/share/pixmaps/openxray_cs.png
+usr/share/pixmaps/openxray_soc.png
```
Kaffeine: Same.
image

Kaffeine:
What is the reason to have a 115,318 lines file with a list of symbols? The game is not supposed to have dependent packages which would require a rebuild on ABI changed.

Well, maybe you care about mods? 10.7 MB file needs some reasonable justification.

sobkas:
> What is the reason to have a 115,318 lines file with a list of symbols? The game is not supposed to have dependent packages which would require a rebuild on ABI changed.
> 
> Well, maybe you care about mods? 10.7 MB file needs some reasonable justification.

I was thinking about it and first thing that would be needed is a proper SONAME-s.
So maybe it would be better to wait with adding symbol file, maybe after looking if all symbols exported by internal libraries are really needed.

debian/rules
```diff
@@ -5,16 +5,21 @@ ifeq ($(findstring parallel=,$(DEB_BUILD_OPTIONS)),)
...
-UBUNTU_RELEASE  := $(shell lsb_release -sr)
+	dh $@ --with quilt --buildsystem=cmake --parallel
 
 extra_flags += \
--DCMAKE_C_COMPILER=gcc-8 \
--DCMAKE_CXX_COMPILER=g++-8 \
--DCMAKE_BUILD_TYPE="Release" \
+-DCMAKE_C_COMPILER=gcc-10 \
+-DCMAKE_CXX_COMPILER=g++-10 \
+-DCMAKE_BUILD_TYPE="RelWithDebInfo" \
```
Xottab-DUTY, line 13: Please, leave the Release build type!
Package is considered as a production version, in which we are stripping everything that is not needed for a player.

For any other purposes, e.g. debugging - custom build should be made from sources manually.

debian/patches/series
```diff
@@ -0,0 +1 @@
+001_dont_scan_app_root.patch
```
Xottab-DUTY: It's mad idea to provide a patch in the repository where we have original sources!!! :D

But even if we go ok with adding a patch, I strictly consider it as a separate pull request.

debian/rules
```diff
@@ -5,16 +5,21 @@ ifeq ($(findstring parallel=,$(DEB_BUILD_OPTIONS)),)
...
 
 override_dh_auto_configure:
 	dh_auto_configure -- $(extra_flags)
+override_dh_clean:
```
Xottab-DUTY, line 22:
```suggestion

override_dh_clean:
```

Xottab-DUTY:
> I was thinking about it and first thing that would be needed is a proper SONAME-s.
So maybe it would be better to wait with adding symbol file, maybe after looking if all symbols exported by internal libraries are really needed.

Could you please explain why we need it?

debian/rules
sobkas: During packaging deb, binaries are stripped and debug symbols are put into openxray-dbgsym package(so if someone want to fill bug report, they can install it and provide full backtrace) so binary looks like this:
file /usr/bin/xr_3da 
/usr/bin/xr_3da: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=86dead74faf750d85704dc78a7193342d7542018, stripped
"stripped" part indicates that there are no debug info there.
Same for libraries:
file /usr/lib/x86_64-linux-gnu/xrEngine.so
/usr/lib/x86_64-linux-gnu/xrEngine.so: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=b0b4edc31bf3824a70322795a3098c107c881c76, stripped

I'm using RelWithDebInfo so that I get release build with debug info, that is put in separate package.

Xottab-DUTY: Oh, so `openxray-dbgsym` gets created automatically?
This is awesome! I wanted to suggest the idea of a separate package with debug symbols instead of `RelWithDebInfo`, but it seems that this is what we actually need.

sobkas:
> > I was thinking about it and first thing that would be needed is a proper SONAME-s.
> > So maybe it would be better to wait with adding symbol file, maybe after looking if all symbols exported by internal libraries are really needed.
> 
> Could you please explain why we need it?

I was trying to say that we don't need symbol file for now(if ever) so there will be no such file in new revision.

debian/patches/series
sobkas: Problem is that if flags |= CLocatorAPI::flScanAppRoot; is present binary scans directory where it's placed, and in case of debian package it's /usr/bin
And in this case there is:
ls /usr/bin/|wc -l
2714 files
It takes a lot of time to do so.
I would prefer to make it a cmake switch rather than patch, but as it's now it's broken.
I think that patch as temporary solution and replacing it with cmake switch in future is ok.

debian/source/format
```diff
@@ -0,0 +1 @@
+3.0 (quilt)
```
Xottab-DUTY: Just in case, we can add a new line in the end of file.
```suggestion
3.0 (quilt)

```

Xottab-DUTY:
So, except for [1 comment](https://github.com/OpenXRay/xray-16/pull/1422#discussion_r1306691456), is this ready for merge? :)

sobkas:
> So, except for [1 comment](https://github.com/OpenXRay/xray-16/pull/1422#discussion_r1306691456), is this ready for merge? :)

Fixed, added debian/openxray-data.lintian-overrides and debian/copyright(placeholder for now, creating proper one will take some time)

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you very much!

PR #1423, opened 2023-08-20
[Linux]: Allow to disable problematic functionality

sobkas:
Option to disable RPATH and portable mode to be used for Linux packages
Needed for #1422

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,15 +32,25 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${COMPILE_OUTPUT_FOLDER})
 set(CMAKE_PDB_OUTPUT_DIRECTORY ${COMPILE_OUTPUT_FOLDER})
 set(CMAKE_COMPILE_PDB_OUTPUT_DIRECTORY ${COMPILE_OUTPUT_FOLDER})
 
-# Set the shared object search path for the installation
-set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR})
+if (DISABLE_PORTABLE_MODE)
+    add_definitions(-DDISABLE_PORTABLE_MODE)
+endif()
 
-# Set the shared object search path for the build tree
-# On Linux/BSD we can use $ORIGIN to search for modules next to the executable
-if (NOT APPLE)
-    set(CMAKE_BUILD_RPATH $ORIGIN)
-else()
-    set(CMAKE_BUILD_RPATH ${COMPILE_OUTPUT_FOLDER})
+if (NOT DEFINED ALLOW_RPATH)
+    set(ALLOW_RPATH TRUE)
+endif()
+
+if (ALLOW_RPATH)
+    # Set the shared object search path for the installation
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -32,15 +32,25 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${COMPILE_OUTPUT_FOLDER})
 set(CMAKE_PDB_OUTPUT_DIRECTORY ${COMPILE_OUTPUT_FOLDER})
 set(CMAKE_COMPILE_PDB_OUTPUT_DIRECTORY ${COMPILE_OUTPUT_FOLDER})
 
-# Set the shared object search path for the installation
-set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR})
+if(DISABLE_PORTABLE_MODE)
```
Xottab-DUTY:
```suggestion
if (DISABLE_PORTABLE_MODE)
```

CMakeLists.txt
```diff
@@ -32,15 +32,25 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${COMPILE_OUTPUT_FOLDER})
...
+    add_definitions(-DDISABLE_PORTABLE_MODE)
+endif()
 
-# Set the shared object search path for the build tree
-# On Linux/BSD we can use $ORIGIN to search for modules next to the executable
-if (NOT APPLE)
-    set(CMAKE_BUILD_RPATH $ORIGIN)
-else()
-    set(CMAKE_BUILD_RPATH ${COMPILE_OUTPUT_FOLDER})
+if(NOT DEFINED ALLOW_RPATH)
```
Xottab-DUTY:
```suggestion
if (NOT DEFINED ALLOW_RPATH)
```

Xottab-DUTY (approved):

src/xrCore/xrCore.cpp
```diff
@@ -275,8 +275,9 @@ void xrCore::Initialize(pcstr _ApplicationName, pcstr commandLine, LogCallback c
 #ifdef _EDITOR // for EDITORS - no cache
         flags &= ~CLocatorAPI::flCacheFiles;
 #endif // _EDITOR
+#ifndef DISABLE_PORTABLE_MODE
         flags |= CLocatorAPI::flScanAppRoot;
-
+#endif
```
Xottab-DUTY, line 280: Maybe we can just check in runtime if it's system folder?

sobkas: > Maybe we can just check in runtime if it's system folder?

Then what file/directory would be appropriate to check to be sure we are in system folder?
Second option is to search PATH for AppRoot so we know we are not in system folder.

Xottab-DUTY: Let's just merge the PR, as is for now. It's ok as is.

Xottab-DUTY: Maybe we'll find a robust way to determine a system folder.
I was thinking in terms of some sort of a function from SDL2 or the system itself :)

PR #1424, opened 2023-08-20
Allow player to kill themselves in singleplayer (debug only)

yohjimane:


src/xrGame/console_commands_mp.cpp
```diff
@@ -168,7 +168,15 @@ class CCC_Kill : public IConsole_Command
     virtual void Execute(LPCSTR args)
     {
         if (IsGameTypeSingle())
+        {
+#ifdef DEBUG
```
Xottab-DUTY:
```suggestion
#ifndef MASTER_GOLD
```

Xottab-DUTY (approved):
We have a policy to expose useful debug functionality in non Master Gold Release builds. I think this is ok to expose and could be useful, so I changed the #if condition a bit :)

PR #1425, opened 2023-08-20
Fix build warnings in src/Layers

Masterkatze:


src/Layers/xrRender/r__dsgraph_build.cpp
```diff
@@ -37,14 +37,12 @@ ICF float CalcSSA(float& distSQ, Fvector& C, float R)
 
 void R_dsgraph_structure::insert_dynamic(IRenderable* root, dxRender_Visual* pVisual, Fmatrix& xform, Fvector& Center)
 {
-    CRender& RI = RImplementation;
```
Xottab-DUTY, line 44: Why?

sobkas: It's useless, it's does nothing but waste memory.
It's only places that uses RI. instead of RImplementation.

Masterkatze: Use of `RI` was under `#ifdef`, but declaration was not, which gives warning about unused variable.

PR #1432, opened 2023-08-27
[Linux]: Fix other crash in Garbage

sobkas:
For some reason game sometimes crashes while in Garbage.

### DIFF
--- a/src/xrGame/quadtree.h
+++ b/src/xrGame/quadtree.h
@@ -50,6 +50,7 @@ public:
         IC T* get_object()
         {
             VERIFY(m_free);
+            VERIFY(m_free);
             if (!m_free)
                 return nullptr;
 
--- a/src/xrGame/quadtree.h
+++ b/src/xrGame/quadtree.h
@@ -50,6 +50,9 @@ public:
         IC T* get_object()
         {
             VERIFY(m_free);
+            if (!m_free)
+                return nullptr;
+
             T* node = m_free;
             m_free = m_free->next();
             ZeroMemory(node, sizeof(T));
--- a/src/xrGame/quadtree_inline.h
+++ b/src/xrGame/quadtree_inline.h
@@ -92,8 +92,14 @@ IC void CSQuadTree::insert(_object_type* object)
         if (depth == m_max_depth)
         {
             CListItem* list_item = m_list_items->get_object();
+            if (!list_item)
+                return;
+
             list_item->m_object = object;
             list_item->m_next = (CListItem*)((void*)(*node));
+            if (!list_item->m_next)
+                return;
+
             *node = (CQuadNode*)((void*)list_item);
             ++m_leaf_count;
             return;
@@ -101,6 +107,9 @@ IC void CSQuadTree::insert(_object_type* object)
 
         if (!*node)
             *node = m_nodes->get_object();
+
+        if (!node)
+            return;
 
         distance *= .5f;
         u32 index = neighbour_index(object->position(), center, distance);
@@ -267,6 +276,9 @@ IC _object_type* CSQuadTree::remove(
     const _object_type* object, CQuadNode*& node, Fvector center, float distance, int depth)
 {
     VERIFY(node);
+    if (!node)
+        return nullptr;
+
     if (depth == m_max_depth)
     {
         CListItem*& node_leaf = ((CListItem*&)((void*&)(node)));


src/xrGame/quadtree.h
```diff
@@ -49,7 +49,10 @@ class CQuadTree
         virtual ~CFixedStorage() { xr_free(m_objects); }
         IC T* get_object()
         {
-            VERIFY(m_free);
+            if (m_free == NULL)
+            {
+              return NULL;
+            }
```
Xottab-DUTY: Codestyle, plus it's better to leave VERIFY here, because it's not normal when m_free is null.
```suggestion
            VERIFY(m_free);
            if (!m_free)
                return nullptr;

```

src/xrGame/quadtree_inline.h
```diff
@@ -92,8 +92,20 @@ IC void CSQuadTree::insert(_object_type* object)
         if (depth == m_max_depth)
         {
             CListItem* list_item = m_list_items->get_object();
+
+            if (list_item == NULL)
+            {
+                return;
+            }
```
Xottab-DUTY:
```suggestion
            if (!list_item)
                return;
```

src/xrGame/quadtree_inline.h
```diff
@@ -266,7 +283,10 @@ TEMPLATE_SPECIALIZATION
 IC _object_type* CSQuadTree::remove(
     const _object_type* object, CQuadNode*& node, Fvector center, float distance, int depth)
 {
-    VERIFY(node);
+    if (node == NULL)
+    {
+        return NULL;
+    }
```
Xottab-DUTY:
```suggestion
    VERIFY(node);
    if (!node)
        return nullptr;

```

src/xrGame/quadtree_inline.h
```diff
@@ -102,6 +114,11 @@ IC void CSQuadTree::insert(_object_type* object)
         if (!*node)
             *node = m_nodes->get_object();
 
+        if (node == NULL)
```
Xottab-DUTY: `node` gets dereferenced 3 lines above :D
Also, we are in the cycle here. I'm not sure at the moment if we should return or continue the cycle.

Xottab-DUTY (changes requested):

src/xrGame/quadtree_inline.h
sobkas: Now get_object() can return null, needs to check for it.

sobkas: Based on code already there return is ok, continue causes crashes.

src/xrGame/quadtree.h
sobkas: Fixed

src/xrGame/quadtree_inline.h
sobkas: Fixed

src/xrGame/quadtree.h
Xottab-DUTY: Partially fixed, but not fully because:

1. We always use nullptr, NULL is not used (and is replaced to nullptr during code refactoring)
2. We don't check for nullptr explicitly, where it's possible to not do it.

sobkas: > Partially fixed, but not fully because:
> 
>     1. We always use nullptr, NULL is not used (and is replaced to nullptr during code refactoring)
> 
>     2. We don't check for nullptr explicitly, where it's possible to not do it.

Well there is need to check if we have nullptr, otherwise there is a crash.

Xottab-DUTY: > > Partially fixed, but not fully because:
> > ```
> > 1. We always use nullptr, NULL is not used (and is replaced to nullptr during code refactoring)
> > 
> > 2. We don't check for nullptr explicitly, where it's possible to not do it.
> > ```
> 
> Well there is need to check if we have nullptr, otherwise there is a crash.

I mean, we use `if (!ptr)` of `if (ptr)` checks rather than directly comparing to nullptr.

src/xrGame/quadtree_inline.h
Xottab-DUTY: > Now get_object() can return null, needs to check for it.

But here we check if `node` is null, while `get_object` assigns to `*node`, not `node` 🤔

src/xrGame/quadtree.h
```diff
@@ -49,7 +49,9 @@ class CQuadTree
         virtual ~CFixedStorage() { xr_free(m_objects); }
         IC T* get_object()
         {
-            VERIFY(m_free);
+            if (!m_free)
```
Xottab-DUTY, line 54:
```suggestion
            VERIFY(m_free);
            if (!m_free)
```

src/xrGame/quadtree_inline.h
```diff
@@ -92,8 +92,16 @@ IC void CSQuadTree::insert(_object_type* object)
         if (depth == m_max_depth)
         {
             CListItem* list_item = m_list_items->get_object();
+
```
Xottab-DUTY:
```suggestion
```

src/xrGame/quadtree_inline.h
```diff
@@ -92,8 +92,16 @@ IC void CSQuadTree::insert(_object_type* object)
...
+            if (!list_item)
+                return;
+
             list_item->m_object = object;
             list_item->m_next = (CListItem*)((void*)(*node));
+
+            if (!list_item->m_next )
```
Xottab-DUTY:
```suggestion
            if (!list_item->m_next)
```

Xottab-DUTY (commented):
In the `Files` tab of a pull request, you can just press `Add suggestion to batch` and merge all my suggestions as one commit instead of porting them and recommiting manually.

So, I'm again here, with these suggestions. :)
We still need to leave those VERIFY in the code, because original devs assumed that it is not normal if `m_free` is nullptr.
Since QuadTree is a generic container it's good to keep it robust with nullptr checks, but it's also good to keep these assertions (VERIFYs) for debugging purposes.

Xottab-DUTY:
Could you give me the save file with the crash?

src/xrGame/quadtree_inline.h
sobkas: > > Now get_object() can return null, needs to check for it.
> 
> But here we check if `node` is null, while `get_object` assigns to `*node`, not `node` 🤔



> > Now get_object() can return null, needs to check for it.
> 
> But here we check if `node` is null, while `get_object` assigns to `*node`, not `node` 🤔

node is CQuadNode** node so *node is pointer I think? Otherwise it wouldn't compile.

sobkas:
> Could you give me the save file with the crash?

I started shooting at bandits and it crashed(it might take some time).
[a234.sav.gz](https://github.com/OpenXRay/xray-16/files/12516292/a234.sav.gz)

sobkas:
> In the `Files` tab of a pull request, you can just press `Add suggestion to batch` and merge all my suggestions as one commit instead of porting them and recommiting manually.

I hate that there are two commit then, I prefer fix commits manually.
It's bad habit I took from using gerrit/baremetal git.

sobkas:
It's needed to be merged, without it I have rare but occurring crashes.

Xottab-DUTY:
Superseded by #1439, and then by #1440

src/xrGame/quadtree_inline.h
```diff
@@ -102,6 +108,9 @@ IC void CSQuadTree::insert(_object_type* object)
         if (!*node)
             *node = m_nodes->get_object();
 
+        if (!node)
```
Xottab-DUTY, line 111: So, [back to the old discussion](https://github.com/OpenXRay/xray-16/pull/1432#discussion_r1306691967), we can't do `*node` (dereference) two lines above if `node` is `nullptr`, so this `if (!node)` check is actually **late** and we need to check for nullptr earlier.

PR #1439, opened 2023-09-09
Fix crash in Garbage

sobkas:
For some reason game sometimes crashes while in Garbage.
It might also happen in other locations.


Xottab-DUTY:
This was accidentally closed, and then superseded by #1440

PR #1440, opened 2023-09-09
[Linux]: Fix crash in Garbage

sobkas:
For some reason game sometimes crashes while in Garbage. It might also happend in other locations.


src/xrGame/quadtree_inline.h
```diff
@@ -102,6 +108,9 @@ IC void CSQuadTree::insert(_object_type* object)
         if (!*node)
             *node = m_nodes->get_object();
 
+        if (!node)
+            return;
```
Xottab-DUTY, line 112: [Back to the old discussion](https://github.com/OpenXRay/xray-16/pull/1432#discussion_r1306691967), this check doesn't make sense, since we already dereferenced the `node` pointer two lines above.

Xottab-DUTY:
f08cf8083a914492f386f334779263dc2eace847 should have been fixed the issue.

PR #1449, opened 2023-09-19
[SHOC/Hang]: Reaching end of file while loading content from archive causes hang at startup.

sobkas:
LocatorAPI could reach end of file while loading content and got stuck in infinite loop.


Xottab-DUTY (approved):

Xottab-DUTY:
Thanks!

PR #1457, opened 2023-10-05
Fix static linking in cmake on linux #210

### LINKED ISSUE #210: Support static linking

Vlad2001MFS:
Related to issue #210


AMS21:
Is this still being worked on? I would really like to see something like this merged

Xottab-DUTY:
> Is this still being worked on? I would really like to see something like this merged

There are static variables that are being automatically added to the list of script export nodes. Linker just throws out these variables, and we didn't found a proper way to fix this yet.

This PR has a workaround for that – just manually specifying all nodes in https://github.com/OpenXRay/xray-16/pull/1457/files#diff-6336379129a38e317e4b780bc63a8bf0e146a9418fc5a7df830132326d261053, which is bad of course.

AMS21:
Alright thanks for the info, maybe I'll try to take a look at it

AMS21:
After some quick searching around I've found the [used](https://clang.llvm.org/docs/AttributeReference.html#used) and [retain](https://clang.llvm.org/docs/AttributeReference.html#retain) attributes which seem to do what we need here although retain is only available starting with clang and lld 13 or binutils 2.36 released 2021-01-24 and gcc 11

Xottab-DUTY:
So, I asked @Vlad2001MFS to send to PR to highlight the code areas that need to be changed to properly support static linking!
Much thanks for that!

We currently have both renderers and xrGame linked to xr_3da, and only problematic code that prevents us to enable static linking are script exporting nodes.
They are to be refactored, but for now this PR can be closed as it solved it's purpose.

Again thanks!

PR #1464, opened 2023-10-16
[Fix] Binder net packet reader io

Neloreck:
## Changes

- Fixing case when writing size_t to net packet, but reading u32, creates issues with custom binders for weapon objects and try to read data

### DIFF
--- a/src/xrGame/WeaponMagazinedWGrenade.cpp
+++ b/src/xrGame/WeaponMagazinedWGrenade.cpp
@@ -779,7 +779,7 @@ void CWeaponMagazinedWGrenade::save(NET_Packet& output_packet)
 {
     inherited::save(output_packet);
     save_data(m_bGrenadeMode, output_packet);
-    save_data(m_magazine2.size(), output_packet);
+    save_data(static_cast<u32>(m_magazine2.size()), output_packet);
 }
 
 void CWeaponMagazinedWGrenade::load(IReader& input_packet)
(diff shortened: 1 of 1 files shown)


Xottab-DUTY:
Ohhhh, that means we had incompatibility between x64 and x86 game saves also?

Xottab-DUTY:
~~Ohhh, this will introduce incompatibility between x64 and x86 game saves (they won't be identical after this change)~~
Oh, they aren't compatible already. This needs to be fixed.

Xottab-DUTY:
Can't look at the engine code right now to find the place where we actually write size_t to a net packet, so I just briefly say that we should never use size_t in net packets and binary formats, so we need to change the code to write u32 as it was before.
(or u64 if it's reaaally necessary)

Neloreck:
I would stick with u32. Then `save` method should be modified instead

Neloreck:
We write size of magazine here

```
    save_data(m_magazine2.size(), output_packet);
```

And get

```
    _NODISCARD _CONSTEXPR20 size_type size() const noexcept {
        auto& _My_data = _Mypair._Myval2;
        return static_cast<size_type>(_My_data._Mylast - _My_data._Myfirst);
    }
```

Should I just do static cast to u32 to make it work?
I believe it is very unlikely to get overflow here + it will match original 32bit engine

Xottab-DUTY:
@Neloreck I agree!

Xottab-DUTY (approved):

Xottab-DUTY:
Now we need to analyze all changes we have and merge everything that breaks saves compatibility within one time period (e.g. a day?)

PR #1465, opened 2023-10-25
Fix LuaJIT not passing cmake generator correctly

AMS21:
This would previously cause problems when your environments default generator is a multi-config generator like `Ninja Multi-Config` since LuaJIT would always use the default one and not the one OpenXRay was configured to build with.


Xottab-DUTY:
@Masterkatze, could you check it and merge?

Xottab-DUTY (changes requested):
@AMS21, it seems you have replaced tabs with spaces and that produced a diff larger than necessary, could you replace them back to tabs on the lines that you have changed?

AMS21:
Yeah my editor automatically replaces tabs with spaces. Fixed that

Masterkatze (approved):

Xottab-DUTY (approved):

PR #1466, opened 2023-10-25
Fix memory leak in `CCC_DebugFonts`

### LINKED ISSUE #1123: Fix memory leak in CCC_DebugFonts
https://github.com/OpenXRay/xray-16/blob/14a049e809c923488f77079d5d722198736d08fd/src/xrGame/console_commands.cpp#L1058

AMS21:
Not sure about the projects conventions for naming or handling OOM. But the original code also didn't care about `xr_new` potentially returning null.

This patch also prevents you from opening multiple font debug menus at the same time.

This fixes #1123

### DIFF
--- a/src/xrGame/console_commands.cpp
+++ b/src/xrGame/console_commands.cpp
@@ -1080,7 +1080,12 @@ class CCC_DebugFonts : public IConsole_Command
 {
 public:
     CCC_DebugFonts(LPCSTR N) : IConsole_Command(N) { bEmptyArgsHandled = true; }
-    ~CCC_DebugFonts() { xr_free(m_ui); }
+
+    ~CCC_DebugFonts()
+    {
+        xr_free(m_ui);
+    }
+
     virtual void Execute(LPCSTR args)
     {
         if (!m_ui)
--- a/src/xrGame/console_commands.cpp
+++ b/src/xrGame/console_commands.cpp
@@ -1080,11 +1080,17 @@ class CCC_DebugFonts : public IConsole_Command
 {
 public:
     CCC_DebugFonts(LPCSTR N) : IConsole_Command(N) { bEmptyArgsHandled = true; }
+    ~CCC_DebugFonts() { xr_free(m_ui); }
     virtual void Execute(LPCSTR args)
     {
-        // BUG: leak
-        (xr_new<CUIDebugFonts>())->ShowDialog(true);
+        if (!m_ui)
+            m_ui = xr_new<CUIDebugFonts>();
+
+        m_ui->ShowDialog(true);
     }
+
+private:
+    CUIDebugFonts* m_ui;
 };
 
 class CCC_DebugNode : public IConsole_Command


src/xrGame/console_commands.cpp
```diff
@@ -1080,11 +1080,17 @@ class CCC_DebugFonts : public IConsole_Command
...
-        // BUG: leak
-        (xr_new<CUIDebugFonts>())->ShowDialog(true);
+        if (!ui)
+            ui = xr_new<CUIDebugFonts>();
+
+        ui->ShowDialog(true);
     }
+
+private:
+    CUIDebugFonts* ui;
```
Xottab-DUTY: Let's name it `m_ui` to look it a bit better

src/xrGame/console_commands.cpp
```diff
@@ -1080,11 +1080,17 @@ class CCC_DebugFonts : public IConsole_Command
 {
 public:
     CCC_DebugFonts(LPCSTR N) : IConsole_Command(N) { bEmptyArgsHandled = true; }
+    ~CCC_DebugFonts() { xr_free(ui); }
     virtual void Execute(LPCSTR args)
     {
-        // BUG: leak
-        (xr_new<CUIDebugFonts>())->ShowDialog(true);
+        if (!ui)
+            ui = xr_new<CUIDebugFonts>();
```
Xottab-DUTY: With this solution, there's an issue: CUIDebugFonts calls xr_delete on itself when you press Escape, so CCC_DebugFonts class will store an invalid pointer after the memory deletion.

Xottab-DUTY (changes requested):

src/xrGame/console_commands.cpp
Hrusteckiy: Is it possible to draw more than one instance of one font type? I'd like to draw letterica18 in two places of screen for example.

Hrusteckiy:
Is it possible to draw more than one instance of one font type? I'd like to draw letterica18 in two places of screen for example.

Xottab-DUTY:
> Is it possible to draw more than one instance of one font type? I'd like to draw letterica18 in two places of screen for example.

This PR is not related to fonts, it's related to fonts debugger class :D
But yes, you can, just create new instance of a font.

src/xrGame/console_commands.cpp
AMS21: I've tried to test this but using `debug_fonts` multiple times seems to work without crashing.
Also looking at the code for `HideDialog` I don't see a deallocation but I may be missing something here.

src/xrGame/console_commands.cpp
```diff
@@ -1080,11 +1080,17 @@ class CCC_DebugFonts : public IConsole_Command
 {
 public:
     CCC_DebugFonts(LPCSTR N) : IConsole_Command(N) { bEmptyArgsHandled = true; }
+    ~CCC_DebugFonts() { xr_free(m_ui); }
```
Xottab-DUTY:
```suggestion

    ~CCC_DebugFonts()
    {
        xr_free(m_ui);
    }

```

Xottab-DUTY:
Thanks!

src/xrGame/console_commands.cpp
Xottab-DUTY: Ohhh, if it was calling xr_delete on itself, then there would be no memory leak, but there was no xr_delete call.
I got mixed up a bit))

src/xrGame/console_commands.cpp
```diff
@@ -1080,11 +1080,22 @@ class CCC_DebugFonts : public IConsole_Command
...
-        // BUG: leak
-        (xr_new<CUIDebugFonts>())->ShowDialog(true);
+        if (!m_ui)
+            m_ui = xr_new<CUIDebugFonts>();
+
+        m_ui->ShowDialog(true);
     }
+
+private:
+    CUIDebugFonts* m_ui;
```
Masterkatze, line 1098: `xr_unique_ptr`?

AMS21: That would be even better IMO

PR #1470, opened 2023-10-25
xrGame: Fix undelegated constructor for `FractionState`

AMS21:
Previously we would only create a temporary object which would immediately be destroyed instead of calling the default constructor.


Xottab-DUTY (approved):

Xottab-DUTY:
Thanks!

PR #1475, opened 2023-10-26
xrEngine: Merge CRenderDeviceData into CRenderDevice

AMS21:


Xottab-DUTY:
Oh! That was actually in my TODO backlog! Thank you!

Would you mind to refactor it even more?
It would be cool to make it look like old CRenderDevice from the previous version of the engine (e.g. `m_rcWindowBounds`, `m_rcWindowClient` being at the beginning of the class and etc.)
Look here: https://github.com/OpenXRay/xray-15/blob/xd_dev/cs/engine/xrEngine/device.h

AMS21:
Tried my best to match it to the CRenderDevice you linked :)

Xottab-DUTY (approved):

Xottab-DUTY:
Thanks!

PR #1476, opened 2023-10-26
xrCore: Fix null pointer dereference in `std::hash<shared_str>`

AMS21:
Caught with the undefined behavior sanitizer

### DIFF
--- a/src/xrCore/xrstring.h
+++ b/src/xrCore/xrstring.h
@@ -198,7 +198,8 @@ struct std::hash<shared_str>
 {
     [[nodiscard]] size_t operator()(const shared_str& str) const noexcept
     {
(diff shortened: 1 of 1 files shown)


src/xrCore/xrstring.h
```diff
@@ -198,6 +198,10 @@ struct std::hash<shared_str>
 {
     [[nodiscard]] size_t operator()(const shared_str& str) const noexcept
     {
+        if (str._get() == nullptr)
+        {
+            return std::hash<pcstr>{}(nullptr);
+        }
         return std::hash<pcstr>{}(str._get()->value);
```
Xottab-DUTY:
```suggestion
        const auto str_val = str._get();
        return std::hash<pcstr>{}(str_val ? str_val->value : nullptr);
```

Issue #1474, opened 2023-10-26
Thoughts on improving and enforcing coding style


AMS21:
Here are some general ideas to improve coding style.

Implementing these would generally be split into 2 steps.
1. Running the option on the entire code base.
2. Creating a GitHub workflow which run on every commit/pull request to verify it in the future.

- [x] Replacing tabs with spaces [ref](https://stackoverflow.com/questions/11492179/tabs-vs-space-indentation) #1477 
- [x] Source files should be encoded with UTF-8 #1480 [ref](https://softwareengineering.stackexchange.com/questions/152780/should-my-source-code-be-in-utf-8)
- [x] Ensuring files don't contain windows-style line endings #1480 
Should be automatically handled by .gitattributes but can't hurt to check and enforce.
- [x] Running clang-format #1496 
- [x] Removing trailing white spaces #1477 
Not a big problem but look bad IMO [ref](https://softwareengineering.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal)
- [x] Ensuring files end with a new line #1477 
Not required but is kind of the convention with C/C++.

To address one problem immediately. Mass formatting commits tend to mess up the `git blame`. But fortunately for this very problem the `--ignore-rev` option was created which allows you to ignore certain commits and even load these from a file.

And GitHub supports this commit ignoring out of the box see [this](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).

So all of these mass formatting commits would be added to the `.git-blame-ignore-revs` file and should not show up in the git blame. For local use you might have to set it up manually like this.

```sh
git config --local blame.ignoreRevsFile ".git-blame-ignore-revs"
```

What are the community thoughts on this?

I have implemented all of these options in past so it shouldn't be a big problem.

Xottab-DUTY:
Agree on everything, except details about clang-format'ing the code:
We had a discussion in #278 about the code style modification, and I also suggested to change code style for the entire code base to `snake_case`. The team didn't like the latter, and first kinda did end in nothing.

So, now, we have a very permitting mixed code style: since the original code already has mixed `snake_case` and `PascalCase` and Hungarian notation, we allow every developer to do the same. Generally, sticking to the code style of the part of the code, or the code style of the file you are working on. But developer is also allowed to stick to the his personal preference. It just should not like an alien to the existing X-Ray code base =)
Because of that, clang-format config should be modified to allow everything above and only enforce few things:
[...]

PR #1477, opened 2023-10-26
Fix and enforce code style regarding spaces

AMS21:
As discussed in #1474 this is the first part regarding spaces.

We now enforce:
1. No tabs only spaces except in C++ source code
2. No trailing spaces
3. Files must end with a newline

These rules are now check with the help of the new `stylecheck` github workflow.

Currently we check everything that ends in one of the following file extensions.
- `.h`
- `.hpp`
- `.inl`
- `.c`
- `.cpp`
- `.cs`
- `.rc`
- `.yaml`
- `.yml`
- `.md`
- `.txt`
- `.cmake`
- `.sh`

While the following path are ignored
- `External/`
- `src/_CppDependOut/`
- `sdk`
- `src/utils/mp_gpprof_server/libraries`
- `res`

Here an example of what the output of the workflow run would look like if you added code which violates these checks.

Sample


Xottab-DUTY:
Aaaaahhhh, the diff is too big.

And now I can remember the detail that I didn't remember right away in #1474:
We ignore tabs in comments, allowing them. This is done mostly to reduce the diff, but it also sometimes helps to detect if the comment lays there since the original engine, or it's a new comment.
And this information is actually useful in several cases, when we are reading the code.

Also, these folders should be ignored and treated as External, as they contain third-party code:
1. `/src/utils/mp_gpprof_server/libraries`
2. `sdk`

Also, `res` should be ignored too – it contains modified game assets and we want to maximally reduce the diff with the original game files which use tabs.

AMS21:
Okay that complicates things a bit. There probably is a one line shell command to only replace tabs with spaces which are not inside a comment but I'm not sure I can write one that easy.

So I guess it would be best for now to ignore the entire tabs/spaces stuff with our source code (.cpp/.h etc.) and maybe deal with that later if at all I guess.

Xottab-DUTY:
> So I guess it would be best for now to ignore the entire tabs/spaces stuff with our source code (.cpp/.h etc.) and maybe deal with that later if at all I guess.

And that makes it actually easier at this point!

Xottab-DUTY (approved):

Xottab-DUTY:
/joke Ah, after all these years we now have a proper license.

Xottab-DUTY:
Thanks!

AMS21:
My pleasure!

clayne:
Honestly these types of PRs that make huge sweeping whitespace changes:

1. Really throw off annotation/blame history making it a hassle to trace things back unless the user knows what they're doing.
2. Are mostly unnecessary unless it's to impose consistency one way or the other - but they have the risk of injecting opinions into how it should be (obviously this is discussed before hand).
3. To inject my own opinion on the whole tabs vs spaces thing, tabs are superior because tabs are meta character for indentation whereas only repetitive spaces represent that. Tabs for indentation, spaces for alignment. Editors can be configured to display as needed on the end user side.

Hrusteckiy:
@clayne get out of here, stalker.

AMS21:
> 1. Really throw off annotation/blame history making it a hassle to trace things back unless the user knows what they're doing.

You are correct and that is the exact reason this commit also adds the `.git-blame-ignore-revs` file which GitHub [supports](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).
While for local use you might have to use `git config --local blame.ignoreRevsFile ".git-blame-ignore-revs"`.

> 2. Are mostly unnecessary unless it's to impose consistency one way or the other - but they have the risk of injecting opinions into how it should be (obviously this is discussed before hand).

Yes, the whole point was to enforce consistency and of course it would have been equally viable to enforce tabs instead of spaces. But it is kinda is the standard nowadays and spaces have the advantage to make code look the same everywhere regardless of settings or editor.

> 3. To inject my own opinion on the whole tabs vs spaces thing, tabs are superior because tabs are meta character for indentation whereas only repetitive spaces represent that. Tabs for indentation, spaces for alignment. Editors can be configured to display as needed on the end user side.

Please note that as discussed above the spaces instead of tabs rule is __not__ enforced for the actual source code (.h/.cpp). If you take a look at the actual [commit](https://github.com/OpenXRay/xray-16/pull/1477/commits/3970dd95199e99fecc50e55868019cd08f21925e) you will that basically only .txt files were affected by this change.
Yes editors can be configured as needed by that's were the inconsistencies arise. Someone with 4 space tabs and one with 8 space tabs will see the same file a bit different.

PR #1479, opened 2023-10-26
Refactor `script_game_object` money functions to use u32

### LINKED ISSUE #1173: Negative money balance after paid service

AMS21:
Since `CInventoryOwner` already stored `money` as a `u32` it only makes since for us to also treat it like one.

Also fixed a logic bug which would have allowed the object to transfer money which it didn't have.

Since both values are unsigned subtracting them from another will never
produce a negative value.

This bug was also present before changing the function signature since the `money`
variable was implicitly converted to an `u32` resulting in the same problem.

### DIFF
--- a/src/xrGame/script_game_object.h
+++ b/src/xrGame/script_game_object.h
@@ -369,8 +369,8 @@ public:
     void DropItemAndTeleport(CScriptGameObject* pItem, Fvector position);
     void ForEachInventoryItems(const luabind::functor<void>& functor);
     void TransferItem(CScriptGameObject* pItem, CScriptGameObject* pForWho);
-    void TransferMoney(int money, CScriptGameObject* pForWho);
-    void GiveMoney(int money);
+    void TransferMoney(u32 money, CScriptGameObject* pForWho);
+    void GiveMoney(u32 money);
     u32 Money();
     void MakeItemActive(CScriptGameObject* pItem);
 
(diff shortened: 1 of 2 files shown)


Xottab-DUTY:
So, does that fix #1173 by the way?

letnull19A:
wow

AMS21:
> So, does that fix #1173 by the way?

Well not really. You don't get negative money. The money transfer just does not happen but the dialog still works and you get the information just without losing money.

It does produce a LUA error now so the check definitely works:
```
! [LUA]  Character does not have enought money
stack traceback:
        ...K.E.R. - Call of Pripyat\gamedata\scripts\dialogs.script:496: in function 'relocate_money_from_actor'
        ... - Call of Pripyat\gamedata\scripts\dialogs_zaton.script:134: in function <... - Call of Pripyat\gamedata\scripts\dialogs_zaton.script:133>
```

AMS21:
Any objections to getting this merged? :)

PR #1484, opened 2023-10-27
Fix several cases of uninitialized values that were used and caused undefined behavior

AMS21:


src/xrServerEntities/xrServer_Objects_ALife.cpp
```diff
@@ -838,7 +838,7 @@ void CSE_ALifeLevelChanger::FillProps(LPCSTR pref, PropItemVec& items)
 // CSE_ALifeObjectPhysic
 ////////////////////////////////////////////////////////////////////////////
 CSE_ALifeObjectPhysic::CSE_ALifeObjectPhysic(LPCSTR caSection)
-    : CSE_ALifeDynamicObjectVisual(caSection), CSE_PHSkeleton(caSection)
+    : CSE_ALifeDynamicObjectVisual(caSection), CSE_PHSkeleton(caSection), freezed{false}
```
Xottab-DUTY: We usually use parentheses in the initializer list, so using braces would look like an alien :)
```suggestion
    : CSE_ALifeDynamicObjectVisual(caSection), CSE_PHSkeleton(caSection), freezed(false)
```

src/xrGame/stalker_animation_manager.h
```diff
@@ -69,8 +69,8 @@ class CStalkerAnimationManager
     CMissile* m_missile;
 
 private:
-    bool m_call_script_callback;
-    bool m_call_global_callback;
+    bool m_call_script_callback{false};
+    bool m_call_global_callback{false};
```
Xottab-DUTY: For braces, we always insert spaces inside them.
```suggestion
    bool m_call_script_callback{ false };
    bool m_call_global_callback{ false };
```

Xottab-DUTY (changes requested):
Some minor codestyle changes needed.

AMS21:
There should no more aliens be around :)

Xottab-DUTY (approved):

src/xrGame/obstacles_query_inline.h
```diff
@@ -10,7 +10,13 @@
 #define OBSTACLES_QUERY_INLINE_H
 
 IC obstacles_query::obstacles_query() { init(); }
-IC void obstacles_query::init() { m_crc = 0; }
+
+IC void obstacles_query::init()
+{
+    m_crc = 0;
+    m_actual = true;
```
Xottab-DUTY, line 17: Comment for future, if someone would investigate something:
`m_actual` wasn't set to any value and setting it to `true` here is correct – `m_obstacles` is empty at this point, so there's no need to call `compute_area()` function which `m_actual` is only used for.

PR #1485, opened 2023-10-27
xrGame: Early return in `obstacles_query::set_intersection` if obstacles is empty

AMS21:
Also fixes undefined behavior since in that case we would reference a nullptr.

### DIFF
--- a/src/xrGame/obstacles_query.cpp
+++ b/src/xrGame/obstacles_query.cpp
@@ -15,6 +15,9 @@
 
 void obstacles_query::set_intersection(const obstacles_query& query)
 {
+    if (m_obstacles.empty())
+        return;
+
     // XXX: probably replace xr_alloca
     const u32 n = m_obstacles.size();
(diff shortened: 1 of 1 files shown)


Xottab-DUTY:
Currently not sure if we are doing everything correctly here, I need a bit of research on how the function works.
(e.g. do we need to set m_actual to false? or maybe some other action?)

/offtopic Oh, that `&*ptr` pattern is so frequent!

AMS21:
I don't think we need to set `m_actual` to `false`.

If `m_obstacles` is empty then `n` is assigned to `0`
And `buffer_size` is assigned `n * sizeof(OBSTACKES::value_type)` which is would also be `0`
So the if check for `obstacles().size() == n` results in `0 == 0` which is `true` and we return and don't set `m_actual` to `false`

But we might still need to do something else though.

/offtopic It's a bit weird that we're using `m_obstacles` and `obstacles()` in the same function but they are the same.

PR #1486, opened 2023-10-27
Fix `Wunused-variable` warnings (#713)

### LINKED ISSUE #713: Compiler warnings elimination
Starting from the creation of this task, #116 is a subtask of this task.

MSVC:
- [ ] MSBuild
  - [ ] Enable "warnings are errors" mode *(is it possible?)*
- [ ] Compiler
  - [x] Enable conformance mode
  - [x] Enable "warnings are errors" mode
  - [ ] Raise the warning level to 4
    - [x] Level 1
    - [x] Level 2
    - [ ] Level 3
    - [ ] Level 4
- [ ] Linker
  - [ ] Enable "warnings are errors" mode

GCC:
- CMake and `make`
  - [ ] Enable "warnings are errors" mode *(is it possible?)*
- [ ] Compiler
  - [x] Disable permissive mode
  - [ ] *Maybe enable [pedantic mode](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)*
  - [ ] Enable "warnings are errors" mode
  - [ ] Raise the warning level to 4
    - [ ] Level 1
    - [ ] Level 2
    - [ ] Level 3
    - [ ] Level 4
- [ ] Linker
  - [ ] Enable "warnings are errors" mode

For history:
- [ ] List all warnings elimination commits in this issue

Tools:
PVS-Studio
CppDepend (includes cppcheck, Clang-Tidy, Vera++)
ReSharper C++ (includes Clang and Clang-Tidy check plus it's own analysis)
Visual Assist (has it's own code inspection)

https://habr.com/ru/articles/75123/
https://habr.com/ru/companies/pvs-studio/articles/303664/

AMS21:
Mostly resorted to commenting out variables in case they might still be useful to someone someway.

What confused me is the bit in `IPureClient::Disconnect()`. This almost looks like we're missing some cleanup here.

### DIFF
--- a/src/utils/xrQSlim/MxQSlim.cpp
+++ b/src/utils/xrQSlim/MxQSlim.cpp
@@ -298,7 +298,6 @@ void MxEdgeQSlim::apply_mesh_penalties(MxQSlimEdge* info)
     if (nfailed)
         bias += nfailed * meshing_penalty;
 
-    static u32 a = 0;
     //	if (a)
     {
         double Nmin1 = check_local_inversion(info->v1, info->v2, info->vnew);
@@ -481,7 +480,7 @@ void MxEdgeQSlim::update_pre_contract(const MxPairContraction& conx)
         if (u == v1 || varray_find(star, u))
         {
             // This is a useless link --- kill it
-            bool found = varray_find(edge_links(u), e, &j);
+            [[maybe_unused]] bool found = varray_find(edge_links(u), e, &j);
             VERIFY(found);
             edge_links(u).remove(j);
             heap.remove(e);
--- a/src/utils/xrQSlim/MxStdModel.cpp
+++ b/src/utils/xrQSlim/MxStdModel.cpp
@@ -48,8 +48,7 @@ MxVertexID MxStdModel::alloc_vertex(float x, float y, float z)
     vertex_mark_valid(id);
 
     face_links.add(xr_new<MxFaceList>());
-    unsigned int l = face_links.last_id();
-    VERIFY(l == id);
+    VERIFY(face_links.last_id() == id);
     VERIFY(neighbors(id).length() == 0);
 
     return id;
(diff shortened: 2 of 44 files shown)


Xottab-DUTY:
Uh-oh, this conflicts with #1425. @Masterkatze is working on warnings task, and a CMake refactoring.
So, I'm okay if you both will work on that, but we need to synchronize the work, so that we don't spend time on duplicating changes :)

AMS21:
Oh very good point. Thanks for pointing that out :)

Xottab-DUTY:
> What confused me is the bit in `IPureClient::Disconnect()`. This almost looks like we're missing some cleanup here.

From what I heard, there's a ton of stack overflows and other bugs.
From what I know, they are actually USED to turn multiplayer servers down, exploited for cheats, etc.

(only on Windows, Linux builds don't support multiplayer at all currently)

AMS21:
Removed all changes to files inside `src/Layers` as to not conflict with #1425

AMS21:
> > What confused me is the bit in `IPureClient::Disconnect()`. This almost looks like we're missing some cleanup here.
> 
> From what I heard, there's a ton of stack overflows and other bugs. From what I know, they are actually USED to turn multiplayer servers down, exploited for cheats, etc.
> 
> (only on Windows, Linux builds don't support multiplayer at all currently)

Oh no but I don't think fixing these problems is a high priority.

src/utils/xrQSlim/MxQSlim.cpp
```diff
@@ -481,8 +480,7 @@ void MxEdgeQSlim::update_pre_contract(const MxPairContraction& conx)
         if (u == v1 || varray_find(star, u))
         {
             // This is a useless link --- kill it
-            bool found = varray_find(edge_links(u), e, &j);
-            VERIFY(found);
+            VERIFY(varray_find(edge_links(u), e, &j));
```
Xottab-DUTY: `varray_find` is used to initialize `j` variable, and placing the function call into `VERIFY` breaks the code on Release configurations, because VERIFY macro expands to nothing in Release, so, `j` is being left uninitialized.
MSVC even throws warning about that.
```suggestion
            [[maybe_unused]] bool found = varray_find(edge_links(u), e, &j);
            VERIFY(found);
```

AMS21: Oh very good catch

Xottab-DUTY (approved):

PR #1489, opened 2023-10-31
Fixed comparison class size with pointer size (not data pointer size)

GermanAizek:
hi
Stars in sky have converged, I'm giving you this gift.

tumblr_njqilaXukL1syaimlo1_500


Xottab-DUTY:
Thanks! But the check is correct, it is intended to get the size of `this`, instead of `*this*`.

PR #1491, opened 2023-11-03
xrPhysics: Fix potential copy-paste error in `SetLimitsActive` for axis `-1`

AMS21:
Fair warning I'm not very familiar with Physics programming. But looking at these they seem like classical copy and paste errors.

For example assigning `m_Joint1` `dParamLoStop` to `axis[0].low` twice but never assign `dParamHiStop` to `axis[0].high` like other code paths do.
Also there are 2 cases where due to a missing `break` the same values get assigned again.

But please take a close look at this and correct me if I'm wrong :)


Xottab-DUTY:
@v2v3v4 what do you think?

v2v3v4:
Seems to be a fix, but the only thing is that joints of this type(full_control) with a -1(all axes) limit do not apply.
This is a fixed ball with the ability to rotate on all axes without limitation.

PR #1492, opened 2023-11-03
Let the compiler assume that `VERIFY` expressions are true in release builds

AMS21:
Should allow the compiler to generate better code.

Small caparison of file sizes with gcc-13.2 with and without this patch.
Note: That file size is not a really good metric for comparison when it comes to performance. This rather shows that the patch actually **does** make some difference.

| File              | Before   | After    | diff   | Diff %              |
| ----------------- | -------- | -------- | ------ | ------------------- |
| xrAICore.so       | 558944   | 559040   | -96    | -0.017175244747238% |
| xrAPI.so          | 7664     | 7664     | 0      | 0%                  |
| xrCDB.so          | 287112   | 283136   | -3976  | -1.384825433977%    |
| xrCore.so         | 684664   | 684656   | -8     | -0.001168456352313% |
| xrEngine.so       | 2312128  | 2308000  | -4128  | -0.17853682841088%  |
| xrGame.so         | 44601216 | 44562992 | -38224 | -0.085701699254119% |
| xrLCUtil.so       | 13392    | 13392    | 0      | 0%                  |
| xrLuabind.so      | 173480   | 173480   | 0      | 0%                  |
| xrLuajit.so       | 620768   | 620768   | 0      | 0%                  |
| xrNetServer.so    | 103936   | 103936   | 0      | 0%                  |
| xrParticles.so    | 204440   | 204440   | 0      | 0%                  |
| xrRender_GL.so    | 3700032  | 3707784  | +7752  | +0.20951170152042%  |
| xrScriptEngine.so | 511144   | 511264   | +120   | +0.023476750191727% |
| xrSound.so        | 202240   | 198224   | -4016  | -1.9857594936709%   |
| xrUICore.so       | 2412280  | 2416448  | +4168  | +0.17278259571857%  |
| xr_3da            | 13736    | 13736    | 0      | 0%                  |
| Total             | 56407176 | 56368960 | -38216 | -0.067750245110658% |

### DIFF
--- a/src/xrAICore/Navigation/PatrolPath/patrol_point_inline.h
+++ b/src/xrAICore/Navigation/PatrolPath/patrol_point_inline.h
@@ -16,27 +16,33 @@ inline bool CPatrolPoint::operator==(const CPatrolPoint& rhs) const
 
 IC const Fvector& CPatrolPoint::position() const
 {
+#ifdef DEBUG
     VERIFY(m_initialized);
+#endif
     return (m_position);
 }
 
 IC const u32& CPatrolPoint::flags() const
 {
+#ifdef DEBUG
     VERIFY(m_initialized);
+#endif
     return (m_flags);
 }
 
 IC const shared_str& CPatrolPoint::name() const
 {
+#ifdef DEBUG
     VERIFY(m_initialized);
+#endif
     return (m_name);
 }
 
 IC const u32& CPatrolPoint::level_vertex_id(
     const CLevelGraph* level_graph, const CGameLevelCrossTable* cross, const CGameGraph* game_graph) const
 {
-    VERIFY(m_initialized);
 #ifdef DEBUG
+    VERIFY(m_initialized);
     verify_vertex_id(level_graph, cross, game_graph);
 #endif
     return (m_level_vertex_id);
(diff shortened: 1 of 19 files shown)


Xottab-DUTY:
Oh no, adding `#ifdef DEBUG` to a `VERIFY` which is already `DEBUG` only looks really ugly

AMS21:
Hm, could add something like `VERIFY_DBG()` for these cases which would work exactly like the old one.

Would that be a better solution here?

Xottab-DUTY:
> Oh no, adding `#ifdef DEBUG` to a `VERIFY` which is already `DEBUG` only looks really ugly

Hmmm, why we even do that? Those VERIFYs are failing to compile?

AMS21:
> > Oh no, adding `#ifdef DEBUG` to a `VERIFY` which is already `DEBUG` only looks really ugly
> 
> Hmmm, why we even do that? Those VERIFYs are failing to compile?

Exactly these are VERIFYs on function or member that are inside of `#ifdef DEBUG` which was fine before because if `DEBUG` is not defined when `VERIFY` expands to nothing but if it expands to `XR_ASSUME(x)` and `x` is only defined with `DEBUG` then we get a compile error.

Xottab-DUTY:
Aaahhh!

Nevermind then, I thought they were failing to compile for another reason.
Sorry 😂

AMS21:
> Aaahhh!
> 
> Nevermind then, I thought they were failing to compile for another reason. Sorry 😂

No worries all good :)

Xottab-DUTY:
I'm not sure, but this could increase the build speed.

AMS21:
> I'm not sure, but this could increase the build speed.

Do you mean build time, i.e. the time it takes to build the project.
In that case yes, as the compiler has to evaluate more and has more information and work to do when optimizing.

PR #1510, opened 2023-11-09
Fix definitions of `type_min` and `type_zero` in `xr_types.h`

### LINKED ISSUE #195: Bug in _types.h
https://github.com/OpenXRay/xray-16/blob/70f130c6fc719293c3b6abb1ccacbb084eaee6d1/src/xrCore/_types.h#L31
Is there problem with determination min values here ?
For example -std::numeric_limits<int>::max() != std::numeric_limits<int>::min() also for signed types.
http://en.cppreference.com/w/cpp/types/numeric_limits
Also can replace type_zero to std::numeric_limits<int>::lowest() ?

AMS21:
While looking at the code inside `xr_types.h` I've notices some very strange definitions which are used later in the code to define constants.

```cpp
// Type limits
template <typename T>
constexpr auto type_max = std::numeric_limits<T>::max();

template <typename T>
constexpr auto type_min = -std::numeric_limits<T>::max();

template <typename T>
constexpr auto type_zero = std::numeric_limits<T>::min();
```

`type_min` is not the minimum value of integer types it is actually the minimum + 1. This only works for floating point types.
`type_zero` is the actual minimum of a type. But is only 0 for unsigned types. For signed types its a negative value.

Thus we have some very strange constants.

```cpp
constexpr int int_min = type_min<int>;         // actually -2147483647 not -2147483648
constexpr int int_zero = type_zero<int>;       // actually -2147483648 not 0
...
constexpr float flt_zero = type_zero<float>;   // actually 1.175494e-38 not 0.0
...
constexpr double dbl_zero = type_zero<double>; // actually 2.225074e-308 not 0.0
```

Fixes #195

### DIFF
--- a/src/Layers/xrRenderPC_R1/LightProjector.cpp
+++ b/src/Layers/xrRenderPC_R1/LightProjector.cpp
@@ -243,7 +243,7 @@ void CLightProjector::calculate()
         v.sub(v_Cs, v_C);
         ;
 #ifdef DEBUG
-        if ((v.x * v.x + v.y * v.y + v.z * v.z) <= flt_zero)
+        if ((v.x * v.x + v.y * v.y + v.z * v.z) <= flt_min)
         {
             IGameObject* OO = dynamic_cast<IGameObject*>(R.O);
             Msg("Object[%s] Visual[%s] has invalid position. ", *OO->cName(), *OO->cNameVisual());
@@ -274,7 +274,7 @@ void CLightProjector::calculate()
         }
 #endif
         // handle invalid object-bug
-        if ((v.x * v.x + v.y * v.y + v.z * v.z) <= flt_zero)
+        if ((v.x * v.x + v.y * v.y + v.z * v.z) <= flt_min)
         {
             // invalidate record, so that object will be unshadowed, but doesn't crash
             R.dwTimeValid = Device.dwTimeGlobal;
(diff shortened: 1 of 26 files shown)


Xottab-DUTY:
This is related to #195.

One of the reasons why this inconsistency still wasn't fixed is that I really don't know if everything will work correct after the change.

Knowing the codebase, some code logic could depend on this inconsistency. The situation when some code was broken because another unrelated code has been changed happened many times

AMS21:
Okay I've now added type_lowest and accompanying constants and changed all usages of these constants across the code base to keep the old behavior.

Making these changes I'm a bit suspicious about some of the uses of `flt_zero` now `flt_min`. Some of them look to me like they actually meant to check for 0.

src/Layers/xrRenderPC_R1/LightProjector.cpp
```diff
@@ -243,7 +243,7 @@ void CLightProjector::calculate()
         v.sub(v_Cs, v_C);
         ;
 #ifdef DEBUG
-        if ((v.x * v.x + v.y * v.y + v.z * v.z) <= flt_zero)
+        if ((v.x * v.x + v.y * v.y + v.z * v.z) <= flt_min)
```
AMS21, line 246: Can this ever be true?

src/utils/xrMiscMath/matrix.cpp
```diff
@@ -155,7 +155,7 @@ _matrix<T>& _matrix<T>::invert(const _matrix<T>& a)   // important: this is 4x3
 		a._12 * (a._21 * a._33 - a._23 * a._31) +
 		a._13 * (a._21 * a._32 - a._22 * a._31));
 
-	VERIFY(_abs(fDetInv) > flt_zero);
+	VERIFY(_abs(fDetInv) > flt_min);
```
AMS21, line 158: Similar here this would always be true since `flt_min` is negative

PR #1515, opened 2023-11-11
Increase robustness of `XMLDocument` parser

AMS21:
- Initialize `m_bIgnoreMissingEndTagError`
- Pass down `bool fatal` to `ParseFile`. Previously we would still hit assertions for invalid `#includes` with `fatal` set to `false`
- Fix potential use of `nullptr` when `strstr` returns `nullptr`
- Introduce a non failing variant of `r_string` called `try_r_string`
- Fix various problems with the handling of `#include`, remove not needed string copy and give feedback on invalid things instead of silently ignoring problems.
- Fix an infinite loop in tinyxml when parsing an invalid UTF-8 sequence
- Fix infinite recursion or cyclic includes causing a stack overflow

### DIFF
--- a/src/xrCore/XML/XMLDocument.cpp
+++ b/src/xrCore/XML/XMLDocument.cpp
@@ -5,15 +5,76 @@
 pcstr UI_PATH = UI_PATH_DEFAULT;
 pcstr UI_PATH_WITH_DELIMITER = UI_PATH_DEFAULT_WITH_DELIMITER;
 
-XMLDocument::XMLDocument() : m_xml_file_name(), m_root(nullptr), m_pLocalRoot(nullptr) {}
+XMLDocument::XMLDocument() : m_xml_file_name(), m_root(nullptr), m_pLocalRoot(nullptr), m_bIgnoreMissingEndTagError(false) {}
 
 XMLDocument::~XMLDocument() { ClearInternal(); }
 
 void XMLDocument::ClearInternal() { m_Doc.Clear(); }
 
-void ParseFile(pcstr path, CMemoryWriter& W, IReader* F, XMLDocument* xml)
+enum class ParseIncludeResult
 {
-    string4096 str;
+    Success,   /// There is a valid #include and 'out_include_name' returns the filename
+    Error,     /// There is a #include but there is some problem
+    NoInclude, /// There is no #include on this line
+};
+
+// Given a string of the form: '#include "filename"' we return the filename in 'out_include_name'
+ParseIncludeResult ParseInclude(pstr string, pcstr& out_include_name)
+{
+    // Skip any whitespace characters
+    while (*string != '\0' && std::isblank(*string))
+    {
+        ++string;
+    }
+
+    // Check for #include
+    static constexpr pcstr IncludeTag = "#include";
+    if (std::strncmp(string, IncludeTag, 8) != 0)
+        return ParseIncludeResult::NoInclude;
+
+    string += 8;
+
+    // Skip any whitespace characters
+    while (*string != '\0' && std::isblank(*string))
+        ++string;
+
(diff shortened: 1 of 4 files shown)


src/xrCore/XML/XMLDocument.cpp
```diff
@@ -30,17 +86,18 @@ void ParseFile(pcstr path, CMemoryWriter& W, IReader* F, XMLDocument* xml)
...
-
-        // Skip any spaces or tabs
-        pcstr begin_of_include = str;
-        while (*begin_of_include != '\0' && std::isblank(*begin_of_include))
-            ++begin_of_include;
+        string4096 str;
+        if (!F->try_r_string(str, sizeof(str)))
+        {
+            R_ASSERT3(!fatal, "XML file[%s] parsing failed. Line is too long (>= 4096)", path);
+            Msg("! XML file[%s] parsing failed. Line is too long (>= 4096)", path);
```
Xottab-DUTY, line 101: ~~This is not needed, because R_ASSERT message is already being included in the log~~

Update: oh no, it's actually needed!
Nevermind.

src/xrCore/XML/XMLDocument.cpp
```diff
@@ -5,23 +5,79 @@
...
+enum class ParseIncludeResult
 {
-    string4096 str;
+    Success,   /// There is a valid #include and 'out_include_name' returns the filename
+    Error,     /// There is a #include but there is some problem
+    NoInclude, /// We #include on this line
+};
 
+// Give a string of the form: '#include "filename"' we return the filename in 'out_include_name'
+ParseIncludeResult ParseInclude(pstr string, pcstr* out_include_name)
```
Xottab-DUTY: Why do we use pointer here?
```suggestion
ParseIncludeResult ParseInclude(pstr string, pcstr& out_include_name)
```

src/xrCore/XML/XMLDocument.cpp
```diff
@@ -5,23 +5,79 @@
...
 
 void XMLDocument::ClearInternal() { m_Doc.Clear(); }
 
-void ParseFile(pcstr path, CMemoryWriter& W, IReader* F, XMLDocument* xml)
+enum class ParseIncludeResult
 {
-    string4096 str;
+    Success,   /// There is a valid #include and 'out_include_name' returns the filename
+    Error,     /// There is a #include but there is some problem
+    NoInclude, /// We #include on this line
```
Xottab-DUTY: Did you mean this?
```suggestion
    NoInclude, /// We don't have #include on this line
```

AMS21: Yes I did :)

src/xrCore/XML/XMLDocument.cpp
AMS21: Yes good catch using a reference here is much cleaner

AMS21:
Implemented suggested changes

AMS21:
Left the fuzzer I wrote running last night and it found another problem with a file including itself causing a stack overflow

Xottab-DUTY (approved):

Xottab-DUTY:
It turned out that we must parse unterminated includes, because they exist in Clear Sky.
Look at the map_spots.xml:
```xml
<map_spots>

#include "ui\map_spots_mp.xml"
#include "ui\map_spots_relations.xml"
#include "ui\map_spots_complex.xml"
#include "ui\map_spots_find_upgrade.xml

... the rest of the file
```

The last map_spots_find_upgrade.xml include doesn't have the ending quote. But original code parses that correctly.

AMS21:
Hm, shouldn't be too hard to fix. I'll guess treating the quotes the same as any white spaces should fix that.

Will take a look :)

PR #1520, opened 2023-11-14
GitHub Actions: Allow dependabot to update GitHub actions

AMS21:
Dependabot is also able to automatically update the GitHub actions we use so why not :^)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,10 @@ updates:
   open-pull-requests-limit: 10
   labels:
   - External (3rd party)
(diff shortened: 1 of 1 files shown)


Xottab-DUTY:
For some reason I didn't knew about that. Hm, maybe this wasn't available when I was configuring dependabot.

AMS21:
> For some reason I didn't knew about that. Hm, maybe this wasn't available when I was configuring dependabot.

Yeah I also remember dependabot not support this, but no idea when they added that.

PR #1523, opened 2023-11-14
Generalize bug fix from #1512

AMS21:
This is a more generalized solution to the bug fix described in #1512 which requires less `#ifdef` and allows other places to also benefit from this fix.

### DIFF
--- a/src/Common/PlatformLinux.inl
+++ b/src/Common/PlatformLinux.inl
@@ -429,7 +429,7 @@ inline int _rmdir(const char *path)
 }
 #define _write write
 #define _strupr strupr
-#define _read read
+#define _read xr_read
 #define _set_new_handler std::set_new_handler
 #define _finite isfinite
 inline int _mkdir(const char *dir) { return mkdir(dir, S_IRWXU); }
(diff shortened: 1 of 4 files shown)


src/Common/PlatformLinux.inl
```diff
@@ -1108,3 +1108,29 @@ inline tm* localtime_safe(const time_t *time, struct tm* result){ return localti
 #define xr_strerror(errno, buffer, bufferSize) strerror_r(errno, buffer, sizeof(buffer))
 
 using xrpid_t = pid_t;
+
+// This is a drop-in replacement for calls to linux 'read' function. We use this because unlike other OSes
+// Linux 'read' function can return less then the requested number of bytes and might need to be called multiple times.
+// Apart from this, it behaves exactly as you would expect and matches the other OSes implementation.
+// See also: https://www.man7.org/linux/man-pages/man2/read.2.html
+inline ssize_t xr_read(int file_handle, void *buffer, size_t count)
```
375gnu, line 1116: Actually it make sense to make this implementation common, so any new platform with such behaviour will be handled automatically. And it will not harm other platforms.

AMS21: Alright where would be a good place to put this? Directly into Platform.h?

Xottab-DUTY (approved):

PR #1525, opened 2023-11-14
Fix custom condition factor not saving properly (#1524)

### LINKED ISSUE #1524: Condition factor added with use_condition param in LTX doesn't save properly
**Describe the bug**
There is some code in engine which does work initially, but doesn't save properly to packets:
```
    // Added by Axel, to enable optional condition use on any item
    m_flags.set(FUsingCondition, READ_IF_EXISTS(pSettings, r_bool, section, "use_condition", false));
```

**To Reproduce**
1. Add some log like so in CInventoryItem::Load/Save funcs: `Msg("debug - save item %s with condition %f", m_object->cNameSect().c_str(), m_fCondition)`
1. load into game
2. drop a weapon/item, shoot it
3. pick it up, save game
4. check logs from step 1 - correct values saved
5. load game
6. check logs from step 1 - correct values saved
7. save game
8. check logs from step 1 - observe bug, loaded values are reset to 1.f (unless the item condition has changed between steps 5 and 7; then it will be equal to `1 - conditionDecremented`)

**Expected behavior**
use_condition param should save to packets 

 - OS: Windows 10
 - OpenXRay build version: 98f80503e11d30861d9fe1079138544ac7866680

yohjimane:
~Doesn't work yet because pSE_InventoryItem m_fCondition is not updated correctly, and overrides the correct value.~ That issue is fixed, should be ready to merge now!

Will break saves; requires new save

credits: lucy

### DIFF
--- a/src/xrGame/ActorHelmet.cpp
+++ b/src/xrGame/ActorHelmet.cpp
@@ -74,20 +74,6 @@ bool CHelmet::net_Spawn(CSE_Abstract* DC)
     return (res);
 }
 
-void CHelmet::net_Export(NET_Packet& P)
-{
-    inherited::net_Export(P);
-    P.w_float_q8(GetCondition(), 0.0f, 1.0f);
-}
-
-void CHelmet::net_Import(NET_Packet& P)
-{
-    inherited::net_Import(P);
-    float _cond;
-    P.r_float_q8(_cond, 0.0f, 1.0f);
-    SetCondition(_cond);
-}
-
 void CHelmet::OnH_A_Chield()
 {
     inherited::OnH_A_Chield();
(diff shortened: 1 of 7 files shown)


yohjimane:
Fixed now - there was some code which was overriding the correct condition.

I removed that line, checked vanilla condition on save/load with multiple item types (weapon, armor, helmet, eatable item) - everything works as expected.

I also checked custom condition added with `use_condition` param, and that works as expected too*

*Eatable items also need to have `max_uses` defined to function properly.

Hrusteckiy:
Does this work with elements that only have `IsUsingCondition()` flag?

yohjimane:
@Hrusteckiy Yes, it should work. I tested some outfits out- those use `FUsingCondition` flag instead of `use_condition` from ltx and the condition saves/loads correctly.

PR #1545, opened 2023-11-27
Separate new shaders into r5

yohjimane:
Seems to work okay for vanilla R4 so far. ~There are some R3 bugs~ Nevermind, SSAO is bugged in DX10 before any new shader stuff was merged.


src/Layers/xrRender/xrRender_console.cpp
```diff
@@ -852,6 +852,29 @@ class CCC_Fog_Reload : public IConsole_Command
...
 #endif // (RENDER == R_R3) || (RENDER == R_R4)
 
+class CCC_new_shader_support : public CCC_Mask
+{
+public:
+    CCC_new_shader_support(pcstr N, Flags32* V, u32 M) : CCC_Mask(N, V, M) { ; };
+    void Execute(LPCSTR args) override
+    {
+#if RENDER == R_R4
+        if (HW.FeatureLevel >= D3D_FEATURE_LEVEL_11_0)
```
yohjimane, line 862: @Xottab-DUTY is this the correct way to check if we're using DX11 renderer? Seems to work okay on my end, but i wasn't sure.

Xottab-DUTY: > @Xottab-DUTY is this the correct way to check if we're using DX11 renderer? Seems to work okay on my end, but i wasn't sure.

Yes, it is the correct way!

Xottab-DUTY:
Thanks!

PR #1548, opened 2023-11-28
* FIX: [Engine] add new_shader_support to r2 _options so it persists properly

yohjimane:


src/Layers/xrRender_R2/r2.cpp
```diff
@@ -581,7 +581,9 @@ void CRender::create()
     }
 #endif
 
+    o.new_shader_support = false;
 #if defined(USE_DX11)
+    o.new_shader_support = ps_r2_ls_flags_ext.test(R4FLAGEXT_NEW_SHADER_SUPPORT);
```
Xottab-DUTY: Still should check for feature level here:
```suggestion
    o.new_shader_support = HW.FeatureLevel >= D3D_FEATURE_LEVEL_11_0 && ps_r2_ls_flags_ext.test(R4FLAGEXT_NEW_SHADER_SUPPORT);
```

Xottab-DUTY (approved):

PR #1549, opened 2023-11-28
* FIX: [Engine] Initialize new shader options before ResourceManager loads

yohjimane:
This doesn't cause issues in vanilla OXR, but can cause issues if there are new shaders added in R5 which do not exist in R3.

### DIFF
--- a/src/Layers/xrRender_R2/r2.cpp
+++ b/src/Layers/xrRender_R2/r2.cpp
@@ -209,6 +209,33 @@ static bool must_enable_old_cascades()
     return exist;
 }
 
+void CRender::OnDeviceCreate(pcstr shName)
+{
+    o.new_shader_support = 0;
+    o.ssfx_rain = 0;
+    o.ssfx_blood = 0;
+    o.ssfx_branches = 0;
+    o.ssfx_hud_raindrops = 0;
+
+#if defined(USE_DX11)
+    o.new_shader_support = HW.FeatureLevel >= D3D_FEATURE_LEVEL_11_0 && ps_r2_ls_flags_ext.test(R4FLAGEXT_NEW_SHADER_SUPPORT);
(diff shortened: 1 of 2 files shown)


src/Layers/xrRender_R2/r2.cpp
```diff
@@ -209,6 +209,33 @@ static bool must_enable_old_cascades()
...
+    o.ssfx_hud_raindrops = FS.exist(fn, "$game_shaders$", "r3\\deffer_base_hud_bump", ".ps") ? 1 : 0;
+
+    Msg("- NEW SHADER SUPPORT ENABLED %i", o.new_shader_support);
+    Msg("- SSS HUD RAINDROPS SHADER INSTALLED %i", o.ssfx_hud_raindrops);
+    Msg("- SSS RAIN SHADER INSTALLED %i", o.ssfx_rain);
+    Msg("- SSS BLOOD SHADER INSTALLED %i", o.ssfx_blood);
+    Msg("- SSS BRANCHES SHADER INSTALLED %i", o.ssfx_branches);
+#endif
+
+    D3DXRenderBase::OnDeviceCreate(shName);
```
yohjimane, line 236: Doing `D3DXRenderBase::OnDeviceCreate` AFTER some options are initialized is the fix here.

I'm wondering if we need to also use value of `GetShaderPath()` instead of hardcoding `r3\\` too..

Xottab-DUTY: I think we don't even need these `o.ssfx_` members, because they always should true when new shader support enabled. And we can then check for `o.new_shader_support` in the renderer instead of `o.ssfx_`
(if it's not, I believe we should add these shaders, so it is always true)

YES, this is a deviation from SSS original code, but... it generally looks better in our case since we have changed shaders folder to `r5`

PR #1555, opened 2023-12-06
Added a checking bad phrase structure

ChugunovRoman:
This verify checks for an already existing phrase added with the same ID, which is an incorrect dialog structure. Each new phrase must have a unique identifier. This check is simply more informative than just `m_edges.end() == std::find(m_edges.begin(), m_edges.end(), vertex->vertex_id())`. The following crash will immediately follow:

```
FATAL ERROR
 
[error] Expression    : m_edges.end() == std::find(m_edges.begin(), m_edges.end(), vertex->vertex_id())
[error] Function      : add_edge
[error] File          : /home/ruut/Documents/qemu_shared/xray-16/src/xrAICore/Navigation/graph_vertex_inline.h
[error] Line          : 68
[error] Description   : assertion failed
 

stack trace:

xrDebug::GatherInfo(char*, unsigned long, ErrorLocation const&, char const*, char const*, char const*, char const*)
xrDebug::Fail(bool&, ErrorLocation const&, char const*, char const*, char const*, char const*)
CGraphVertex<CPhrase*, shared_str, CGraphAbstract<CPhrase*, float, shared_str, Loki::EmptyType> >::add_edge(CGraphVertex<CPhrase*, shared_str, CGraphAbstract<CPhrase*, float, shared_str, Loki::EmptyType> >*, float const&)
CGraphAbstract<CPhrase*, float, shared_str, Loki::EmptyType>::add_edge(shared_str const&, shared_str const&, float const&)
CPhraseDialog::AddPhrase(char const*, shared_str const&, shared_str const&, int)
CPhraseDialog::AddPhrase_script(char const*, char const*, char const*, int)
luabind::detail::invoke_struct<luabind::meta::type_list<>, luabind::meta::type_list<CPhrase*, CPhraseDialog&, char const*, char const*, char const*, int>, CPhrase* (CPhraseDialog::*)(char const*, char const*, char const*, int)>::call_struct<true, false, luabind::meta::index_list<0u, 1u, 2u, 3u, 4u> >::call(lua_State*, CPhrase* (CPhraseDialog::*&)(char const*, char const*, char const*, int), std::tuple<luabind::default_converter<CPhraseDialog&, void>, luabind::default_converter<char const*, void>, luabind::default_converter<char const*, void>, luabind::default_converter<char const*, void>, luabind::default_converter<int, void> >&)
luabind::detail::invoke_struct<luabind::meta::type_list<>, luabind::meta::type_list<CPhrase*, CPhraseDialog&, char const*, char const*, char const*, int>, CPhrase* (CPhraseDialog::*)(char const*, char const*, char const*, int)>::invoke(lua_State*, luabind::detail::function_object const&, luabind::detail::invoke_context&, CPhrase* (CPhraseDialog::*&)(char const*, char const*, char const*, int))
luabind::detail::function_object_impl<CPhrase* (CPhraseDialog::*)(char const*, char const*, char const*, int), luabind::meta::type_list<CPhrase*, CPhraseDialog&, char const*, char const*, char const*, int>, luabind::meta::type_list<> >::invoke_defer(lua_State*, luabind::detail::function_object_impl<CPhrase* (CPhraseDialog::*)(char const*, char const*, char const*, int), luabind::meta::type_list<CPhrase*, CPhraseDialog&, char const*, char const*, char const*, int>, luabind::meta::type_list<> >*, luabind::detail::invoke_context&, int&)
luabind::detail::function_object_impl<CPhrase* (CPhraseDialog::*)(char const*, char const*, char const*, int), luabind::meta::type_list<CPhrase*, CPhraseDialog&, char const*, char const*, char const*, int>, luabind::meta::type_list<> >::entry_point(lua_State*)
/home/ruut/Documents/qemu_shared/xray-16/bin/x64/Debug/xrLuaJIT.so(+0xa3d5) [0x7f51526a73d5]
/home/ruut/Documents/qemu_shared/xray-16/bin/x64/Debug/xrLuaJIT.so(lua_pcall+0xae) [0x7f51526bab5e]
luabind::detail::pcall(lua_State*, int, int)
void luabind::detail::call_function_struct<void, luabind::meta::type_list<>, luabind::meta::index_list<1u>, 1u, &luabind::detail::pcall, true>::call<CPhraseDialog*>(lua_State*, CPhraseDialog*&&)
void luabind::call_pushed_function<void, luabind::meta::type_list<>, CPhraseDialog*>(lua_State*, CPhraseDialog*&&)
void luabind::call_function<void, luabind::meta::type_list<>, CPhraseDialog*>(luabind::adl::object const&, CPhraseDialog*&&)
CPhraseDialog::load_shared(char const*)
CSharedClass<SPhraseDialogData, shared_str, false>::load_shared(shared_str, char const*)
CPhraseDialog::Load(shared_str)
CPhraseDialogManager::AddAvailableDialog(shared_str, CPhraseDialogManager*)
```

The new verify:
```
 
FATAL ERROR
 
[error] Expression    : !_vertex
[error] Function      : AddPhrase
[error] File          : /home/ruut/Documents/qemu_shared/xray-16/src/xrGame/PhraseDialog.cpp
[error] Line          : 259
[error] Description   : Dublicate phrase ID: [244] for phrase: [dm_do_not_know_10]. Existed phrase by this ID: [dm_no_more_8]
 

stack trace:

xrDebug::GatherInfo(char*, unsigned long, ErrorLocation const&, char const*, char const*, char const*, char const*)
xrDebug::Fail(bool&, ErrorLocation const&, char const*, char const*, char const*, char const*)
xrDebug::Fail(bool&, ErrorLocation const&, char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, char const*, char const*)
CPhraseDialog::AddPhrase(char const*, shared_str const&, shared_str const&, int)
CPhraseDialog::AddPhrase_script(char const*, char const*, char const*, int)
luabind::detail::invoke_struct<luabind::meta::type_list<>, luabind::meta::type_list<CPhrase*, CPhraseDialog&, char const*, char const*, char const*, int>, CPhrase* (CPhraseDialog::*)(char const*, char const*, char const*, int)>::call_struct<true, false, luabind::meta::index_list<0u, 1u, 2u, 3u, 4u> >::call(lua_State*, CPhrase* (CPhraseDialog::*&)(char const*, char const*, char const*, int), std::tuple<luabind::default_converter<CPhraseDialog&, void>, luabind::default_converter<char const*, void>, luabind::default_converter<char const*, void>, luabind::default_converter<char const*, void>, luabind::default_converter<int, void> >&)
luabind::detail::invoke_struct<luabind::meta::type_list<>, luabind::meta::type_list<CPhrase*, CPhraseDialog&, char const*, char const*, char const*, int>, CPhrase* (CPhraseDialog::*)(char const*, char const*, char const*, int)>::invoke(lua_State*, luabind::detail::function_object const&, luabind::detail::invoke_context&, CPhrase* (CPhraseDialog::*&)(char const*, char const*, char const*, int))
luabind::detail::function_object_impl<CPhrase* (CPhraseDialog::*)(char const*, char const*, char const*, int), luabind::meta::type_list<CPhrase*, CPhraseDialog&, char const*, char const*, char const*, int>, luabind::meta::type_list<> >::invoke_defer(lua_State*, luabind::detail::function_object_impl<CPhrase* (CPhraseDialog::*)(char const*, char const*, char const*, int), luabind::meta::type_list<CPhrase*, CPhraseDialog&, char const*, char const*, char const*, int>, luabind::meta::type_list<> >*, luabind::detail::invoke_context&, int&)
luabind::detail::function_object_impl<CPhrase* (CPhraseDialog::*)(char const*, char const*, char const*, int), luabind::meta::type_list<CPhrase*, CPhraseDialog&, char const*, char const*, char const*, int>, luabind::meta::type_list<> >::entry_point(lua_State*)
/home/ruut/Documents/qemu_shared/xray-16/bin/x64/Debug/xrLuaJIT.so(+0xa3d5) [0x7f51526a73d5]
/home/ruut/Documents/qemu_shared/xray-16/bin/x64/Debug/xrLuaJIT.so(lua_pcall+0xae) [0x7f51526bab5e]
luabind::detail::pcall(lua_State*, int, int)
void luabind::detail::call_function_struct<void, luabind::meta::type_list<>, luabind::meta::index_list<1u>, 1u, &luabind::detail::pcall, true>::call<CPhraseDialog*>(lua_State*, CPhraseDialog*&&)
void luabind::call_pushed_function<void, luabind::meta::type_list<>, CPhraseDialog*>(lua_State*, CPhraseDialog*&&)
void luabind::call_function<void, luabind::meta::type_list<>, CPhraseDialog*>(luabind::adl::object const&, CPhraseDialog*&&)
CPhraseDialog::load_shared(char const*)
CSharedClass<SPhraseDialogData, shared_str, false>::load_shared(shared_str, char const*)
CPhraseDialog::Load(shared_str)
CPhraseDialogManager::AddAvailableDialog(shared_str, CPhraseDialogManager*)
CActor::UpdateAvailableDialogs(CPhraseDialogManager*)
```

### DIFF
--- a/src/xrGame/PhraseDialog.cpp
+++ b/src/xrGame/PhraseDialog.cpp
@@ -254,8 +254,7 @@ CPhrase* CPhraseDialog::AddPhrase(
 {
     CPhrase* phrase = NULL;
     CPhraseGraph::CVertex* _vertex = data()->m_PhraseGraph.vertex(phrase_id);
-
-    VERIFY2(!_vertex, make_string("Dublicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()));
+    VERIFY2(!_vertex, make_string("Duplicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()));
 
     if (!_vertex)
     {
--- a/src/xrGame/PhraseDialog.cpp
+++ b/src/xrGame/PhraseDialog.cpp
@@ -254,6 +254,9 @@ CPhrase* CPhraseDialog::AddPhrase(
 {
     CPhrase* phrase = NULL;
     CPhraseGraph::CVertex* _vertex = data()->m_PhraseGraph.vertex(phrase_id);
+
+    VERIFY2(!_vertex, make_string("Dublicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()));
+
     if (!_vertex)
     {
         phrase = xr_new<CPhrase>();
--- a/src/xrGame/PhraseDialog.cpp
+++ b/src/xrGame/PhraseDialog.cpp
@@ -254,7 +254,7 @@ CPhrase* CPhraseDialog::AddPhrase(
 {
     CPhrase* phrase = NULL;
     CPhraseGraph::CVertex* _vertex = data()->m_PhraseGraph.vertex(phrase_id);
-    VERIFY2(!_vertex, make_string("Duplicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()));
+    VERIFY2(!_vertex, make_string("Duplicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()).c_str());
 
     if (!_vertex)
     {


ChugunovRoman:
@Xottab-DUTY But I’m not sure if this check should be removed from the MASTER_GOLD?

ChugunovRoman:
Ah, I forgot that checks are disabled in MASTER_GOLD...

src/xrGame/PhraseDialog.cpp
```diff
@@ -254,6 +254,9 @@ CPhrase* CPhraseDialog::AddPhrase(
 {
     CPhrase* phrase = NULL;
     CPhraseGraph::CVertex* _vertex = data()->m_PhraseGraph.vertex(phrase_id);
+
+    VERIFY2(!_vertex, make_string("Dublicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()));
+
     if (!_vertex)
```
AMS21, line 259: Should this check then also be removed since `_vertex` should essentially always be null?

ChugunovRoman: @AMS21 `_vertex` will not be null only if a new phrase with the same id is added. For example, if a phrase was previously added in the context of one dialog with ID 244, and if another phrase with the same ID 244 is added for the same dialog, then the `_vertex` will not be zero.

ChugunovRoman: This mistake can easily be made when a mod developer creates a complex dialog from scripts.

AMS21: Wait I'm confused if it's valid for `_vertex` to be non-null than the added `VERIFY` is wrong or am I missing something here?

The way I understand we use things like `VERIFY` is to check for something that should not happen because something is seriously wrong. (Having 2 phrases with the same ID sounds like this scenario to me).

Also since #1492 was merged the check would very likely be removed since the compiler is told to assume `_vertex` is null

ChugunovRoman: On [256](https://github.com/OpenXRay/xray-16/pull/1555/files#diff-5a47ae2a8b262938a35d803b14650aa1f50a585a87f2917c138e4c92eae43333R256) line `_vertex` will be null when adding a new phrase by ID. But if a phrase has already been added for the same ID, then the `_vertex` will not be null. This means that a phrase has already been added for this ID. And this behavior is a bug. And this `VERIFY` will appear only in the debug build.
Or are you talking about replacing the `VERIFY` on `R_ASSERT`?

AMS21: Yes I do understand how `_vertex` _can_ be null.

I was talking about the if statement right under that for which I recommended removing it since AFTER the `VERIFY` you added, `_vertex` should never be null and in release mode the compiler will actually assume that and likely remove the check.

But since adding a duplicate id triggers an assertions later I think it's fine to leave it as `VERIFY` although using `R_ASSERT` should also be completely fine.

ChugunovRoman: @AMS21 okay, I updated the check and remove the if statement

Xottab-DUTY:
@ChugunovRoman, please, make it not crash. Prefer robustness and always try not to crash instead of using R_ASSERT.

src/xrGame/PhraseDialog.cpp
Xottab-DUTY: No, R_ASSERT is not fine because we want engine to not crash for players.
Using R_ASSERT_CURE is better.

ChugunovRoman: @Xottab-DUTY what is R_ASSERT_CURE? I haven't seen such a macro

ChugunovRoman:
> @ChugunovRoman, please, make it not crash. Prefer robustness and always try not to crash instead of using R_ASSERT.

I returned the VERIFY

AMS21:
> I returned the VERIFY

I believe what @Xottab-DUTY is trying to say that it would be even better if we didn't crash at all. So instead of an `VERIFY` or `R_ASSERT` we would have something like this:

```cpp
if (_vertex)
{
    Msg(make_string("! Dublicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()));
    return;
}
```

Which would give mod makers and player a better message to understand what exactly the problem is and where but not crash the game.

Although you may need to change to function to return a bool and also add an early return in the calling functions.

ChugunovRoman:
@AMS21 There will be no crash in the release build. Because VERIFY is used. And in the debug assembly needs such crash because, allow you to catch an error in a timely manner and fix it. I would to keep the VERIFY.
@Xottab-DUTY Do you agree with me?

Xottab-DUTY:
> @AMS21 There will be no crash in the release build. Because VERIFY is used. And in the debug assembly needs such crash because, allow you to catch an error in a timely manner and fix it. I would to keep the VERIFY. @Xottab-DUTY Do you agree with me?

But why `if (!_vertex)` check is removed?

ChugunovRoman:
> But why `if (!_vertex)` check is removed?

Now if the game doesn’t crash, it’s probably worth returning this check.

AMS21:
I don't think my point here is really getting across, but it's not too important and might also be a me problem.

Sorry for all the confusion :sweat: 

I do think in essence this is a good addition.

src/xrGame/PhraseDialog.cpp
```diff
@@ -254,6 +254,9 @@ CPhrase* CPhraseDialog::AddPhrase(
 {
     CPhrase* phrase = NULL;
     CPhraseGraph::CVertex* _vertex = data()->m_PhraseGraph.vertex(phrase_id);
+
+    VERIFY2(!_vertex, make_string("Dublicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()));
```
Xottab-DUTY:
```suggestion
    VERIFY2(!_vertex, make_string("Duplicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()));
```

src/xrGame/PhraseDialog.cpp
```diff
@@ -254,6 +254,8 @@ CPhrase* CPhraseDialog::AddPhrase(
 {
     CPhrase* phrase = NULL;
     CPhraseGraph::CVertex* _vertex = data()->m_PhraseGraph.vertex(phrase_id);
+    VERIFY2(!_vertex, make_string("Duplicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()));
```
Xottab-DUTY:
```suggestion
    VERIFY2(!_vertex, make_string("Duplicate phrase ID: [%s] for phrase: [%s]. Existed phrase by this ID: [%s]", phrase_id.c_str(), text, _vertex->data()->GetText()).c_str());
```

src/xrGame/PhraseDialog.cpp
Xottab-DUTY: > The way I understand we use things like VERIFY is to check for something that should not happen because something is seriously wrong. (Having 2 phrases with the same ID sounds like this scenario to me).

@AMS21, VERIFY is sometimes used to detect things that should not happen because they are seriously wrong, but it's not only that. It's also used as a development aid for things that are *recommended* to not happen, but they are allowed to.
Only after merging this PR and getting #1759 crash I understood that we should not use XR_ASSUME for VERIFY =)

PR #1556, opened 2023-12-06
xrGame: Fix buffer overflow in `ParseControlString`

AMS21:
The given buffers are actually not big enough for the size specified in the `sscanf` argument.

To quote [cppreference](https://en.cppreference.com/w/c/io/fscanf)
> Always stores a null character in addition to the characters matched (so the argument array must have room for at least __width+1__ characters)

Full credit this was found by CodeQL

Original diagnostics:
>  src/xrGame/console_commands_mp.cpp:606: This 'sscanf string argument' operation requires 33 bytes but the destination is only 32 bytes. [cpp/very-likely-overrunning-write]
>  src/xrGame/console_commands_mp.cpp:606: This 'sscanf string argument' operation requires 17 bytes but the destination is only 16 bytes. [cpp/very-likely-overrunning-write]

### DIFF
--- a/src/xrGame/console_commands_mp.cpp
+++ b/src/xrGame/console_commands_mp.cpp
@@ -598,12 +598,10 @@ protected:
     shared_str m_action_param;
     bool ParseControlString(LPCSTR args_string)
     {
-        char action_name[17];
-        action_name[0] = 0;
-        char param_name[33];
-        param_name[0] = 0;
+        string32 action_name{};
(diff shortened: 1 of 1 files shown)


Xottab-DUTY (approved):

Xottab-DUTY:
I'd prefer using `stringN` types instead of directly defining char arrays, and since I think it's beneficial to just increase the size of the string buffers, I did that instead. :)

Issue #173, opened 2017-03-09
Merge useful changes from r8384


Xottab-DUTY:
Our repo based on 1.6.0.2 patch, but there is newer revision from which we can merge useful changes.
You can found r8384 on famous gameru.net forum where Loxotron leaked sources

CrossVR:
Well first we need to know what actually changed. Preferably we'd run a diff on the revision we originally used and the new revision that is now available.

Xottab-DUTY:
I've been already done that) And after that I submitted this issue)
Mostly there is multiplayer related, some refactorings and other stuff

CrossVR:
Do you want to do the merge yourself?

Xottab-DUTY:
Not sure of it. Main problem is that I don't know what is worth merging and what is not)

PR #1557, opened 2023-12-06
Fix incorrect format specifiers (#173)

AMS21:


src/xrCore/Animation/SkeletonMotions.cpp
```diff
@@ -332,7 +332,7 @@ void motions_container::dump()
         sz += it->second->mem_usage();
         Msg("#%3d: [%3d/%5d Kb] - %s", k, it->second->m_dwReference, it->second->mem_usage() / 1024, it->first.c_str());
     }
-    Msg("--- items: %d, mem usage: %d Kb ", container.size(), sz / 1024);
+    Msg("--- items: %uz, mem usage: %uz Kb ", container.size(), sz / 1024);
```
Xottab-DUTY: The correct usage is to specify parameter in-between `%` and `u`, so it should be `%zu`, not `%uz`.
```suggestion
    Msg("--- items: %uz, mem usage: %zu Kb ", container.size(), sz / 1024);
```

Please, fix similar things in other files.

src/xrGame/debug_text_tree.h
Xottab-DUTY, line 1: I've fixed issues in this file in recent commits.

src/xrCore/Animation/SkeletonMotions.cpp
AMS21: Oh wow, learned something new :)

AMS21:
Rebased and fixed annotated problems

Xottab-DUTY (approved):

Xottab-DUTY:
Hurray, merge! Thanks!

PR #1594, opened 2024-01-27
Task manager enhancements

olefirenque:
I would like to argue about the note "possible data-races can make it even more random":
  - data-races are likely to disrupt the random distribution
  - I suppose single random object still can cause contention by cache line collisions

I would also like to propose some heuristic for stealing single task. I think since it barely contends with other threads, it might be useful to do a few more iterations.

P.S: Are there any convenient deterministic benchmarks for comparing performance with such minor changes in xray?

### DIFF
--- a/src/xrCore/Threading/TaskManager.cpp
+++ b/src/xrCore/Threading/TaskManager.cpp
@@ -37,8 +37,6 @@ xr_unique_ptr<TaskManager> TaskScheduler;
 
 static constexpr size_t OTHER_THREADS_COUNT = 1; // Primary thread
 
-static thread_local CRandom random{};
-
 static u32 ttapi_dwFastIter = 0;
 
 class TaskStorageSize
@@ -69,7 +67,7 @@ static constexpr size_t TASK_STORAGE_MASK = TASK_STORAGE_SIZE - 1;
 class FallbackTaskAllocator
 {
     std::atomic_size_t m_allocated{};
-    Task               m_storage[TASK_STORAGE_SIZE]{};
+    Task               m_storage[TASK_STORAGE_SIZE];
 
 public:
     Task* allocate()
@@ -90,7 +88,7 @@ public:
 class TaskAllocator
 {
     size_t m_allocated{};
-    Task   m_storage[TASK_STORAGE_SIZE]{};
+    Task   m_storage[TASK_STORAGE_SIZE];
 
 public:
     Task* allocate()
@@ -168,6 +166,7 @@ public:
     std::atomic_bool sleeps{};
     Event event;
     size_t id;
+    CRandom random;
 } static thread_local s_tl_worker;
 
 static TaskWorker* s_main_thread_worker = nullptr;
@@ -293,6 +292,7 @@ void TaskManager::TaskWorkerStart()
         ScopeLock scope(&workersLock);
         workers.emplace_back(&s_tl_worker);
         s_tl_worker.id = workers.size();
+        s_tl_worker.random = CRandom(static_cast<s32>(s_tl_worker.id));
     }
     workersCount.fetch_add(1, std::memory_order_release);
     activeWorkersCount.fetch_add(1, std::memory_order_relaxed);
@@ -382,11 +382,10 @@ Task* TaskManager::TryToSteal() const
         return nullptr; // thread itself
     }
 
-
     int steal_attempts = 3;
     while (--steal_attempts >= 0)
     {
-        TaskWorker* other = workers[random.randI(count)];
+        TaskWorker* other = workers[s_tl_worker.random.randI(count)];
         if (other != &s_tl_worker)
         {
             auto* task = other->steal();
--- a/src/xrCore/Threading/TaskManager.cpp
+++ b/src/xrCore/Threading/TaskManager.cpp
@@ -37,6 +37,8 @@ xr_unique_ptr<TaskManager> TaskScheduler;
 
 static constexpr size_t OTHER_THREADS_COUNT = 1; // Primary thread
 
+static thread_local CRandom random{};
+
 static u32 ttapi_dwFastIter = 0;
 
 class TaskStorageSize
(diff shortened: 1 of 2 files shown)


src/xrCore/Threading/TaskManager.cpp
```diff
@@ -88,7 +90,7 @@ class FallbackTaskAllocator
 class TaskAllocator
 {
     size_t m_allocated{};
-    Task   m_storage[TASK_STORAGE_SIZE];
+    Task   m_storage[TASK_STORAGE_SIZE]{};
```
Xottab-DUTY: This is left uninitialized on purpose – to reduce construction time.
In fact, this storage is a ring buffer with `Task` destructors never being called on it, so it's guaranteed to have garbage.
Hence, no default initialization is ever needed.
```suggestion
    Task   m_storage[TASK_STORAGE_SIZE];
```

src/xrCore/Threading/TaskManager.cpp
```diff
@@ -67,7 +69,7 @@ static constexpr size_t TASK_STORAGE_MASK = TASK_STORAGE_SIZE - 1;
 class FallbackTaskAllocator
 {
     std::atomic_size_t m_allocated{};
-    Task               m_storage[TASK_STORAGE_SIZE];
+    Task               m_storage[TASK_STORAGE_SIZE]{};
```
Xottab-DUTY: This class works the same as `TaskAllocator` class.
Refer to the comment about that class.
```suggestion
    Task               m_storage[TASK_STORAGE_SIZE];
```

src/xrCore/Threading/TaskManager.hpp
```diff
@@ -22,7 +22,6 @@ class TaskWorker;
 
 class XRCORE_API TaskManager final
 {
-private:
```
Xottab-DUTY: Please, don't remove `private` and `public` labels. These are there to break the code into logical blocks.

src/xrCore/Threading/TaskManager.cpp
```diff
@@ -37,6 +37,8 @@ xr_unique_ptr<TaskManager> TaskScheduler;
 
 static constexpr size_t OTHER_THREADS_COUNT = 1; // Primary thread
 
+static thread_local CRandom random{};
```
Xottab-DUTY: It's better to still have this variable in the TaskManager class.

src/xrCore/Threading/TaskManager.cpp
```diff
@@ -380,13 +382,18 @@ Task* TaskManager::TryToSteal(TaskWorker* thief)
         return nullptr; // thread itself
     }
 
-    TaskWorker* other = workers[random.randI(count)];
-    if (other != thief)
+
```
Xottab-DUTY:
```suggestion
```

Xottab-DUTY (changes requested):
Generally code style changes needed.
Also, we don't need to default-initialize allocators storage.

src/xrCore/Threading/TaskManager.cpp
olefirenque:
```c++
class XRCORE_API TaskManager final
{
private:
    ...
    static thread_local CRandom random;
```

```
"private: static CRandom TaskManager::random": данные с длительностью хранилища потока не могут иметь интерфейс dll
```

olefirenque: I think it can be moved to `s_tl_worker`

src/xrCore/Threading/TaskManager.cpp
```diff
@@ -166,6 +166,7 @@ class TaskWorker : public TaskQueue, public TaskWorkerStats
     std::atomic_bool sleeps{};
     Event event;
     size_t id;
+    CRandom random;
```
Xottab-DUTY:
```suggestion
    CRandom random{ s32(std::intptr_t(this)) };
```

src/xrCore/Threading/TaskManager.cpp
```diff
@@ -291,6 +292,7 @@ void TaskManager::TaskWorkerStart()
         ScopeLock scope(&workersLock);
         workers.emplace_back(&s_tl_worker);
         s_tl_worker.id = workers.size();
+        s_tl_worker.random = CRandom(static_cast<s32>(s_tl_worker.id));
```
Xottab-DUTY:
```suggestion
```

src/xrCore/Threading/TaskManager.cpp
```diff
@@ -380,13 +382,17 @@ Task* TaskManager::TryToSteal(TaskWorker* thief)
         return nullptr; // thread itself
     }
 
-    TaskWorker* other = workers[random.randI(count)];
-    if (other != thief)
+    int steal_attempts = 3;
+    while (--steal_attempts >= 0)
```
Xottab-DUTY:
```suggestion
    int steal_attempts = 3;
    while (steal_attempts > 0)
```

src/xrCore/Threading/TaskManager.cpp
```diff
@@ -380,13 +382,17 @@ Task* TaskManager::TryToSteal(TaskWorker* thief)
...
-            other->event.Set(); // Wake up, you have work to do!
-        return task;
+        TaskWorker* other = workers[s_tl_worker.random.randI(count)];
+        if (other != &s_tl_worker)
+        {
+            auto* task = other->steal();
+            if (!other->empty() && other->sleeps.load(std::memory_order_relaxed))
+                other->event.Set(); // Wake up, you have work to do!
+            return task;
+        }
```
Xottab-DUTY:
```suggestion
        if (other == &s_tl_worker)
            continue;
        auto* task = other->steal();
        if (!other->empty() && other->sleeps.load(std::memory_order_relaxed))
            other->event.Set(); // Wake up, you have work to do!
        if (task)
            return task;
        --steal_attempts;
```

Xottab-DUTY:
Oh, wow
4666 and 4777 :D
image

Xottab-DUTY:
@olefirenque, thanks!

PR #1600, opened 2024-02-01
Bunch of microoptimizations

olefirenque:


src/xrGame/cover_manager.cpp
```diff
@@ -74,37 +74,42 @@ void CCoverManager::compute_static_cover()
...
+    
     const CLevelGraph& graph = ai().level_graph();
-    const u32 levelVertexCount = ai().level_graph().header().vertex_count();
+    const LevelGraph::CHeader &levelGraphHeader = graph.header();
+    const u32 levelVertexCount = levelGraphHeader.vertex_count();
+
+    m_covers = xr_new<CPointQuadTree>(levelGraphHeader.box(), levelGraphHeader.cell_size() * .5f, 8 * 65536, 4 * 65536);
 
+    // avoiding extra allocations with a static storage for m_covers
+    static xr_vector<std::optional<CCoverPoint>> quadTreeStaticStorage;
```
olefirenque, line 85: I'm not sure if this is a good trade-off between allocs/RSS. It would be nice to hear an opinion on this

src/xrAICore/Navigation/level_graph_inline.h
```diff
@@ -571,6 +571,14 @@ IC void CLevelGraph::clear_mask_no_check(u32 vertex_id) { m_access_mask[vertex_i
 template <typename P>
 IC void CLevelGraph::iterate_vertices(
     const Fvector& min_position, const Fvector& max_position, const P& predicate) const
+{
+    auto [I, E] = get_range(min_position, max_position);
```
Xottab-DUTY: Although there's `I` and `E` in the codebase everywhere (original code), we don't use that with the current codestyle.
```suggestion
    auto [it, ite] = get_range(min_position, max_position);
```

src/xrGame/space_restrictor.cpp
```diff
@@ -100,7 +100,14 @@ void CSpaceRestrictor::net_Destroy()
...
     if (!actual())
-        prepare();
+    {
+        static std::mutex prepareMutex;
+        std::lock_guard lock(prepareMutex);
+
+        // Double-checked locking
+        if (!actual())
+            prepare();
+    }
```
olefirenque, line 110: I believe it was only non-concurrent part for `xr_parallel_for` in `CSpaceRestrictionShape::fill_shape`. Since it is called only once for a sequence of `CSpaceRestrictor::inside` calls, maybe it would be better to extract this to reduce contention

src/xrAICore/Navigation/level_graph_inline.h
```diff
@@ -571,6 +571,14 @@ IC void CLevelGraph::clear_mask_no_check(u32 vertex_id) { m_access_mask[vertex_i
...
+
+    for (; I != E; ++I)
+        predicate(*I);
+}
+
+IC std::pair<CLevelGraph::CLevelVertex *, CLevelGraph::CLevelVertex *> CLevelGraph::get_range(const Fvector& min_position, const Fvector& max_position) const
```
Xottab-DUTY:
```suggestion
IC std::pair<CLevelGraph::CLevelVertex*, CLevelGraph::CLevelVertex*> CLevelGraph::get_range(const Fvector& min_position, const Fvector& max_position) const
```

src/xrGame/cover_manager.h
```diff
@@ -44,7 +44,9 @@ class CCoverManager
 
 protected:
     CPointQuadTree* m_covers;
-    xr_vector<bool> m_temp;
+    // vector<bool> is not applicable for `m_temp`
+    // since it is filled in parallel_for (https://timsong-cpp.github.io/cppwp/container.requirements.dataraces).
+    xr_vector<int> m_temp;
```
olefirenque, line 49: I suppose it wasn't thread-safe before

src/xrAICore/Navigation/level_graph_inline.h
```diff
@@ -591,8 +599,7 @@ IC void CLevelGraph::iterate_vertices(
     else
         E = end;
 
-    for (; I != E; ++I)
-        predicate(*I);
+    return {I, E};
```
Xottab-DUTY: Codestyle: always leave whitespace between `{}` if they are non-empty. (and always leave no whitespace, if they are empty)
 ```suggestion
    return { I, E };
```

File: src/Layers/xrRender_R2/r2_R_lights.cpp [hunk 7]
```diff
@@ -42,25 +43,26 @@ void CRender::render_lights(light_Package& LP)
...
         {
             LP_smap_pool.initialize(RImplementation.o.smapsize);
-            std::sort(source.begin(), source.end(), [](light* l1, light* l2)
-            {
-                const u32 a0 = l1->X.S.size;
-                const u32 a1 = l2->X.S.size;
-                return a0 > a1; // reverse -> descending
-            });
-            for (size_t test = 0; test < source.size(); ++test)
+            for (auto &L : source)
```
Xottab-DUTY:
```suggestion
            for (auto& L : source)
```

src/xrGame/cover_manager.cpp
```diff
@@ -74,37 +74,42 @@ void CCoverManager::compute_static_cover()
...
-
-            m_temp[i] = false;
         }
     });
-
-    for (u32 i = 0; i < levelVertexCount; ++i)
-        if (m_temp[i] && critical_cover(i))
-            m_covers->insert(xr_new<CCoverPoint>(ai().level_graph().vertex_position(ai().level_graph().vertex(i)), i));
+    for (auto &p : quadTreeStaticStorage)
+        if (p.has_value())
```
olefirenque, line 111: It seemed that accessing `m_temp` and calling `critical_cover` took some time here
```c++
for (u32 i = 0; i < levelVertexCount; ++i)
        if (m_temp[i] && critical_cover(i))
            m_covers->insert(xr_new<CCoverPoint>(ai().level_graph().vertex_position(ai().level_graph().vertex(i)), i));
```
image

src/Layers/xrRender_R2/r2_R_lights.cpp
```diff
@@ -42,25 +43,26 @@ void CRender::render_lights(light_Package& LP)
         }
     }
 
-    // 2. refactor - infact we could go from the backside and sort in ascending order
```
Xottab-DUTY, line 31: Don't remove this comment, please, as it relates to the comment section in the beginning of the function :)

src/Layers/xrRender_R2/r2_R_lights.cpp
```diff
@@ -129,14 +129,14 @@ void CRender::render_lights(light_Package& LP)
         }
     };
 
-    const auto& flush_lights = [&]()
+    const auto& flush_lights = [] (CRender& render, xr_vector<task_data_t>& lights_queue)
```
olefirenque: I guess it's doubtful, but I suppose a lambda without capture is more likely to be inlined.

Xottab-DUTY (commented):
Some small initial review. I didn't checked changes in xrGame yet.
Please, fix the code style there as per review comments :)

src/Layers/xrRender_R2/r2_R_lights.cpp
Xottab-DUTY: `CRender` is being instantiated only once as a global variable called `RImplementation`.
Instead of passing it as `CRender&` render variable, it's better to just capture this.

Xottab-DUTY: > I guess it's doubtful, but I suppose a lambda without capture is more likely to be inlined.

Ah, I see. But what if we capture only `this`?
And also I'm not sure if inlining helps here.

src/Layers/xrRender_R2/r2_R_lights.cpp
olefirenque: It has been removed for consistency since `source` is already sorted in ascending order, maybe this comment should be changed

src/xrGame/space_restriction_shape.cpp
```diff
@@ -79,10 +80,43 @@ void CSpaceRestrictionShape::fill_shape(const CCF_Shape::shape_def& shape)
...
+        for (auto x : m_border_chunk)
+            m_border.push_back(x);
+    });
 
 #ifdef DEBUG
-    ai().level_graph().iterate_vertices(start, dest, CShapeTestPredicate(this));
+    xr_parallel_for(TaskRange(begin, end), [this, &mergeMutex, &graph] (const auto& range) {
+        xr_vector<u32> m_test_storage_chunk;
+        m_test_storage_chunk.reserve(range.size());
+        for (auto &vertex : range)
```
Xottab-DUTY:
```suggestion
        for (auto& vertex : range)
```

src/xrGame/space_restriction_shape.cpp
```diff
@@ -79,10 +80,43 @@ void CSpaceRestrictionShape::fill_shape(const CCF_Shape::shape_def& shape)
...
+        std::lock_guard lock(mergeMutex);
+        if (m_border.capacity() < m_border.size() + m_border_chunk.size())
+            m_border.reserve(m_border.size() + m_border_chunk.size());
+        for (auto x : m_border_chunk)
+            m_border.push_back(x);
+    });
 
 #ifdef DEBUG
-    ai().level_graph().iterate_vertices(start, dest, CShapeTestPredicate(this));
+    xr_parallel_for(TaskRange(begin, end), [this, &mergeMutex, &graph] (const auto& range) {
```
Xottab-DUTY:
```suggestion
    xr_parallel_for(TaskRange(begin, end), [this, &mergeMutex, &graph] (const auto& range)
    {
```

src/xrGame/space_restriction_shape.cpp
```diff
@@ -79,10 +80,43 @@ void CSpaceRestrictionShape::fill_shape(const CCF_Shape::shape_def& shape)
...
     default: NODEFAULT;
     }
-    ai().level_graph().iterate_vertices(start, dest, CBorderMergePredicate(this));
+
+    CLevelGraph& graph = ai().level_graph();
+
+    std::mutex mergeMutex;
+
+    auto [begin, end] = graph.get_range(start, dest);
+    xr_parallel_for(TaskRange(begin, end), [this, &mergeMutex, &graph] (const auto& range) {
```
Xottab-DUTY:
```suggestion
    xr_parallel_for(TaskRange(begin, end), [this, &mergeMutex, &graph] (const auto& range)
    {
```

src/Layers/xrRender_R2/r2_R_lights.cpp
Xottab-DUTY: Yea, we can at least remove `- infact we could go from the backside and sort in ascending order` part and leave it like this:
`// 2. refactor`

Later we can add some brief info about how it works maybe.

src/xrGame/space_restriction_shape.cpp
```diff
@@ -79,10 +80,45 @@ void CSpaceRestrictionShape::fill_shape(const CCF_Shape::shape_def& shape)
...
+        {
+            if (inside(graph.vertex_id(&vertex), true) &&
+                !inside(graph.vertex_id(&vertex), false))
+                m_border_chunk.push_back(graph.vertex_id(&vertex));
+        }
+        std::lock_guard lock(mergeMutex);
+        if (m_border.capacity() < m_border.size() + m_border_chunk.size())
+            m_border.reserve(m_border.size() + m_border_chunk.size());
+        for (auto x : m_border_chunk)
+            m_border.push_back(x);
```
olefirenque, line 103: I did a little research on this, and it seemed to me that the body of `CBorderMergePredicate::operator(...)` is a bigger problem than merging a certain number of chunks under lock
image

olefirenque: Ok, it might be irrelevant since I accidentally used `UI Freeze` option.
I'm sorry for the misinformation. Apparently, this flame graph represents only those samples that were sampled when the NPC's spawning was lagging.

> The UI Freeze event indicates time intervals where the application was unable to respond to user input. More specifically, these are time intervals where window messages were not pumped for more than 200 ms or processing of a particular message took more than 200 ms.

src/Layers/xrRender_R2/r2_R_lights.cpp
```diff
@@ -129,14 +130,14 @@ void CRender::render_lights(light_Package& LP)
...
-    const auto& flush_lights = [&]()
+    const auto& flush_lights = [] (xr_vector<task_data_t>& lights_queue)
     {
         for (const auto& [L, task, batch_id] : lights_queue)
         {
             VERIFY(task);
             TaskScheduler->Wait(*task);
 
-            auto& dsgraph = get_context(batch_id);
+            auto& dsgraph = RImplementation.get_context(batch_id);
```
Xottab-DUTY: This is still look weird because capturing `this` and then using CRender functions directly looks more natural.

And it's not proven that inlining would make a difference. Can we measure it?

olefirenque: Well, at least for gcc 13.2, it doesn't seem to make any difference.
https://godbolt.org/z/8WafxGxoG
https://godbolt.org/z/M56ns3afq

I'll return the capture

Xottab-DUTY: But still, don't just return original code which captures everything. It's better to capture `this`, that makes lambda-object smaller (if the compiler doesn't automatically deduce that it needs to capture `this`)

Xottab-DUTY: Ah, you did it already :)

src/Layers/xrRenderDX11/dx11SH_Texture.cpp
```diff
@@ -117,23 +117,17 @@ void CTexture::surface_set(ID3DBaseTexture* surf)
...
+                ViewDesc.Texture2DMSArray.ArraySize = desc.ArraySize;
+            
             srv_per_slice.resize(desc.ArraySize);
-            for (u32 id = 0; id < desc.ArraySize; ++id)
-            {
-                _RELEASE(srv_per_slice[id]);
+            for (auto &x : srv_per_slice)
+                _RELEASE(x);
+
+            CHK_DX(HW.pDevice->CreateShaderResourceView(pSurface, &ViewDesc, srv_per_slice.data()));
```
olefirenque: I am an absolute lamer in matters of rendering and the DX11 api, but It seems like in can be simplified in this way

Xottab-DUTY: > It seems like in can be simplified in this way

No, it can't be simplified. This was made on purpose – SRV per slice. It basically makes the engine work with original game shaders, because they were written to work with one texture, not texture arrays.

So, your change breaks the game :)

olefirenque: I played with "release master gold" build for ~15 minutes and it didn't break. 
But in mixed build type it feels like it decreases performance for ~3-5fps.

Doesn't this call just create a shader payload that should be equal in the end?

olefirenque: well, ok, I see this now
image

Xottab-DUTY: It doesn't create the same payload.
We want to create a number of SRVs, where each created SRV contains only one texture (slice) from a texture array srv_all.
With this change, only one SRV that contains an entire texture array with all textures instead of just one texture is being created.

Xottab-DUTY: I thought that would just break the shaders, but it behaves differently: D3D11 runtime just always binds first element of the texture array during the shaders launch, where we want to use different slices, not only the first.

olefirenque: Ok, thanks. By the way, do you have an idea how these calls could be handled concurrently if I did something like this:
```c++
xr_parallel_for(TaskRange<u32>(0, desc.ArraySize), [this, ViewDesc] (const auto& range) {
    D3D_SHADER_RESOURCE_VIEW_DESC ViewDescLocal = ViewDesc; 
    UINT* FirstArraySlice = &ViewDescLocal.Texture2DMSArray.FirstArraySlice;
    if (desc.SampleDesc.Count <= 1)
        FirstArraySlice = &ViewDescLocal.Texture2DArray.FirstArraySlice;
    
    for (u32 id = range.begin(); id != range.end(); ++id)
    {
        _RELEASE(srv_per_slice[id]);
        *FirstArraySlice = id;
        CHK_DX(HW.pDevice->CreateShaderResourceView(pSurface, &ViewDescLocal, &srv_per_slice[id]));
    }
});
```

olefirenque: And is rewriting shaders is an option?

Xottab-DUTY: > And is rewriting shaders is an option?

Yes and no.
We can have modified shaders but still must maintain compatibility with vanilla shaders.



> Ok, thanks. By the way, do you have an idea how these calls could be handled concurrently if I did something like this:
> 
> ```c++
> xr_parallel_for(TaskRange<u32>(0, desc.ArraySize), [this, ViewDesc] (const auto& range) {
>     D3D_SHADER_RESOURCE_VIEW_DESC ViewDescLocal = ViewDesc; 
>     UINT* FirstArraySlice = &ViewDescLocal.Texture2DMSArray.FirstArraySlice;
>     if (desc.SampleDesc.Count <= 1)
>         FirstArraySlice = &ViewDescLocal.Texture2DArray.FirstArraySlice;
>     
>     for (u32 id = range.begin(); id != range.end(); ++id)
>     {
>         _RELEASE(srv_per_slice[id]);
>         *FirstArraySlice = id;
>         CHK_DX(HW.pDevice->CreateShaderResourceView(pSurface, &ViewDescLocal, &srv_per_slice[id]));
>     }
> });
> ```

I'm not sure if `CreateShaderResourceView` is thread-safe, but fixing just this code part while keeping same approach is still suboptimal.
Normally, we should not call `CreateShaderResourceView` every frame, which we are exactly doing currently.
And it's better to implement that change in a separate pull request if you attempt to do it.

@vTurbine might tell more

Xottab-DUTY: > And it's better to implement that change in a separate pull request if you attempt to do it.

That will be a major code rewrite, so I'm not sure how much efforts it will take.

olefirenque: > Ok, thanks. By the way, do you have an idea how these calls could be handled concurrently if I did something like this:
> 
> ```c++
> xr_parallel_for(TaskRange<u32>(0, desc.ArraySize), [this, ViewDesc] (const auto& range) {
>     D3D_SHADER_RESOURCE_VIEW_DESC ViewDescLocal = ViewDesc; 
>     UINT* FirstArraySlice = &ViewDescLocal.Texture2DMSArray.FirstArraySlice;
>     if (desc.SampleDesc.Count <= 1)
>         FirstArraySlice = &ViewDescLocal.Texture2DArray.FirstArraySlice;
>     
>     for (u32 id = range.begin(); id != range.end(); ++id)
>     {
>         _RELEASE(srv_per_slice[id]);
>         *FirstArraySlice = id;
>         CHK_DX(HW.pDevice->CreateShaderResourceView(pSurface, &ViewDescLocal, &srv_per_slice[id]));
>     }
> });
> ```

Apparently, this only has a negative effect

before:
image

after:
image

It seems like `CreateShaderResourceView` is threadsafe, but not concurrent

src/xrGame/cover_manager.cpp
375gnu: It's not recommended to use global/static objects which allocate memory dynamically.

Xottab-DUTY:
@olefirenque, hi! Sorry for a big delay in review & merging!
I've made big improvements to the task manager: now it supports lambda with captures.
Could you rebase this PR to accommodate the changes?

src/xrGame/space_restriction_shape.cpp
Xottab-DUTY: @olefirenque, what profiler did you use?

AMS21: To me this strongly looks like [tracy](https://github.com/wolfpld/tracy) but I might be wrong there

PR #1606, opened 2024-02-12
Use CMake 3.22.1 until 3.28.2->3.28.3 update

Masterkatze:


Xottab-DUTY:
I think we don't need this already, we can just wait until the updated 3.28.3 version will be settled down on GitHub Actions runners.

Ubuntu Debug amd64 already got updated and compiles successfully now!
https://github.com/OpenXRay/xray-16/actions/runs/7881493553/job/21505204788

I think that will take maybe a week?

Masterkatze:
Yeah, probably, I just wanted to check if it could be done quickly anyway.

Xottab-DUTY:
Lol, that took less than a day. Issue was addressed it seems.
https://github.com/OpenXRay/xray-16/actions/runs/7887529899/job/21522982295?pr=1608

PR #1608, opened 2024-02-13
CMake: Use CMAKE_SYSTEM_PROCESSOR instead of ARCH_TYPE variable

AMS21:
This was caused by using the `CMAKE_SIZEOF_VOID_P` variable before it is actually set by cmake (after a call to project).

https://cmake.org/cmake/help/latest/variable/CMAKE_SIZEOF_VOID_P.html

> This is set to the size of a pointer on the target machine, and is
> determined when a compiled language is enabled. (...)

Also the `else(...)` construct in CMake works a bit different than the original author expected since the expression is not evaluated and thus we always set `ARCH_TYPE` to `x86` which is not correct.

Also took the liberty to properly output the `CMAKE_PROJECT_VERSION` and the `ARCH_TYPE` variables.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,18 @@ project(OpenXRay
     HOMEPAGE_URL "https://github.com/OpenXRay/xray-16"
     LANGUAGES CXX C
 )
-message(STATUS "CMAKE_PROJECT_VERSION: ${}")
+
+# Detect arch type ( x86 or x64 )
+if (CMAKE_SIZEOF_VOID_P EQUAL 8)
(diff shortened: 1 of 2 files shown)


Xottab-DUTY (approved):
Much thanks!!!

When I was refactoring that part of the code during the creation of 52c3e2e3f15a9bfeecc85f1283230bec63e588ce commit, I noticed that ARCH_TYPE was used to set the binaries output folder and CMAKE_SYSTEM_PROCESSOR is better fit there.
But I decided to not touch it yet.

I don't remember why ARCH_TYPE was introduced, but we surely can just remove it.

PR #1624, opened 2024-03-03
[POC] Restore 3rd party dependencies using NuGet for MSVC generator

pavel-rodionov:
Attempt to bring project closer to migration from hardcoded MSVS solutions to Cmake based generator. 

Verification:
* Install nuget cli `winget install -e --id Microsoft.NuGet`
* Run Cmake generator `cmake ..\xray-16 -G "Visual Studio 17 2022" -A x64 -DXRAY_USE_LUAJIT=OFF`
* Verify that packages listed as nuget dependencies are downloaded into `./packages`
* Open MSVC solution. Verify that target `xrLC_Light` does include `d3dx9` headers and they are included from expected location `..\..\..\packages\Microsoft.DXSDK.D3DX.9.29.952.8\build\native\include\d3dx9.h`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,11 +149,13 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT XRAY_USE_DEFAULT_CXX_LIB)
     endif()
 endif()
 
-add_compile_options(-Wno-attributes)
-if (APPLE)
-    add_compile_options(-Wl,-undefined,error)
-else()
-    add_compile_options(-Wl,--no-undefined)
+if (NOT MSVC)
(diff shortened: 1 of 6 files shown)


MerijnHendriks:
Is nuget from dotnet sdk also usable with this pr? What about offline builds? (nuget restore requires an internet connection, otherwise build fails).

pavel-rodionov:
@MerijnHendriks 
If nuget cli is available in your system PATH it should work. With that said I haven't tested it.

Regarding offline builds - when files are already downloaded and available in local file system it works offline just fine. To initially restore files you obviously need an active network connection (that is on pair with current solution when VS restore them using dependencies defined in solutions)

PR #1635, opened 2024-03-15
Remove incorrect resolution scaling code

yohjimane:
The code incorrectly takes a smaller subregion and scales it up, rather than updating the internal resolution and scaling it up.


Xottab-DUTY:
Thanks!
That was incorrect code indeed!
It should be used differently, according to this article: https://learn.microsoft.com/en-us/windows/uwp/gaming/multisampling--scaling--and-overlay-swap-chains

PR #1667, opened 2024-05-06
Fix crash during TaskScheduler creation

sobkas:
Fixes run when compiled with gcc higher than 10

Tested it on gcc-14 and binary worked.
But need more tests.

### DIFF
--- a/src/xrCore/Threading/TaskManager.cpp
+++ b/src/xrCore/Threading/TaskManager.cpp
@@ -219,7 +219,7 @@ TaskManager::~TaskManager()
     }
 }
 
-void TaskManager::MakeThreads()
+void TaskManager::SpawnThreads()
 {
     workers.reserve(std::thread::hardware_concurrency());
 
--- a/src/xrCore/Threading/TaskManager.cpp
+++ b/src/xrCore/Threading/TaskManager.cpp
@@ -192,7 +192,22 @@ void CalcIterations()
 TaskManager::TaskManager()
 {
     ZoneScoped;
+    workers.reserve(std::thread::hardware_concurrency());
+    RegisterThisThreadAsWorker();
+}
+
+void TaskManager::SpawnThreads()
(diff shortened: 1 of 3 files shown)


sobkas:
Ok, fixed branch

Xottab-DUTY:
~~What's the issue?~~

image

src/xrCore/Threading/TaskManager.hpp
```diff
@@ -78,6 +78,7 @@ class XRCORE_API TaskManager final
 public:
     void RegisterThisThreadAsWorker();
     void UnregisterThisThreadAsWorker();
+    void MakeThreads();
```
Xottab-DUTY:
```suggestion
    void SpawnThreads();
```

src/xrCore/xrCore.cpp
```diff
@@ -274,6 +274,7 @@ void xrCore::Initialize(pcstr _ApplicationName, pcstr commandLine, bool init_fs,
         Msg("\ncommand line %s\n", Params);
         _initialize_cpu();
         TaskScheduler = xr_make_unique<TaskManager>();
+        TaskScheduler->MakeThreads();
```
Xottab-DUTY:
```suggestion
        TaskScheduler->SpawnThreads();
```

src/xrCore/Threading/TaskManager.cpp
```diff
@@ -230,6 +219,20 @@ TaskManager::~TaskManager()
     }
 }
 
+void TaskManager::MakeThreads()
```
Xottab-DUTY:
```suggestion
void TaskManager::SpawnThreads()
```

Xottab-DUTY (approved):

sobkas:
As it fixes problems with gcc newer than 10, maybe adding that to build information should be ok?

PR #1668, opened 2024-05-06
debian/: Switch to use gcc-14

sobkas:
Switch to use gcc-14
Depends on: https://github.com/OpenXRay/xray-16/pull/1667

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+openxray (1.6.02-15) UNRELEASED; urgency=medium
+
+  * Move to use gcc-14
+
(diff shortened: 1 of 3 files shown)


Xottab-DUTY:
Not that long ago I find out that we already have DEB package specs – they are being generated by CMake. If I understand correctly.
That means we have code/logic duplication and, what's worse, these manually created specs are different from what CMake generates.
Are there any disadvantages if we just delete `debian` folder and move all that logic to CMake files?

sobkas:
Static debian/ folder allows me to compile debian packages using my existing infrastructure.
Also debian/ dir is simpler to maintain than CMake.

PR #1672, opened 2024-05-15
SDL3: Initial SDL3 port

sobkas:
WIP: Inital SDL3 Port


sobkas:
Supporting both SDL2 and SDL3 will negatively influence code quality.
Until full switch to SDL3 is ok, this will stay closed.

PR #1676, opened 2024-05-17
Initial multitrackbar component, SSS options page, hud_fov menu option

yohjimane:
Adds:
- Multitrackbar component, to easily update Fvector3/Fvector4 console params
- Options page to configure SSS options in Options > Video tab > Shader Options + supplementary xml translations from original mod
- Changes hud_fov to work independently of regular fov, adds trackbar option for hud_fov in Game options tab


yohjimane:
Need to do some cleanup, add russian translations, and add regular aspect ratio shader options page

yohjimane:
Gonna revert those hud_fov refactors in favor of just adding menu option in this commit- those ended up causing a few issues :(

PR #1678, opened 2024-05-20
Clear Sky:  Missing variable in gun setting file

sobkas:
Crash when some weapons(eg. wpn_val) with built in suppressor don't have snd_silncer_shot only snd_shoot

### DIFF
--- a/src/xrGame/WeaponMagazined.cpp
+++ b/src/xrGame/WeaponMagazined.cpp
@@ -98,7 +98,8 @@ void CWeaponMagazined::Load(LPCSTR section)
             m_sSilencerSmokeParticles = pSettings->r_string(section, "silencer_smoke_particles");
 
         //Alundaio: LAYERED_SND_SHOOT Silencer
-        m_layered_sounds.LoadSound(section, "snd_silncer_shot", "sndSilencerShot", false, m_eSoundShot);
+        if (WeaponSoundExist(section, "snd_silncer_shot"))
+            m_layered_sounds.LoadSound(section, "snd_silncer_shot", "sndSilencerShot", false, m_eSoundShot);
         if (WeaponSoundExist(section, "snd_silncer_shot_actor"))
             m_layered_sounds.LoadSound(section, "snd_silncer_shot_actor", "sndSilencerShotActor", false, m_eSoundShot);
         //-Alundaio
@@ -1065,7 +1066,9 @@ void CWeaponMagazined::InitAddons()
         }
     }
 
-    if (IsSilencerAttached() /* && SilencerAttachable() */)
+    if (IsSilencerAttached() && pSettings->line_exist(*cNameSect(), "silencer_flame_particles")
+        && pSettings->line_exist(*cNameSect(), "silencer_light_color")
+        && pSettings->line_exist(*cNameSect(), "silencer_smoke_particles"))
     {
         m_sFlameParticlesCurrent = m_sSilencerFlameParticles;
         m_sSmokeParticlesCurrent = m_sSilencerSmokeParticles;


src/xrCore/xr_ini.h
```diff
@@ -195,6 +195,7 @@ class XRCORE_API CInifile
     CLASS_ID r_clsid(pcstr S, pcstr L) const;
     CLASS_ID r_clsid(const shared_str& S, pcstr L) const { return r_clsid(*S, L); }
     pcstr r_string(pcstr S, pcstr L) const; // Left quotes in place
+    pcstr r_string_try(pcstr S, pcstr L) const; // Left quotes in place
```
Xottab-DUTY: This function duplicates `read_if_exists<pcstr>`.

src/xrGame/WeaponMagazined.cpp
```diff
@@ -98,7 +98,10 @@ void CWeaponMagazined::Load(LPCSTR section)
             m_sSilencerSmokeParticles = pSettings->r_string(section, "silencer_smoke_particles");
 
         //Alundaio: LAYERED_SND_SHOOT Silencer
-        m_layered_sounds.LoadSound(section, "snd_silncer_shot", "sndSilencerShot", false, m_eSoundShot);
+        if (!m_layered_sounds.LoadSoundTry(section, "snd_silncer_shot", "sndSilencerShot", false, m_eSoundShot))
+        {
+          m_layered_sounds.LoadSound(section, "snd_shoot", "sndSilencerShot", false, m_eSoundShot);
+        }
```
Xottab-DUTY: The existing pipeline when processing LTX is to explicitly check for line existence is to use `line_exist` function.
There are helpers like `read_if_exists` functions and `READ_IF_EXISTS` macro, they should be used when we can assign the value immediately and straightforward.

For weapon sounds we have a helper function `WeaponSoundExist`, please use it, don't duplicate existing functionality.
Also, is it possible to assign already loaded sound collection (`sndShot`) to the one we are processing here (`sndSilencerShot`)? It would be optimal to do this.

Xottab-DUTY (changes requested):
Thanks!
There are some notes about the solution.

Xottab-DUTY:
I just checked the difference between CS and COP sources:
image

CS code was trying to use `sndSilencerShot` only if silencer attached and is actually attachable/detachable, assuming the use of normal `sndShot` when there's no silencer or it's built-in. In COP `SilencerAttachable()` check was commented out.

The most natural fix would be checking for `sndSilencerShot` existence in InitAddons function on line `1068`.
Replace this:
```cpp
    if (IsSilencerAttached() /* && SilencerAttachable() */)
```
With this:
```cpp
    if (IsSilencerAttached() && m_layered_sounds.FindSoundItem("sndSilencerShot", false))
```

Please, change the code accordingly!

sobkas:
> I just checked the difference between CS and COP sources: image
> 
> CS code was trying to use `sndSilencerShot` only if silencer attached and is actually attachable/detachable, assuming the use of normal `sndShot` when there's no silencer or it's built-in. In COP `SilencerAttachable()` check was commented out.
> 
> The most natural fix would be checking for `sndSilencerShot` existence in InitAddons function on line `1068`. Replace this:
> 
> ```c++
>     if (IsSilencerAttached() /* && SilencerAttachable() */)
> ```
> 
> With this:
> 
> ```c++
>     if (IsSilencerAttached() && m_layered_sounds.FindSoundItem("sndSilencerShot", false))
> ```
> 
> Please, change the code accordingly!

Well it fixes also other problem I have found wpn_pb

src/xrGame/WeaponMagazined.cpp
```diff
@@ -90,7 +90,7 @@ void CWeaponMagazined::Load(LPCSTR section)
     m_sSndShotCurrent = "sndShot";
 
     //звуки и партиклы глушителя, если такой есть
-    if (m_eSilencerStatus == ALife::eAddonAttachable || m_eSilencerStatus == ALife::eAddonPermanent)
+    if ((m_eSilencerStatus == ALife::eAddonAttachable || m_eSilencerStatus == ALife::eAddonPermanent) && WeaponSoundExist(section, "snd_silncer_shot"))
```
Xottab-DUTY: Please, don't disable flame particles when shot sound doesn't exist. This code does exactly that.

Xottab-DUTY:
> Well it fixes also other problem I have found wpn_pb

What was that problem?

sobkas:
> > Well it fixes also other problem I have found wpn_pb
> 
> What was that problem?

It doesn't have silencer_light_color and family(all silencer_light_*)
So when it tries to use any of them it crashes.

;silencer_flame_particles               = weapons\generic_weapon03
;silencer_smoke_particles               = weapons\generic_shoot_00

;silencer_light_color                   = 0.6, 0.5, 0.3
;silencer_light_range                   = 0.01
;silencer_light_var_color               = 0.05
;silencer_light_var_range               = 0.5
;silencer_light_time                    = 0.2

sobkas:
[t1.zip](https://github.com/user-attachments/files/15540895/t1.zip)
When used with SRP_v1.1.4:Optional Features/#2. (Choose From This Folder Last)/Alternative Ballistics it causes crash

sobkas:
Without it merged or getting feedback I can't continue to test Clear Sky support

Xottab-DUTY:
Sorry, for the long long delay in the review and merge.

There were some uncertainties I wanted to check and it turned out that they were not groundless.
I've fixed everything and a bit more!

PR #1681, opened 2024-05-25
Glew to epoxy

sobkas:
Switch from ancient GLEW to libepoxy


AMS21:
libepoxys last release is 2 years old so why not use the still actively developed [glad](https://github.com/Dav1dde/glad) library, which also supports Vulkan out of the box?

sobkas:
> libepoxys last release is 2 years old so why not use the still actively developed [glad](https://github.com/Dav1dde/glad) library, which also supports Vulkan out of the box?

GLAD2 depends on generated code and this code is separated for each API.
So for each API  there are files(*.c and *.h) that would be included into OpenXRay.
Epoxy needs only #include to be added.

Xottab-DUTY:
> libepoxys last release is 2 years old so why not use the still actively developed [glad](https://github.com/Dav1dde/glad) library, which also supports Vulkan out of the box?

GLEW's last release is 3 years old :)

Though, there's a bad sign there: libepoxy doesn't have any fresh commits since the last release, while glew has.

sobkas:
> > libepoxys last release is 2 years old so why not use the still actively developed [glad](https://github.com/Dav1dde/glad) library, which also supports Vulkan out of the box?
> 
> GLEW's last release is 3 years old :)
> 
> Though, there's a bad sign there: libepoxy doesn't have any fresh commits since the last release, while glew has.

libepoxy is used by much more recent software, while glew usage is ancient.

src/Layers/xrRenderGL/glHW.cpp
```diff
@@ -5,6 +5,7 @@
 #pragma hdrstop
 
 #include "glHW.h"
+#include <epoxy/gl.h>
 #include "xrEngine/XR_IOConsole.h"
```
Xottab-DUTY, line 10:
```suggestion
#include "xrEngine/XR_IOConsole.h"

#include <epoxy/gl.h>

```

src/Layers/xrRenderGL/glHW.cpp
```diff
@@ -81,8 +82,8 @@ void CHW::CreateDevice(SDL_Window* hWnd)
     // Choose the closest pixel format
     SDL_DisplayMode mode;
     SDL_GetWindowDisplayMode(m_window, &mode);
-    mode.format = SDL_PIXELFORMAT_RGBA8888;
-    // Apply the pixel format to the device context
+        mode.format = SDL_PIXELFORMAT_RGBA8888;
+        // Apply the pixel format to the device context
```
Xottab-DUTY, line 86:
```suggestion
    mode.format = SDL_PIXELFORMAT_RGBA8888;
    // Apply the pixel format to the device context
```

src/Layers/xrRenderGL/glHW.cpp
```diff
@@ -103,6 +104,7 @@ void CHW::CreateDevice(SDL_Window* hWnd)
         const Uint32 flags = SDL_WINDOW_BORDERLESS | SDL_WINDOW_HIDDEN | SDL_WINDOW_OPENGL;
 
         m_helper_window = SDL_CreateWindow("OpenXRay OpenGL helper window", 0, 0, 1, 1, flags);
+
```
Xottab-DUTY, line 107:
```suggestion
```

src/Layers/xrRenderGL/glHW.h
```diff
@@ -64,6 +64,9 @@ class CHW
     bool SeparateShaderObjectsSupported;
     bool ShaderBinarySupported;
     bool ComputeShadersSupported;
+    bool GLKHRdebugSupported;
+    bool GLARBtexturefloatSuppoted;
+    bool GLARBvertexattribbindingSupported;
```
Xottab-DUTY, line 69:
```suggestion
    bool DebugSupported;
    bool TextureFloatSuppoted;
    bool VertexAttribBindingSupported;
```

src/Layers/xrRenderGL/glState.h
```diff
@@ -36,6 +36,7 @@ class glState
     D3D_DEPTH_STENCIL_STATE m_pDepthStencilState;
     D3D_BLEND_STATE m_pBlendState;
     float m_uiMipLODBias;
+    bool GLEXTtexturefilteranisotropicSupported;
```
Xottab-DUTY, line 39: Let's move it to `HW` with `TextureFilterAnisotropicSupported` name.
It's generally better to keep such stuff in one place, and also `glState` can be created on-demand during the game run – that means it would always call `epoxy_has_gl_extension` for each glState instance (there can be many of them) which is suboptimal.

src/Layers/xrRenderGL/glBufferUtils.cpp
```diff
@@ -162,14 +162,14 @@ void SetVertexDeclaration(const VertexElement* dxdecl)
     });
 }
 
-void ConvertVertexDeclaration(const VertexElement* dxdecl, SDeclaration* decl)
+void ConvertVertexDeclaration(const VertexElement* dxdecl, SDeclaration* decl, bool GLARBvertexattribbindingSupported)
```
Xottab-DUTY, line 165: This looks less good.
It's better to just use `HW.GLARBvertexattribbindingSupported` (after renaming it should be `HW.VertexAttribBindingSupported`) directly.

src/Layers/xrRender/ShaderResourceTraits.h
```diff
@@ -61,7 +61,7 @@ inline std::pair<char, GLuint> GLCompileShader(pcstr* buffer, size_t size, pcstr
 
     const GLuint program = glCreateProgram();
     R_ASSERT(program);
-    if (GLEW_VERSION_4_3)
+    if (epoxy_gl_version() >= 43)
```
Xottab-DUTY, line 64: Let's check for `HW.OpenGLVersion` instead of calling external function.

cmake/Findepoxy.cmake
Xottab-DUTY, line 1: Please, try to use this find module: https://github.com/WebKit/WebKit/blob/main/Source/cmake/FindEpoxy.cmake
It provides `Epoxy::Epoxy` CMake target which can be used.

src/Layers/xrRenderPC_GL/CMakeLists.txt
```diff
@@ -396,6 +396,7 @@ target_include_directories(xrRender_GL
     "${CMAKE_SOURCE_DIR}/Externals"
     "${CMAKE_SOURCE_DIR}/Externals/gli"
     "${CMAKE_SOURCE_DIR}/Externals/gli/external"
+    "${epoxy_INCLUDE_DIR}"
```
Xottab-DUTY, line 399: Can be removed after switching to `Epoxy::Epoxy` (https://github.com/OpenXRay/xray-16/pull/1681#discussion_r1615311296).

src/Layers/xrRenderPC_GL/CMakeLists.txt
```diff
@@ -408,8 +409,7 @@ target_link_libraries(xrRender_GL
     xrParticles
     xrScriptEngine
     xrImGui
-    OpenGL::GL
-    GLEW::GLEW
+    "${epoxy_LIBRARY}"
```
Xottab-DUTY, line 412: Try to switch to `Epoxy::Epoxy`.

Xottab-DUTY (changes requested):
Thanks!
I've left mostly code style and some other small improvement request!

AMS21:
> > libepoxys last release is 2 years old so why not use the still actively developed [glad](https://github.com/Dav1dde/glad) library, which also supports Vulkan out of the box?
> 
> GLEW's last release is 3 years old :)
> 
> Though, there's a bad sign there: libepoxy doesn't have any fresh commits since the last release, while glew has.



> > libepoxys last release is 2 years old so why not use the still actively developed [glad](https://github.com/Dav1dde/glad) library, which also supports Vulkan out of the box?
> 
> GLEW's last release is 3 years old :)
> 
> Though, there's a bad sign there: libepoxy doesn't have any fresh commits since the last release, while glew has.

Maybe I'm being missunderstood here I'm all for replacing the ancient GLEW library. My point was rather to maybe consider using the glad library I linked which latest release is about 1 Month old and still actively developed while also supporting all OpenGL extensions as well as Vulkan (which we might want to use in the future). Also glad directly parses the official OpenGL and Vulkan spec to generate it's glue code so even if the library is ever abandoned the parser should still be able to generate bindings for never versions of OpenGL and Vulkan (at least untill they make fundamental changes to the spec).

Xottab-DUTY:
@AMS21 the @sobkas's concern is this:
> GLAD2 depends on generated code and this code is separated for each API.
So for each API there are files(*.c and *.h) that would be included into OpenXRay.
Epoxy needs only #include to be added.

I also had a concern that if you select OpenGL 4.6, then you won't be able to use e.g. OpenGL 3.3 or OpenGL 2.0.
But I just checked the generated files and found that we were (or I was) wrong.

GLAD2 supports working with multiple APIs simultaneously. On it's website you just need to select maximal OpenGL version and any lower version would be included.
It also support header-only option.

Xottab-DUTY:
So main concern with libepoxy is that it's author hadn't been active for two years. No commits, not even PR merges, etc.
It also doesn't have simple global variables indicating if an extension is available. Honestly, they are pretty handy to use.

Xottab-DUTY:
Superseded with #1684.

Issue #280, opened 2018-11-21
OpenGL Renderer


Xottab-DUTY:
Post here all issues you have with OpenGL Renderer.

Current tasks and known bugs:
- [x] Fix crash after second Device restart (type `vid_restart` several times in console to reproduce) (e9427ac939862e103e513a614ee1f9e7ae3c8df4)
- [x] Fix bugs with dynamic lighting
  - [x] Fix bugs with local lights (lamps, campfires, etc) (06bb7657fad7d6578820cdfbf4aa7701ee732496, 615152e548a309d4dd4df3d7beb982f18af7bae5)
- [x] Rewrite sun lighting calculation from D3DXMath (#463, intersects with #261)
- [ ] Backend
  - [ ] Supported OpenGL versions:
    - [x] 4.5
    - [x] 4.1
    - [ ] 3.3
      - [x] Implement shader pipeline that can work without `GL_ARB_separate_shader_objects` (b0e4de824046f77b9de378ea5d6f3de4e53b22a4)
        -   [Useful link to OpenGL Wiki](https://www.khronos.org/opengl/wiki/Shader_Compilation#Separate_programs)
    - [ ] 3.1/3.0 maybe?
    - [ ] 2.2 maybe?
    - [ ] 2.0 maybe?
  - [ ] Implement OpenGL ES support
    - [ ] 3.x (make initial implementation for ES 3.x, if going straight to 2.0 isn't possible)
    - [ ] 2.0 (lower requirement to 2.0, if initial implementation was for 3.x)
  - [ ] Support for uniform buffer objects
  - [ ] Support MSAA
  - [ ] Support geometry shaders
  - [ ] Support compute shaders
  - [ ] Support tesselation shaders
- [ ] Shaders ~~(two ways, should be supported both)~~
  - [ ] Use cross-compiler to compile shaders from DirectX renderer
    - Look into #258 for more info
  - [ ] ~~Implement all shaders in GLSL (`shaders/gl` folder)~~
	- [x] Implement vertex and fragment (pixel) shaders
      - [ ] Make implemented shaders work in all situations
        - [x] Fix *boxy* sky textures with enabled Vertex Texture Fetch (#537)
        - [x] Fix lighting glitches with G-buffer optimization enabled (`r3_gbuffer_opt on`)
		- [ ] Fix SSAO
        - [ ] Fix HBAO shader
	- [ ] ~~MSAA support~~
    - [ ] ~~Implement volumetric fog~~
[...]

PR #1684, opened 2024-05-29
Replace GLEW with GLAD

Xottab-DUTY:
Benefits:
* OpenGL 1.0 to 4.6
* OpenGL ES 1.0 to 3.2
  * Possibility to implement OpenGL ES support: https://github.com/OpenXRay/xray-16/issues/280#issuecomment-2136446760
* Handles function aliases
* One header file
* Carries no dependency (except for user provided function to load OpenGL extension. We use `SDL_GL_GetProcAddress`)
  * GLAD sources are integrated into our build system. This may be considered as disadvantage, but it gives us more portability.
  * Removes dependency from GLEW, which may be missing on some platforms (e.g. latest 32-bit Ubuntu versions)

The generated glad.h includes all available extensions, including `GL_ARB_separate_shader_objects`, but excluding `GL_EXT_separate_shader_objects` due to a conflict. I decided that it would be best to not include it at this moment (if ever).
Read more here: https://community.khronos.org/t/confused-gl-active-program-gl-current-program/65477.


Xottab-DUTY:
@sobkas, @AMS21, please look at it! Does it work under Linux?

AMS21:
> @sobkas, @AMS21, please look at it! Does it work under Linux?

Compiled with gcc 14.1.0 and seems to work fine on my machine :)

alsed:
Hi @Xottab-DUTY  

I'm reading [here](https://bcmpinc.wordpress.com/2015/08/18/creating-an-opengl-4-5-context-using-sdl2-and-glad/)  that in order to have OpenGL functions available it's required to link them through `gladLoadGLLoader`

```
gladLoadGLLoader(SDL_GL_GetProcAddress);
// call OpenGl functions
```
I can not find `gladLoadGLLoader` in the generated glad.h, it's there a way to include it?

I suspect this can fix some inconsistent segfaults in `CreateDevice()` ~in #1761~

Xottab-DUTY:
`gladLoadGLLoader` is an optional function and `SDL_GL_CreateContext` which crashes (according to the info in #1761) is unrelated to GLAD.

PR #1741, opened 2024-12-10
Fixes for trader

Hrusteckiy:
1) disabled Trader's head rotation when dist to actor is more than 20 meters
2) Trader's head rotation limited by 1 radian to left and right from center
3) sounds from Trader will be from his head instead of playing in "player's head"
4) disabled slots stuffing for Trader


Xottab-DUTY:
Superseded by #1742

PR #1745, opened 2024-12-16
SoC Fixes

APAmk2:
This PR Fixes wrong zoom position loading for SoC HUD models & fixes crash when engine trying to find bone with wrong name for grenade launchers.
Also fixes loading of some UI XMLs.

### DIFF
--- a/src/xrGame/Weapon.cpp
+++ b/src/xrGame/Weapon.cpp
@@ -35,6 +35,8 @@
 constexpr pcstr WPN_SCOPE = "wpn_scope";
 constexpr pcstr WPN_SILENCER = "wpn_silencer";
 constexpr pcstr WPN_GRENADE_LAUNCHER = "wpn_launcher";
+// SoC Support - APAMk2
+constexpr pcstr WPN_GRENADE_LAUNCHER_SOC = "wpn_grenade_launcher";
 
 BOOL b_toggle_weapon_aim = FALSE;
 
@@ -1309,6 +1311,11 @@ void CWeapon::UpdateHUDAddonsVisibility()
 
     //.	return;
 
+    u16 bone_id;
+    bone_id = HudItemData()->m_model->LL_BoneID(wpn_grenade_launcher);
+    if (bone_id == BI_NONE && ShadowOfChernobylMode)
+        wpn_grenade_launcher = WPN_GRENADE_LAUNCHER_SOC;
+
     if (ScopeAttachable())
     {
         HudItemData()->set_bone_visible(wpn_scope, IsScopeAttached());
@@ -1401,6 +1408,9 @@ void CWeapon::UpdateAddonsVisibility()
     }
 
     bone_id = pWeaponVisual->LL_BoneID(wpn_grenade_launcher);
+    if (bone_id == BI_NONE && ShadowOfChernobylMode)
+        wpn_grenade_launcher = WPN_GRENADE_LAUNCHER_SOC;
+
     if (GrenadeLauncherAttachable())
     {
         if (IsGrenadeLauncherAttached())
--- a/src/xrGame/Weapon.cpp
+++ b/src/xrGame/Weapon.cpp
@@ -1303,16 +1303,13 @@ void CWeapon::UpdateHUDAddonsVisibility()
     static shared_str wpn_scope = WPN_SCOPE;
     static shared_str wpn_silencer = WPN_SILENCER;
     static shared_str wpn_grenade_launcher = WPN_GRENADE_LAUNCHER;
+    static shared_str wpn_grenade_launcher_soc = WPN_GRENADE_LAUNCHER_SOC;
 
     // actor only
     if (!GetHUDmode())
         return;
 
     //.	return;
-
-    u16 bone_id = HudItemData()->m_model->LL_BoneID(wpn_grenade_launcher);
-    if (bone_id == BI_NONE)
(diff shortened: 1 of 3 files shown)


src/xrUICore/XML/UIXmlInitBase.cpp
```diff
@@ -742,7 +742,7 @@ bool CUIXmlInitBase::InitTabControl(CUIXml& xml_doc, pcstr path,
         newButton->m_btn_id = xml_doc.ReadAttrib("button", i, "id");
         if (!newButton->m_btn_id.size())
         {
-            R_ASSERT4(defaultIdsAllowed, "Tab control tab doesn't have 'id' assigned.", xml_doc.m_xml_file_name, path);
+            R_ASSERT4(ShadowOfChernobylMode || defaultIdsAllowed, "Tab control tab doesn't have 'id' assigned.", xml_doc.m_xml_file_name, path);
```
Xottab-DUTY, line 745: This change effectively disables the crash, but doesn't fix the problem which this crash is indicating.
Tabs in PDA still don't have IDs and don't work.

src/xrGame/Weapon.cpp
```diff
@@ -35,6 +35,8 @@
 constexpr pcstr WPN_SCOPE = "wpn_scope";
 constexpr pcstr WPN_SILENCER = "wpn_silencer";
 constexpr pcstr WPN_GRENADE_LAUNCHER = "wpn_launcher";
+// SoC Support - APAMk2
+constexpr pcstr WPN_GRENADE_LAUNCHER_SOC = "wpn_grenade_launcher";
```
Xottab-DUTY: You will be in the project readme, contributors list and commit history, mentioning your changes using comments in the code is outdated practice :(

So this is better:
```suggestion
constexpr pcstr WPN_GRENADE_LAUNCHER_SOC = "wpn_grenade_launcher";
```

src/xrGame/Weapon.cpp
```diff
@@ -1309,6 +1311,11 @@ void CWeapon::UpdateHUDAddonsVisibility()
 
     //.	return;
 
+    u16 bone_id;
+    bone_id = HudItemData()->m_model->LL_BoneID(wpn_grenade_launcher);
+    if (bone_id == BI_NONE && ShadowOfChernobylMode)
+        wpn_grenade_launcher = WPN_GRENADE_LAUNCHER_SOC;
```
Xottab-DUTY: We try to minimize usage of `<Game>Mode` variables and add support for features from SOC/CS/COP simultaneously.

Add this to the static shared_str's code block:
```cpp
static shared_str wpn_grenade_launcher_soc = WPN_GRENADE_LAUNCHER_SOC;
```
And then change the code to this:
```suggestion
    u16 bone_id = HudItemData()->m_model->LL_BoneID(wpn_grenade_launcher);
    if (bone_id == BI_NONE)
        bone_id = HudItemData()->m_model->LL_BoneID(wpn_grenade_launcher_soc);
```

src/xrGame/Weapon.cpp
```diff
@@ -1401,6 +1408,9 @@ void CWeapon::UpdateAddonsVisibility()
     }
 
     bone_id = pWeaponVisual->LL_BoneID(wpn_grenade_launcher);
+    if (bone_id == BI_NONE && ShadowOfChernobylMode)
+        wpn_grenade_launcher = WPN_GRENADE_LAUNCHER_SOC;
```
Xottab-DUTY: The changes here should be the same as in the above comment for `UpdateHUDAddonsVisibility` function.

Xottab-DUTY (changes requested):
There are some review comments to look at

But anyway, thanks so much!

src/xrGame/Weapon.cpp
```diff
@@ -1309,6 +1310,10 @@ void CWeapon::UpdateHUDAddonsVisibility()
 
     //.	return;
 
+    u16 bone_id = HudItemData()->m_model->LL_BoneID(wpn_grenade_launcher);
+    if (bone_id == BI_NONE)
+        wpn_grenade_launcher = WPN_GRENADE_LAUNCHER_SOC;
```
Xottab-DUTY: Забыл здесь строку тоже поправить
```suggestion
        bone_id = pWeaponVisual->LL_BoneID(wpn_grenade_launcher_soc);
```

src/xrGame/Weapon.cpp
```diff
@@ -1332,16 +1334,18 @@ void CWeapon::UpdateHUDAddonsVisibility()
     else if (m_eSilencerStatus == ALife::eAddonPermanent)
         HudItemData()->set_bone_visible(wpn_silencer, TRUE, TRUE);
 
+    bool useSoCName = HudItemData()->m_model->LL_BoneID(wpn_grenade_launcher) == BI_NONE;
```
Xottab-DUTY, line 1337: 1. It's more consistent to use snake_case for the variable name here.
2. Use SOC name only if COP is unavailable **and** SOC is available (both conditions).
Because if we will use SOC name when it's unavailable, that might introduce confusion in the log. COP name is preferred.
```suggestion
    bool use_soc_name{};
    if (HudItemData()->m_model->LL_BoneID(wpn_grenade_launcher) == BI_NONE)
        use_soc_name = HudItemData()->m_model->LL_BoneID(wpn_grenade_launcher_soc) != BI_NONE;
```

And, yeah, I know the suggested code looks too long and ugly maybe... :(

PR #1757, opened 2024-12-20
Cleaned up some redundant/unused code in xr_object_list

yohjimane:


Xottab-DUTY:
This code is actually correct, not redundant and it is used for a purpose:
m_primary_crows can be changed during SingleUpdate (it calls UpdateCL on an entity, which can call `o_crow`, which will add the object to m_primary_crows, possibly resulting in reallocation), which can result in iterators invalidation, making it crash because range-based for contains old iterators.
If it won't reallocate and won't crash, it still will be doing unwanted actions: calling UpdateCL on objects that were just added to m_primary_crows. In original code, this would be done only in the next frame.

PR #1763, opened 2025-01-05
[CoC] Add HUD animation speed parameter support in ltx files.

Neloreck:
Still testing it while integrating some weapons from CoC/anomaly. Without it game crashes on assertions / has de-synced sounds.

## Changes

- Adding support of third parameter in weapon hud sections, CoC and anomaly uses it everywhere

## Examples

```
anm_idle = protecta_idle
anm_idle_aim = protecta_aim
anm_idle_aim_moving = protecta_move, idle, 0.75
anm_idle_aim_moving_crouch = protecta_move, idle, 0.7
anm_idle_moving = protecta_move
anm_idle_sprint = protecta_sprint
```

```
anm_idle_moving_crouch_w_gl_aim = sg550_crouch_w_gl
anm_idle_moving_g = sg550_move_g
anm_idle_moving_g_aim = sg550_move_g, idle, 0.75
anm_idle_moving_w_gl = sg550_move_w_gl
anm_idle_moving_w_gl_aim = sg550_move_w_gl, idle, 0.75
```

## Links

https://github.com/revolucas/CoC-Xray/blob/551dfba681928b7a1292949cb2156c24174d96d6/src/xrGame/player_hud.cpp#L77
https://github.com/revolucas/CoC-Xray/blob/551dfba681928b7a1292949cb2156c24174d96d6/src/xrGame/player_hud.cpp#L352

### DIFF
--- a/src/xrGame/player_hud.cpp
+++ b/src/xrGame/player_hud.cpp
@@ -24,12 +24,14 @@ constexpr float TENDTO_SPEED_AIM  = 8.f;    // (Для прицеливания)
 // --#SM+# End--
 // clang-format on
 
-float CalcMotionSpeed(const shared_str& anim_name)
+float CalcMotionSpeed(const shared_str& anim_name, const float anim_speed)
 {
-    if (!IsGameTypeSingle() && (anim_name == "anm_show" || anim_name == "anm_hide"))
-        return 2.0f;
+    // Apply custom animation speeds / configuration only for singleplayer games.
+    // Fast reloading / showing / hiding animation does not seem fair.
+    if (IsGameTypeSingle())
+        return anim_speed;
     else
-        return 1.0f;
+        return (anim_name == "anm_show" || anim_name == "anm_hide") ? 2.0f : 1.0f;
 }
 
 const player_hud_motion* player_hud_motion_container::find_motion(const shared_str& name) const
@@ -428,7 +430,7 @@ u32 attachable_hud_item::anim_play(const shared_str& anm_name_b, BOOL bMixIn, co
                                             m_visual_name.c_str(), anim_name_r)
                                             .c_str());
 
-    const float speed = anm->m_anim_speed;
+    const float speed = CalcMotionSpeed(anm->m_base_name, anm->m_anim_speed);
 
     rnd_idx = (u8)Random.randI(anm->m_animations.size());
     const motion_descr& M = anm->m_animations[rnd_idx];
@@ -628,7 +630,7 @@ void player_hud::render_hud(u32 context_id, IRenderable* root)
 
 u32 player_hud::motion_length(const shared_str& anim_name, const shared_str& hud_name, const CMotionDef*& md)
 {
-    const float speed = CalcMotionSpeed(anim_name);
+    const float speed = CalcMotionSpeed(anim_name, 1.0f);
     attachable_hud_item* pi = create_hud_item(hud_name);
     const player_hud_motion* pm = pi->m_hand_motions.find_motion(anim_name);
 
--- a/src/xrGame/player_hud.cpp
+++ b/src/xrGame/player_hud.cpp
@@ -63,7 +63,7 @@ void player_hud_motion_container::load(IKinematicsAnimated* model, const shared_
                 pm.m_base_name = str_item;
 
                 _GetItem(anm.c_str(), 1, str_item);
-                pm.m_additional_name = str_item;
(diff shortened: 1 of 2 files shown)


src/xrGame/player_hud.cpp
```diff
@@ -53,16 +53,20 @@ void player_hud_motion_container::load(IKinematicsAnimated* model, const shared_
...
                 string512 str_item;
                 _GetItem(anm.c_str(), 0, str_item);
                 pm.m_base_name = str_item;
 
                 _GetItem(anm.c_str(), 1, str_item);
-                pm.m_additional_name = str_item;
+                pm.m_additional_name = (strlen(str_item) > 0) ? pm.m_additional_name = str_item : pm.m_base_name;
+
+                _GetItem(anm.c_str(), 2, str_item);
+                pm.m_anim_speed = strlen(str_item) > 0 ? atof(str_item) : 1.f;
```
Xottab-DUTY:
```suggestion
                pm.m_anim_speed = xr_strlen(str_item) > 0 ? atof(str_item) : 1.f;
```

src/xrGame/player_hud.cpp
```diff
@@ -53,16 +53,20 @@ void player_hud_motion_container::load(IKinematicsAnimated* model, const shared_
             {
                 pm.m_base_name = anm;
                 pm.m_additional_name = anm;
+                pm.m_anim_speed = 1.f;
             }
             else
             {
-                R_ASSERT2(_GetItemCount(anm.c_str()) == 2, anm.c_str());
+                R_ASSERT2(_GetItemCount(anm.c_str()) <= 3, anm.c_str());
```
Xottab-DUTY, line 62: The code below requires at least one parameter, and uses additional two.
It's more logical to use this condition:
```suggestion
                R_ASSERT2(_GetItemCount(anm.c_str()) > 0, anm.c_str());
```

src/xrGame/player_hud.cpp
```diff
@@ -53,16 +53,20 @@ void player_hud_motion_container::load(IKinematicsAnimated* model, const shared_
...
                 string512 str_item;
                 _GetItem(anm.c_str(), 0, str_item);
                 pm.m_base_name = str_item;
 
                 _GetItem(anm.c_str(), 1, str_item);
-                pm.m_additional_name = str_item;
+                pm.m_additional_name = (strlen(str_item) > 0) ? pm.m_additional_name = str_item : pm.m_base_name;
```
Xottab-DUTY: The code `pm.m_additional_name = pm.m_additional_name = str_item` looks bad.
I believe, it should be this:
```suggestion
                pm.m_additional_name = xr_strlen(str_item) > 0 ? str_item : pm.m_base_name;
```

src/xrGame/player_hud.cpp
```diff
@@ -414,8 +418,6 @@ void attachable_hud_item::reload_measures()
 
 u32 attachable_hud_item::anim_play(const shared_str& anm_name_b, BOOL bMixIn, const CMotionDef*& md, u8& rnd_idx)
 {
-    const float speed = CalcMotionSpeed(anm_name_b);
```
Xottab-DUTY, line 417: Can we reuse this somehow? We still support multiplayer.

Xottab-DUTY (changes requested):

src/xrGame/player_hud.cpp
Neloreck: Logics before:
1 parameter - use name as base and additional parameter
2+ parameters - assert it is only 2 params and read both

Logics after:
1 parameter - same as before
2+ parameters - make sure it is either 2 or 3 parameters, if someone leaves empty space and wants to only add animation speed

---

Seems like:
1 - adding < 0 assertion and storing count as variant makes sense
2 - probably I should add trimming for pm.m_additional_name strlen check? Or parser of comma separated values already does it? Will need to check it

src/xrGame/player_hud.cpp
Neloreck: Yeah, copy-pasted it with minimal changes. Not sure about ret value from assignment operation

src/xrGame/player_hud.cpp
Neloreck: This util used for two different classes. Can add one more param to use as base and keep it pure function or copy-paste as separate methods/util functions. Sounds like first variant is better

Neloreck: Updated util / excluded non-singleplayer games from animation speed calculations.

Xottab-DUTY (approved):

PR #1767, opened 2025-01-11
Segfaults fixes on OpenGL Render startup

### LINKED ISSUE #1761: [linux] Segfault when starting OpenGL render device
**Describe the bug**
A segmentation fault occurs when starting the engine, specifically when tries to start the render device (openGL).
The problem is not consistent, happens most of the time, but not always.
Occurs in XrCore.so

**To Reproduce**
1. Compile rpm package in debug mode.
2. Install.
3. Execute `xd_3da`  (on linux and default openGL render)

**Expected behavior**
Engine starts

**Screenshots, videos**
N/A

**BugTrap error report**

```
[...]
Starting RENDER device...
 
FATAL ERROR
 
[error] Expression    : <no expression>
[error] Function      : handler_base
[error] File          : /home/alsed/Descargas/xray-16/src/xrCore/xrDebug.cpp
[error] Line          : 860
[error] Description   : segmentation fault
 

stack trace:

xrDebug::GatherInfo(char*, unsigned long, ErrorLocation const&, char const*, char const*, char const*, char const*)
xrDebug::Fail(bool&, ErrorLocation const&, char const*, char const*, char const*, char const*)
/lib64/xrCore.so(+0x1ab2b) [0x7f0123348b2b]
/lib64/xrCore.so(+0x1ab6a) [0x7f0123348b6a]
/lib64/libc.so.6(+0x1a090) [0x7f0122826090]
MESA-INTEL: warning: ../src/intel/vulkan/anv_formats.c:763: FINISHME: support YUV colorspace with DRM format modifiers
MESA-INTEL: warning: ../src/intel/vulkan/anv_formats.c:794: FINISHME: support more multi-planar formats with DRM modifiers

```

 - OS: Fedora 41
 - OpenXRay dev version

**Additional context**
- Engine compiled in debug mode.

When debugging, I get this info:

```
#0  __GI_getenv (name=name@entry=0x7ffff4182d1c "EGL_KHR_create_context") at getenv.c:50

#1  0x00007ffff40dc510 in SDL_getenv_REAL (name=0x7ffff4182d1c "EGL_KHR_create_context") at /usr/src/debug/SDL2-2.30.9-1.fc41.x86_64/src/stdlib/SDL_getenv.c:183

#2  SDL_EGL_HasExtension (_this=_this@entry=0x437ed0, type=type@entry=SDL_EGL_DISPLAY_EXTENSION, ext=ext@entry=0x7ffff4182d1c "EGL_KHR_create_context")
    at /usr/src/debug/SDL2-2.30.9-1.fc41.x86_64/src/video/SDL_egl.c:193

#3  0x00007ffff40df489 in SDL_EGL_CreateContext (_this=0x437ed0, egl_surface=0x1c22440) at /usr/src/debug/SDL2-2.30.9-1.fc41.x86_64/src/video/SDL_egl.c:994

#4  0x00007ffff412d811 in Wayland_GLES_CreateContext (_this=0x437ed0, window=<optimized out>) at /usr/src/debug/SDL2-2.30.9-1.fc41.x86_64/src/video/wayland/SDL_waylandopengles.c:54

#5  0x00007ffff40ec615 in SDL_GL_CreateContext_REAL (window=0x4b58b0) at /usr/src/debug/SDL2-2.30.9-1.fc41.x86_64/src/video/SDL_video.c:4094

#6  0x00007fffbf25d3b0 in CHW::CreateDevice(SDL_Window*) () from /lib64/xrRender_GL.so
```

- Only function 6, `CHW::CreateDevice(SDL_Window*) () ` is from the engine, 5 to 0 are all SDL internal functions.
- The problem seems to occur because a memory address is being passed as if were a valid pointer to a char. This occurs in `SDL_getenv_REAL`, which is receiving `name=0x7ffffff4182d1c “EGL_KHR_create_context”`. `SDL_getenv` expects a `const char *` as a parameter. 
- I suspect somewhere in the stack the pointer gets invalid or corrupted.

alsed:
- fix segfault for main context creation
- fix segfaults when calling openGl functions
- resolves #1761

### DIFF
--- a/src/Layers/xrRenderGL/glHW.cpp
+++ b/src/Layers/xrRenderGL/glHW.cpp
@@ -86,7 +86,7 @@ void CHW::CreateDevice(SDL_Window* hWnd)
 
     // Create the context
     m_context = nullptr; // Clean up first
-    if (! (m_context = SDL_GL_CreateContext(m_window)));
+    if (! (m_context = SDL_GL_CreateContext(m_window)))
     {
         Log("! Could not create drawing context:", SDL_GetError());
         return;
--- a/src/Layers/xrRenderGL/glHW.cpp
+++ b/src/Layers/xrRenderGL/glHW.cpp
@@ -86,7 +86,7 @@ void CHW::CreateDevice(SDL_Window* hWnd)
 
     // Create the context
     m_context = nullptr; // Clean up first
-    if (! (m_context = SDL_GL_CreateContext(m_window));
+    if (! (m_context = SDL_GL_CreateContext(m_window)));
     {
         Log("! Could not create drawing context:", SDL_GetError());
         return;
--- a/src/Layers/xrRenderGL/glHW.cpp
+++ b/src/Layers/xrRenderGL/glHW.cpp
@@ -75,7 +75,6 @@ void CHW::CreateDevice(SDL_Window* hWnd)
     ZoneScoped;
 
     m_window = hWnd;
-
     R_ASSERT(m_window);
 
     // Choose the closest pixel format
@@ -86,8 +85,8 @@ void CHW::CreateDevice(SDL_Window* hWnd)
     SDL_SetWindowDisplayMode(m_window, &mode);
 
     // Create the context
-    m_context = SDL_GL_CreateContext(m_window);
-    if (m_context == nullptr)
+    m_context = nullptr; // Clean up first
+    if (! (m_context = SDL_GL_CreateContext(m_window));
     {
         Log("! Could not create drawing context:", SDL_GetError());
         return;
@@ -123,6 +122,12 @@ void CHW::CreateDevice(SDL_Window* hWnd)
 
     UpdateVSync();
 
+    // Initialize GLAD
+    if ( (gladLoadGL((GLADloadfunc) SDL_GL_GetProcAddress)) == 0)
+    {
+        Log ("! Could not intialize GLAD", SDL_GetError());
+    }
+    
 #ifdef DEBUG
     if (glDebugMessageCallback)
     {


src/Layers/xrRenderGL/glHW.cpp
```diff
@@ -86,8 +85,8 @@ void CHW::CreateDevice(SDL_Window* hWnd)
     SDL_SetWindowDisplayMode(m_window, &mode);
 
     // Create the context
-    m_context = SDL_GL_CreateContext(m_window);
-    if (m_context == nullptr)
+    m_context = nullptr; // Clean up first
+    if (! (m_context = SDL_GL_CreateContext(m_window)));
```
Xottab-DUTY: 1. The `m_context` variable isn't passed to `SDL_GL_CreateContext` function and this function has a return meaning that it will always write to `m_context`, so there's no need to make it `nullptr`.
2. Moreover, the `;` effectively disabled the if and made the error handling code below always enabled, meaning that it would always write `! Could not create drawing context:` to the log and return even if the context has been successfully created.
That also means all the code below like `MakeContextCurrent(IRender::PrimaryContext)` won't be called at all.

I suggest to restore the original code:
```suggestion
    m_context = SDL_GL_CreateContext(m_window);
    if (m_context == nullptr)
```

Xottab-DUTY: > Moreover, the `;` effectively disabled the if and made the error handling code below always enabled

Look at this example:
```cpp
if (condition)
    ; // do nothing
{
    // this code is NOT the part of the if,
    // it will always be executed.
    // it's just a free standing scope
    int i = 0;
    return;
}

// this code never will be executed
int k = 0;
```

src/Layers/xrRenderGL/glHW.cpp
```diff
@@ -123,6 +122,12 @@ void CHW::CreateDevice(SDL_Window* hWnd)
 
     UpdateVSync();
 
+    // Initialize GLAD
+    if ( (gladLoadGL((GLADloadfunc) SDL_GL_GetProcAddress)) == 0)
+    {
+        Log ("! Could not intialize GLAD", SDL_GetError());
+    }
```
Xottab-DUTY, line 129: GLAD is initialized in `Layers/xrRenderPC_GL/r2_test_hw.cpp`. It doesn't have a separate state to save, meaning that call to `gladLoadGL` is global and it's enough to call it once.
So there should be no point in calling it here twice.

Xottab-DUTY (changes requested):

src/Layers/xrRenderGL/glHW.cpp
alsed: A stupid mistake of my part, that ";" shouldn't be there in the first place.
Nevertheless, I understand that my "security" proposal shouldn't make much difference if a segfault occurs.

src/Layers/xrRenderGL/glHW.cpp
alsed: Mmm, could it be that `sdl_window_test_helper() `is creating conflicts with `CreateDevice() `?
I'm not talking about GLAD here, instead of the use of separately created `m_window` and `m_context` in both functions.

alsed: The problem with GLAD could be that the local `m_context`  gets destroyed in `sdl_window_test_helper() `, if this is so, then the functions loaded with GLAD are disabled. This would explain why should be loaded again in `CreateDevice()`.

The solution could be a global bool and a function to confirm that GLAD is loaded, and in case is not, load it again:
```
bool isGladLoaded = false;

void CheckGlad()
{
    if (!isGladLoaded)
    {
        Log("GLAD is not loaded. Attempting to load it...");

        if (gladLoadGL((GLADloadfunc) SDL_GL_GetProcAddress) == 0)
        {
            Log("! Error: Could not load GLAD. Ensure an OpenGL context is active.", SDL_GetError());
            return;
        }

        isGladLoaded = true;
        Log("GLAD successfully loaded.");
    }
    else
    {
        Log("GLAD is already loaded.");
    }
}

```

src/Layers/xrRenderGL/glHW.cpp
Xottab-DUTY: It doesn't add any security, because `SDL_GL_CreateContext` does always write to `m_context`, so this change makes code less readable for no purpose.

src/Layers/xrRenderGL/glHW.cpp
Xottab-DUTY: > The problem with GLAD could be that the local `m_context`  gets destroyed in `sdl_window_test_helper() `, if this is so, then the functions loaded with GLAD are disabled. This would explain why should be loaded again in `CreateDevice()`.

Yes, the local `m_context` gets destroyed in `~sdl_window_test_helper()`, but what's interesting is that GLAD doesn't have a direct dependency on OpenGL context, so I assumed it's okay... But I just looked at the [SDL_GL_GetProcAddress documentation](https://wiki.libsdl.org/SDL2/SDL_GL_GetProcAddress) and found the culprit: this function depends on the current context on some platforms, and don't on other platforms.
So, for sanity we should always reinitialize GLAD when changing contexts.

I refactored the testing and CHW code according to that. Please, test it with the latest 6df6613d197373cb167c34091f3c2e624533741b commit!

alsed: tested, it works!

Xottab-DUTY:
Superseded by 6df6613d197373cb167c34091f3c2e624533741b

PR #1768, opened 2025-01-12
[CoC/Anomaly] Optional grenade launcher switch animation usage.

Neloreck:
Just changing how animation is handled for grenade launcher toggling.

In case of missing animation just toggle switch. It is normal for some weapons with GL where switching mode does not require adjusting weapon / placing hand on another position. Found this one when was playing with integration of anomaly weapons and noticed some missing `anm_switch_g` `anm_switch` animation fields.

Main question as for now is logging -> probably prettier variant without method overloading can be implemented?

## Changes

- Just changing how animation is handled for grenade launcher toggling, allow cases without animation
- Still respect variant with two animations (SoC support?) and add comment for it
- Add param to prevent not_found log messages when doing animation checks / playing SoC animation variant (currently it produces 2 messages if animation is not found, I believe it is always 1 message for SoC)

## Links
https://github.com/xrLil-Batya/xray-monolith/blob/d404b95d92947e83b283d912111fe0957fb5fbb6/src/xrGame/WeaponMagazinedWGrenade.cpp#L839

https://github.com/OpenXRay/xray-16/pull/1763

#1373

### DIFF
--- a/src/xrGame/HudItem.h
+++ b/src/xrGame/HudItem.h
@@ -185,5 +185,7 @@ public:
     virtual CHudItem* cast_hud_item() { return this; }
     void PlayAnimIdleMovingCrouch(); //AVO: new crouch idle animation
     bool isHUDAnimationExist(pcstr anim_name) const;
+    bool isHUDAnimationExist(pcstr anim_name, bool is_silent) const;
     pcstr WhichHUDAnimationExist(pcstr anim_name, pcstr anim_name2) const;
+    pcstr WhichHUDAnimationExist(pcstr anim_name, pcstr anim_name2, bool is_silent) const;
 };
--- a/src/xrGame/HudItem.h
+++ b/src/xrGame/HudItem.h
@@ -184,8 +184,6 @@ public:
 
     virtual CHudItem* cast_hud_item() { return this; }
     void PlayAnimIdleMovingCrouch(); //AVO: new crouch idle animation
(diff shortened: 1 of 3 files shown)


src/xrGame/HudItem.h
```diff
@@ -185,5 +185,7 @@ class CHudItem : public CHUDState
     virtual CHudItem* cast_hud_item() { return this; }
     void PlayAnimIdleMovingCrouch(); //AVO: new crouch idle animation
     bool isHUDAnimationExist(pcstr anim_name) const;
+    bool isHUDAnimationExist(pcstr anim_name, bool is_silent) const;
     pcstr WhichHUDAnimationExist(pcstr anim_name, pcstr anim_name2) const;
+    pcstr WhichHUDAnimationExist(pcstr anim_name, pcstr anim_name2, bool is_silent) const;
```
Xottab-DUTY: This is better IMO:
```suggestion
    bool isHUDAnimationExist(pcstr anim_name, bool silent = false) const;
    pcstr WhichHUDAnimationExist(pcstr anim_name, pcstr anim_name2, bool silent = false) const;
```

Also notice that `is_silent` was renamed to `silent`. Not too much need in using `is` here.

Neloreck: Will update for all methods

Xottab-DUTY (approved):

Xottab-DUTY:
Thanks!

Issue #592, opened 2020-05-09
script var under "run_scripts" can sometimes be null


jfmherokiller:
**Describe the bug**
the script var under the function "run_scripts" can sometimes be null.

**To Reproduce**
1. 'enable the "run_script" command by commenting out the ifdefs'
2.  'change break_on_assert to false '
3. 'build release build and try to run something like 1+1'

**Expected behavior**
Id like to see "2" but no output is still good
**Screenshots**
N/A

**BugTrap error report**
N/A

**Desktop (please complete the following information):**
 - OS: [Windows 10]
 - OpenXRay build version [built from commit 31ce0e5cb1882c1029fee3f5bdf2b822af0e27a1]

**Additional context**

I developed a fix for this here 
```
        if (script != nullptr && script->active())
        {
            m_scripts.push_back(script);
        }
        else
        {
            if (script != nullptr)
            {
                xr_delete(script);
            }
        }
```

Neloreck:
cc @Xottab-DUTY 

Seems like expected behaviour, checked with latest build

run_script expects file name from scripts/ dir and tries to run file + call main() function in it
run_string executes argument as string, had no issues with running simple variants

Valid examples with game files that worked for me:
1) `run_script spawn_items` -> spawned items in actor inventory
2) `run_string log(1+1)` -> printed `[LUA]  2`

![Image](https://github.com/user-attachments/assets/af66b320-46d1-4e05-9b9b-fa540e472500)

Xottab-DUTY:
This was fixed in commit d877421ef4a0e08e21d726b0dbe5012f87a9f021.

PR #1769, opened 2025-01-12
[LUA] Use xr_new for script debug entities in stack

Neloreck:
Fixing issue with segfault when executing scripts from game console.
Simple 'spawn_items' script execution throws exceptions from destructor and point to implementation of memory management methods. I suspect that overloaded deletion/allocation methods should be used in pair to prevent crashes --- need confirmation here.

## Changes

- Using xr_new instead of new for stack entities memory allocation


Xottab-DUTY:
Do you know, if this is somehow fixes #592?
Could you also look at that issue too, if it's not too much trouble?

Neloreck:
@Xottab-DUTY Will check, investigating garbage in lua VM stack when enabling profiling hooks (call hooks specifically)

PR #1797, opened 2025-02-28
xrEngine/defines.cpp: use designated init

Honkazel:
https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-designated-initializers.html

> C++20 supports the designated initializer syntax for aggregate types. By applying it, we can always be sure that aggregates are constructed correctly, because every variable being initialized is referenced by its name.


Xottab-DUTY:
Thanks!

PR #1809, opened 2025-03-04
cmake: fix sanitizer builds missing link flags

AMS21:
Without adding the sanitizer linker flags you will get a linker error since the sanitizer runtime libraries are missing otherwise


Masterkatze (approved):
LGTM

Xottab-DUTY (approved):

Xottab-DUTY:
Thanks!

PR #1811, opened 2025-03-06
CMake: silence all warnings for external libraries

AMS21:
Noticed that during build the external libraries throw up a lot of warnings. Since we can't really fix them I think it would make sense to just silence them, so only the warnings from our actual code shows up

### DIFF
--- a/Externals/CMakeLists.txt
+++ b/Externals/CMakeLists.txt
@@ -5,10 +5,14 @@ else()
 endif()
 
 add_subdirectory(xrLuaFix)
+add_subdirectory(ode)
+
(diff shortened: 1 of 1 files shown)


Xottab-DUTY:
Let's not silence warnings for ODE and xrLuaFix:
1. We have really old ODE 0.5 (with small GSC original modifications and my edits to make it look a bit closer to 0.6 version). It can contain bugs which can affect the game, so we must actually fix it if any problems appear.
2. We can fix xrLuaFix, as it's our own module. (we might want to disable warnings for submodules inside https://github.com/OpenXRay/xrLuaFix.

PR #1812, opened 2025-03-06
xrGame: Bump duplicate phrase ids down to a warning instead of crashing

AMS21:
Apparently in Clear Sky this happens a lot even with the original game files.


Xottab-DUTY:
Nah, this happens even in COP with original gamedata :(

Xottab-DUTY:
Thanks!

PR #1815, opened 2025-03-07
fix: `SMusicTrack::SetVolume` setting the final volume to either 0.0 or 1.0

AMS21:
This was caused by what probably is a typo as the final volume is calculated as `volume * m_Volume` but then assigned to a bool (which C++ just allows you todo).

This was introduced in commit fa7af012e08622961d1c8d3a5d2044757f3d2f56.

Looking at the diff this is not the behaviour we had before.


Xottab-DUTY:
Thanks! Good catch!

PR #1851, opened 2025-04-15
Add Ukrainian translation of OpenXRay

YALdysse:
Hi. I added ukrainian translation of OpenXray text file. Please, accept pull request.

### DIFF
--- a/res/gamedata/configs/text/ukr/openxray.xml
+++ b/res/gamedata/configs/text/ukr/openxray.xml
@@ -1,0 +1,171 @@
+<?xml version="1.0" encoding="windows-1251" ?>
+<string_table>
+	<string id="st_unload_magazine_after_pickup">
+		<text>Розряд. зброю при піднятті</text>
+	</string>
+	<string id="st_unload_magazine_all">
+		<text>розрядити всю зброю</text>
+	</string>
+	<string id="st_unequip">
+		<text>сховати в рюкзак</text>
+	</string>
+	<string id="st_donate">
+		<text>Подарувати</text>
+	</string>
+	<string id="st_loading_lights">
+		<text>Завантаження освітлення</text>
+	</string>
+	<string id="st_client_connecting_to">
+		<text>Клієнт: З'єднання с</text>
+	</string>
+	<string id="st_client_spawning">
+		<text>Клієнт: Створення</text>
+	</string>
+	<string id="st_client_synchronising">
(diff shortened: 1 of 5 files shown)


res/gamedata/configs/text/ukr/openxray.xml
```diff
@@ -0,0 +1,171 @@
...
+		<text>�������� (Russian)</text>
+	</string>
+	<string id="spa">
+		<text>��������� (Spanish)</text>
+	</string>
+	<string id="ukr">
+		<text>��������� (Ukrainian)</text>
+	</string>
+	<string id="jpn">
+		<text>�������� (Japanese)</text>
```
kacejot, line 169: Do we have this in other locales? Where did it come from?

kacejot:
Maybe you also need to update other locales in [res/gamedata/configs/text](https://github.com/YALdysse/xray-16/tree/dev/res/gamedata/configs/text) by adding "Ukrainian" option there.

YALdysse:
> Maybe you also need to update other locales in [res/gamedata/configs/text](https://github.com/YALdysse/xray-16/tree/dev/res/gamedata/configs/text) by adding "Ukrainian" option there.

I want to translate the OpenXray text into Ukrainian rather than translate and correct all the locales.

kacejot:
> > Maybe you also need to update other locales in [res/gamedata/configs/text](https://github.com/YALdysse/xray-16/tree/dev/res/gamedata/configs/text) by adding "Ukrainian" option there.
> 
> I want to translate the OpenXray text into Ukrainian rather than translate and correct all the locales.

You're adding new option. You should be able to pick it from other languages. So I assume you need to add "Ukrainian" option to all existing locale files.

YALdysse:
> You're adding new option. You should be able to pick it from other languages. So I assume you need to add "Ukrainian" option to all existing locale files.

Done.

Xottab-DUTY:
Sorry for the delay, I wanted to correct some small details myself, but I decided to just merge that and correct all localizations in one additional commit (there's an update needed for all localizations).

res/gamedata/configs/text/ukr/openxray.xml
Xottab-DUTY: There is a Japan localization, along with Korean and Chinese... They aren't redistributed officially in Steam, GOG, though.
In fact, I'm not sure if they are official, but it's still nice to left this small addition :)

PR #1881, opened 2025-06-07
Implemented updated level.cform format with integrated CDB cache

SkyLoaderr:
Added support for updated level.cform format with integrated CDB cache.
Main differences:
- The cache is embedded in level.cform, now there is no need to check the checksum of the original level.cform (speeds up level loading a little)
- A list of physical materials used by the collision has been added to updated format. During the level loading process, the material identifiers are reassigned to the current gamemtl.xr file. Now gamemtl.xr changes do not require recompiling the level collision file
- New data is written to the end of the file, which will not break compatibility with engines that do not support this format. New data will simply be ignored
- level.cform and objspace.bin have duplicate vertex and triangle data. With embedded cache this duplication will not occur
- If we compare loading a level with and without a cache, loading a level without a cache takes longer, requires more RAM when building CDB tree, which is very noticeable in heavy locations

Intermediate version of level compilers with the [-cache_cform] option:
https://drive.google.com/file/d/1BN4_WTYdD3zJHKLMdV6IXnMHxYiJPcGN/view


Xottab-DUTY:
> The cache is embedded in level.cform, now there is no need to check the checksum of the original level.cform (speeds up level loading a little)

I would like to highlight some details: it saves about 400 ms on my PC when loading Zaton, I'd call it insignificant yet valuable. The boost becomes sensitive and noticeable when you load really big locations like Living Zone. Imagine loading it from HDD – this feature would save seconds of loading time!

>  New data is written to the end of the file, which will not break compatibility with engines that do not support this format. New data will simply be ignored

This is genius! I've never checked that vertex and triangles data is duplicated. Really good catch.

Xottab-DUTY (approved):

PR #1932, opened 2025-09-01
Fixed loading of fast (shadow) geometry

SkyLoaderr:
This issue was the cause of bug with fake static shadows on some custom maps. Since the level compiler optimizes shadow geometry and may remove some invalid triangles, the game still required the same number of triangles as normal geometry.


Xottab-DUTY:
Thanks!

PR #1933, opened 2025-09-07
Huge script export refactoring

### LINKED ISSUE #210: Support static linking

Xottab-DUTY:
1. Automatic dependency management from the previous system made by @nitrocaster is improved even further: dependencies are directly linked at compile-time. It's faster and more robust.
2. Using DECLARE_SCRIPT_REGISTER_FUNCTION macros and classname::script_register function almost like in the original engine. Simplifies code merging from other projects, simplifies using OpenXRay for those who are new to our engine, but are familiar with other projects engines.
3. Removed cphysics_scripted wrappers because they are not needed.
> Physics classes are exported to Lua directly now, like in SOC/CS engines. Historical reference: in COP GSC has moved main part of the physics code from xrGame to a separate xrPhysics library and they needed to add wrappers for scripting functionality because script engine still was inside xrGame. Since the script engine in OpenXRay is contained in a separate library, we don't need any wrappers.
5. Entire engine can be linked statically now on Windows. (#210)


Xottab-DUTY:
One thing to notice: mixed delegates aren't added to the script export on Linux for some reason.
It's ok to skip that for now, because they are needed for multiplayer and we don't have multiplayer support on Linux, but we will need to fix that still.

PR #1939, opened 2025-09-12
Attemp to make openxray work with SOC

werasik2aa:
Attemp to run and make suitable for SOC game.


Xottab-DUTY:
Superseded with #1940.

PR #1953, opened 2025-09-23
Cmake support on windows, update clang-format

### LINKED ISSUE #1575: Windows development using CMake
**Is your feature request related to a problem? Please describe.**
Windows development is now only possible using Visual Studio. There's much better options like CLion for example. But currently CMake is only properly working on linux.

**Describe the solution you'd like**
I would suggest to move fully to CMake, even on Windows, because it will allow developers to choose IDE's they like. Who want to stay with VS - can still generate VS project for themselves.

Now I have to manually tune `CMakeLists.txt` to open in CLion without errors.

Just a side note:
I also thing SDL include path is not set anywhere in `CMakeLists.txt`, it's only luck that it works on linux.

yohjimane:
Successfully building on win11

tested with this command:

`cmake -S . -B build/vs2022 -G "Visual Studio 17 2022" -A x64 -DCMAKE_BUILD_TYPE=RelWithDebInfo`

then

`cmake --build build/vs2022 --config Mixed`

OR if you prefer to use visual studio:
open the build/vs2022/OpenXray.sln and build desired configuration


src/xrEngine/x_ray.cpp
```diff
@@ -29,8 +29,10 @@
 #ifdef XR_PLATFORM_WINDOWS
 #include <locale>
 
+#ifdef XR_PLATFORM_WINDOWS
```
Xottab-DUTY: You're making ifdef inside ifdef, effectively duplicating it

res/fsgame.ltx
```diff
@@ -10,7 +10,7 @@ $game_data$             = false| true|  $fs_root$|            gamedata\
 $game_ai$               = true|  false| $game_data$|          ai\
 $game_spawn$            = true|  false| $game_data$|          spawns\
 $game_levels$           = true|  false| $game_data$|          levels\
-$game_meshes$           = true|  true|  $game_data$|          meshes\|       *.ogf;*.omf|   Game Object files
+$game_meshes$           = true|  true|  $game_data$|          meshes\|       *.ogf;*.omf;*.ozzx|   Game Object files
```
Xottab-DUTY: fsgame.ltx contains unrelated changes

cmake/XRay.Configurations.cmake
```diff
@@ -18,6 +18,10 @@ if (is_multi_config)
         message(WARNING "CMAKE_BUILD_TYPE is ignored with multi-config generator.")
         unset(CMAKE_BUILD_TYPE)
     endif()
+    
+    if (WIN32)
+    	unset(CMAKE_DEFAULT_BUILD_TYPE)
```
Xottab-DUTY: Why is it being unset?

yohjimane:
```CMake Error:
  Generator
    Visual Studio 17 2022  does not support variable CMAKE_DEFAULT_BUILD_TYPE but it has been specified.
```

It is necessary to fix this error during initial cmake generation

Xottab-DUTY: Ah, it's better to unset it only for Visual Studio generator, because WIN32 can also be something else, not just the VS generator.

yohjimane: Okay, I'll get that updated! iirc Codex added some other WIN32 if blocks- should I update the rest of the WIN32 if blocks in this commit to be conditional on vs generator instead?

Xottab-DUTY: 1. No, in some places checking for WIN32 is really correct.
2. It's better to use generator expressions instead of simple `if`s.

Xottab-DUTY: After you've pushed [`908ab63` (#1953)](https://github.com/OpenXRay/xray-16/pull/1953/commits/908ab63ebdac772c6bb9859063d2686486e60b7c) I just realized that we don't need to unset it all. Just don't set it in the first place:
```cmake
    if (NOT CMAKE_DEFAULT_BUILD_TYPE AND NOT CMAKE_GENERATOR MATCHES "Visual Studio")
        set(CMAKE_DEFAULT_BUILD_TYPE ${XRAY_DEFAULT_BUILD_TYPE})
    endif()
```

yohjimane: Doy, that makes sense. Thanks for catching that, i've fixed it in the latest commit.

PR #1972, opened 2025-11-05
OpenGL does not support parallel draw calls

crashuxx:
It fixes broken "sun shadows"

### DIFF
--- a/src/Layers/xrRender_R2/r2_R_calculate.cpp
+++ b/src/Layers/xrRender_R2/r2_R_calculate.cpp
@@ -82,7 +82,12 @@ void CRender::Calculate()
     // Configure
     o.distortion    = o.distortion_enabled;
     o.mt_calculate  = ps_r2_mt_calculate > 0;
+#ifdef USE_OGL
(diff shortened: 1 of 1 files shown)


src/Layers/xrRender_R2/r2_R_calculate.cpp
```diff
@@ -82,7 +82,12 @@ void CRender::Calculate()
     // Configure
     o.distortion    = o.distortion_enabled;
     o.mt_calculate  = ps_r2_mt_calculate > 0;
+#ifdef USE_OGL
+    o.mt_render     = 0; // OpenGL does not support parallel draw calls
+#else
     o.mt_render     = ps_r2_mt_render > 0;
+#endif
+
```
Xottab-DUTY:
```suggestion
#ifdef USE_DX11
    o.mt_render     = ps_r2_mt_render > 0;
#else
    o.mt_render     = 0; // OpenGL does not support parallel draw calls
#endif
```

Xottab-DUTY:
Thanks!

PR #1986, opened 2025-11-11
Enable MSAA in OpenGL

crashuxx:
This enables the MSAA in the OpenGL renderer.
"Antialias A tested object" need to be disabled in game settings

~~for some reason there's no aa in object that collide with sky , but that can be addressed later~~

~~needs more testing and cleanup, but the core functionality is here~~


crashuxx:
I'm happy with the result I have here, there are still some rabbit holes that can be chased, I marked them with TODO for later

Tested with CS data on Linux AMD 
fmsaa1
fmsaa3

argosurvivors:
There is too much light and too much reflection on the ground plane on both examples, much more than in the original. Is it possible to adapt that in the next update?

PR #1991, opened 2025-11-14
fix: No upgrades being available in CoC technician dialog

### LINKED ISSUE #1852: Call of Chernobyl: No upgrade tiles on any weapon or outfit when talk to any technician
**Describe the bug**
On OpenXray engine in **Call of Chernobyl** there are no upgrade tiles available when you talk to any technician

**To Reproduce**
1. Go to technician and ask for upgrades.
2. Click on item which you want to upgrade. 
3. Upgrade tiles not exist.

**Expected behavior**
There should be tiles with all available upgrades for each weapon or outfit

**Screenshots, videos**

Image

 - OS: Windows 10
 - OpenXRay build version lastest-nightly 5a8cc8f

AMS21:
Fixes https://github.com/OpenXRay/xray-16/issues/1852

I can't test if this works with SoC since the game crashes on startup for me.

### DIFF
--- a/src/xrGame/ui/UITalkDialogWnd.cpp
+++ b/src/xrGame/ui/UITalkDialogWnd.cpp
@@ -147,11 +147,9 @@ void CUITalkDialogWnd::InitTalkDialogWnd()
     }
 }
 
-void CUITalkDialogWnd::Show(const bool notify_scripts /*= true*/)
+void CUITalkDialogWnd::Show()
 {
     InventoryUtilities::SendInfoToActor("ui_talk_show");
-    if (notify_scripts)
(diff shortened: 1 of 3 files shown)


Xottab-DUTY:
I changed the code to a better variant.
This scripts notification was introduced in Clear Sky, where talk dialog was completely closed when you switch to talk or upgrade, then they decided to not close it in COP. What we now do here is just hiding it, not closing, following both SOC and COP behaviours in a correct way.

Xottab-DUTY:
Thanks!

m4ttblue:
I appreciate everyone for fixing this.

PR #1993, opened 2025-11-14
Combine OpenGL uniforms into batches

crashuxx:


src/Layers/xrRenderGL/glDetailManager_VS.cpp
```diff
@@ -84,7 +84,7 @@ void CDetailManager::hw_Render(CBackend& cmd_list)
     //hw_Render_dump			(&*hwc_s_array,	0, 1, c_hdr );
     hw_Render_dump(cmd_list, consts, wave.div(PI_MUL_2), dir2, 0, 1);
 }
-
+#include <glm/glm.hpp>
```
Xottab-DUTY: Includes always should be at the beginning of the file.

src/Layers/xrRenderGL/glDetailManager_VS.cpp
```diff
@@ -148,6 +148,8 @@ void CDetailManager::hw_Render_dump(CBackend& cmd_list, const Fvector4& consts,
                 VERIFY(c_storage);*/
 
                 u32 dwBatch = 0;
+                std::vector<glm::vec4> uniformBuffer;
```
Xottab-DUTY: It should be xr_vector.
```suggestion
                xr_vector<glm::vec4> uniformBuffer;
```

src/Layers/xrRenderGL/glDetailManager_VS.cpp
```diff
@@ -176,21 +180,22 @@ void CDetailManager::hw_Render_dump(CBackend& cmd_list, const Fvector4& consts,
...
                             //RCache.get_ConstantCache_Vertex().b_dirty				=	TRUE;
                             //RCache.get_ConstantCache_Vertex().get_array_f().dirty	(c_base,c_base+dwBatch*4);
+
+                            R_constant_load L = constArray->vs;
+                            if (constArray->destination & RC_dest_pixel) { L = constArray->ps; }
+                            if (constArray->destination & RC_dest_vertex) { L = constArray->vs; }
+                            if (constArray->destination & RC_dest_geometry) { L = constArray->gs; }
+                            if (constArray->destination & RC_dest_all) { L = constArray->pp; }
+
+                            CHK_GL(glProgramUniform4fv(L.program, L.location, uniformBuffer.size(), reinterpret_cast<GLfloat*>(uniformBuffer.data())));
```
Xottab-DUTY: You should add new functionality to the backend instead of bypassing it, i.e. you need to add new function to the CBackend class and use `cmd_list`.

src/Layers/xrRenderGL/glDetailManager_VS.cpp
```diff
@@ -202,6 +207,15 @@ void CDetailManager::hw_Render_dump(CBackend& cmd_list, const Fvector4& consts,
                     u32 dwCNT_prims = dwBatch * Object.number_indices / 3;
                     //RCache.get_ConstantCache_Vertex().b_dirty				=	TRUE;
                     //RCache.get_ConstantCache_Vertex().get_array_f().dirty	(c_base,c_base+dwBatch*4);
+                    R_constant_load L = constArray->vs;
+                    if (constArray->destination & RC_dest_pixel) { L = constArray->ps; }
+                    if (constArray->destination & RC_dest_vertex) { L = constArray->vs; }
+                    if (constArray->destination & RC_dest_geometry) { L = constArray->gs; }
+                    if (constArray->destination & RC_dest_all) { L = constArray->pp; }
```
Xottab-DUTY: 1. We don't allow `if` one-liners. The body should always be on the new line.
2. And, in this case, parenthesis are not needed.
```suggestion
                    if (constArray->destination & RC_dest_pixel)
                        L = constArray->ps;
                    if (constArray->destination & RC_dest_vertex)
                        L = constArray->vs;
                    if (constArray->destination & RC_dest_geometry)
                        L = constArray->gs;
                    if (constArray->destination & RC_dest_all)
                        L = constArray->pp;
```

Xottab-DUTY (changes requested):
What this change does?

src/Layers/xrRenderGL/glDetailManager_VS.cpp
crashuxx: as a newcomer/noob  I have no clue what is xr_vector, but i need a vector

Xottab-DUTY: Just replace std::vector with xr_vector. They are the same, except that the latter uses custom allocator.

src/Layers/xrRenderGL/glDetailManager_VS.cpp
crashuxx: yes, it's a draft PR :)  but big that for comment, don't hesitate

crashuxx:
reduces gl... calls, the same is done on DX side but in a different way. I want to be more declarative here so in the end I will(I want to) use UBO not Uniiforms

crashuxx:
it is reducing frame time in the same place/spot from 5.0-5.3 ms to 4.4-4.6ms

res/gamedata/shaders/gl/deffer_detail_s_flat.vs
```diff
@@ -1,18 +1,24 @@
...
+layout(std140) uniform EnvironmentDetailUBO {
+    mat4 xform;
+    mat4 formView;
+    vec4 consts;
+    vec4 scale;
+    vec4 bias;
+    vec4 wind;
+    vec4 wave;
+    vec3 sun;
+};
```
Xottab-DUTY: Code style:
1. `{` is always on the new line.
2. We use HLSL types

```suggestion
layout(std140) uniform EnvironmentDetailUBO
{
    float4x4 xform;
    float4x4 formView;
    float4 consts;
    float4 scale;
    float4 bias;
    float4 wind;
    float4 wave;
    float4 sun;
};
```

crashuxx:
I closed this as the number of changes is huge as Im changing how the whole renderer is working , i gonna publish it as a new ogl renderer

PR #2019, opened 2025-12-24
Remove shared_str::operator* (#1948)

### LINKED ISSUE #1948: Refactor shared_str
    1. Get rid of `operator*`
    - O(1) – search
5. Code organization:

References, useful links:

Aardes-Roiven:


Xottab-DUTY:
Thanks, very much!!!

PR #2021, opened 2025-12-27
shared_str: add bool operator and prohibit direct comparison with nullptr (#1948)

### LINKED ISSUE #1948: Refactor shared_str

Aardes-Roiven:


src/xrGame/Actor.cpp
```diff
@@ -1500,7 +1500,7 @@ void CActor::shedule_Update(u32 DT)
                 }
                 else if (m_pVehicleWeLookingAt)
                 {
-                    m_sDefaultObjAction = m_pVehicleWeLookingAt->m_sUseAction != nullptr ? m_pVehicleWeLookingAt->m_sUseAction : m_sCarCharacterUseAction;
+                    m_sDefaultObjAction = !m_pVehicleWeLookingAt->m_sUseAction ? m_sCarCharacterUseAction : m_pVehicleWeLookingAt->m_sUseAction;
```
Xottab-DUTY: We also need an `explicit operator bool` in `shared_str` and this is a good moment, when we can add it.

Can you add `explicit operator bool` to shared_str in this PR so that we don't need to reverse this condition?

Aardes-Roiven: sure

Xottab-DUTY (approved):

Xottab-DUTY:
Thanks!

PR #2022, opened 2025-12-28
Replace shared_str::printf with standardized xr_sprintf (#1948)

### LINKED ISSUE #1948: Refactor shared_str

Aardes-Roiven:


src/xrGame/ui/ServerList.cpp
```diff
@@ -868,8 +868,10 @@ void CServerList::SrvInfo2LstSrvInfo(const ServerInfo* pServerInfo)
     m_itemInfo.info.address = address.c_str();
     m_itemInfo.info.map = pServerInfo->m_SessionName;
     m_itemInfo.info.game = GameTypeToStringEx(pServerInfo->m_GameType, true);
-    m_itemInfo.info.players.printf("%d/%d", pServerInfo->m_ServerNumPlayers, pServerInfo->m_ServerMaxPlayers);
-    m_itemInfo.info.ping.printf("%d", pServerInfo->m_Ping);
+
+    xr_sprintf(m_itemInfo.info.players, "%d/%d", pServerInfo->m_ServerNumPlayers, pServerInfo->m_ServerMaxPlayers);
+    xr_sprintf(m_itemInfo.info.ping, "%d", pServerInfo->m_Ping);
+
```
Xottab-DUTY:
```suggestion
    xr_sprintf(m_itemInfo.info.players, "%d/%d", pServerInfo->m_ServerNumPlayers, pServerInfo->m_ServerMaxPlayers);
    xr_sprintf(m_itemInfo.info.ping, "%d", pServerInfo->m_Ping);
```

src/Layers/xrRender/ETextureParams.cpp
```diff
@@ -161,8 +161,15 @@ void STextureParams::FillProp(LPCSTR base_name, PropItemVec& items, PropValue::T
...
-    PHelper().CreateCaption(items, "Source" DELIMITER "Width", shared_str().printf("%d", width));
-    PHelper().CreateCaption(items, "Source" DELIMITER "Height", shared_str().printf("%d", height));
+
+    shared_str width_str;
+    xr_sprintf(width_str, "%d", width);
+    PHelper().CreateCaption(items, "Source" DELIMITER "Width", width_str);
+
+    shared_str height_str;
+    xr_sprintf(height_str, "%d", height);
+    PHelper().CreateCaption(items, "Source" DELIMITER "Height", height_str);
```
Xottab-DUTY:
```suggestion
    shared_str temp_str;
    xr_sprintf(temp_str, "%d", width);
    PHelper().CreateCaption(items, "Source" DELIMITER "Width", temp_str);

    xr_sprintf(temp_str, "%d", height);
    PHelper().CreateCaption(items, "Source" DELIMITER "Height", v);
```

src/xrCore/xrstring.h
```diff
@@ -176,21 +176,20 @@ class shared_str
...
+    string4096 buf;
+    va_list p;
+    va_start(p, format_string);
+    int vs_sz = vsnprintf(buf, sizeof(buf) - 1, format_string, p);
+    buf[sizeof(buf) - 1] = 0;
+    va_end(p);
+    if (vs_sz >= 0)
+        destination = buf;
+    return vs_sz;
+}
```
Xottab-DUTY:
```suggestion
inline int __cdecl xr_sprintf(shared_str& destination, pcstr format_string, ...)
{
    string4096 buf;
    va_list args;
    va_start(args, format_string);
    const int vs_sz = vsnprintf(buf, sizeof(buf) - 1, format_string, args);
    buf[sizeof(buf) - 1] = 0;
    va_end(p);
    if (vs_sz >= 0)
        destination = buf;
    return vs_sz;
}
```

Xottab-DUTY (approved):

src/xrCore/xrstring.h
```diff
@@ -176,21 +176,20 @@ class shared_str
...
 };
 
+inline int __cdecl xr_sprintf(shared_str& destination, pcstr format_string, ...)
+{
+    string4096 buf;
+    va_list args;
+    va_start(args, format_string);
+    const int vs_sz = vsnprintf(buf, sizeof(buf) - 1, format_string, args);
+    buf[sizeof(buf) - 1] = 0;
+    va_end(p);
```
Xottab-DUTY:
```suggestion
    va_end(args);
```

Xottab-DUTY:
Thanks!

PR #2027, opened 2026-01-01
Use elfutils to generate stack traces on linux

Igigog:
This changes the way engine stack traces are captured on Linux. Here's the comparison (commit f70e397673627e3c6294a7b329179fc5cac561c1, build type: Debug):

Before:
image

After:
Screenshot From 2026-01-01 15-35-15

A few important things:
- These color codes are printed to the log file as a mess of escape characters, though you can filter them out manually afterwards. I think that's fine, given that they help a lot and it's only the case on Linux.
- We'll need to update Wiki page to add new dependency. It's `elfutils-dev` on Alpine, `libdw-dev` on Ubuntu and `elfutils-devel` on Fedora (when in doubt, check CI file)

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -110,7 +110,7 @@ jobs:
       with:
         arch: ${{ matrix.platform.arch }}
         branch: 'latest-stable'
-        packages: build-base cmake git mold sdl2-dev lzo-dev libjpeg-turbo-dev openal-soft-dev libogg-dev libtheora-dev libvorbis-dev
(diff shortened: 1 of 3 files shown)


Xottab-DUTY:
I think introducing new low-level dependency is suboptimal.
Could you make it optional and not include it into CI builds?

Talking about desired future changes, it's waay better to use https://github.com/jeremy-rifkin/libassert. Under the hood it uses https://github.com/jeremy-rifkin/cpptrace which supports generating stack traces using several backends, so we don't need to support the code ourselves.
Could you research if we can use cpptrace directly?
I also can't find any elfutils usage in cpp trace, so maybe we can send a PR with new backend support (I'm not sure if it's needed).

PR #2041, opened 2026-01-31
Mark rsqrt function as noexcept

AMS21:
Apparently `clang-21` will not compile unless we mark `rsqrt` as `noexcept`.

Since it just calls `_sqrt` which is already marked `noexcept` this should be fine.

### DIFF
--- a/src/utils/xrMiscMath/vector.cpp
+++ b/src/utils/xrMiscMath/vector.cpp
@@ -140,7 +140,7 @@ float angle_inertion_var(float src, float tgt, float min_speed, float max_speed,
 	return src;
 }
 
-double rsqrt(double v) { return 1.0 / _sqrt(v); }
+double rsqrt(double v) noexcept { return 1.0 / _sqrt(v); }
 
 //////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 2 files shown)


Xottab-DUTY:
Is it ever used? I see that it returns double, which aren't used almost always.

AMS21:
> Is it ever used? I see that it returns double, which aren't used almost always.

You're right it's actually unused. I've updated the PR to instead simply remove the `rsqrt` function.

AMS21:
Wait it's not unused. It used right [here](https://github.com/OpenXRay/xray-16/blob/138df3e8c755552efffc4ffb41fee2dbdda97fd4/src/utils/xrMiscMath/xrMiscMath.cpp#L11). No idea how I was able to compile with the `rsqrt` removed :/

Xottab-DUTY:
> Wait it's not unused. It used right [here](https://github.com/OpenXRay/xray-16/blob/138df3e8c755552efffc4ffb41fee2dbdda97fd4/src/utils/xrMiscMath/xrMiscMath.cpp#L11). No idea how I was able to compile with the `rsqrt` removed :/

Successfully compiled on Fedora somehow: https://github.com/OpenXRay/xray-16/actions/runs/21565121323/job/62135268722

Xottab-DUTY:
Thanks!

PR #2043, opened 2026-02-06
Deleted eax.

1therealcloud:
I don’t know why it’s here if it isn’t linked anywhere and is only used by `eax.h`.


Xottab-DUTY:
It's used in xrSound, we still support EAX.

1therealcloud:
image

1therealcloud:
Yes, it’s used in `xrSound`, but it isn’t linked.

Xottab-DUTY:
Well, agree. The `eax.lib` isn't being linked directly. But `eax.dll` is still loaded during launch. Please, remove `eax.lib` only, but don't touch `eax.dll`.

1therealcloud:
For example, the game works fine for me without `eax.dll`. Is it really required?

Xottab-DUTY:
It works, because it's optional, but we want to provide this dll.

1therealcloud:
Alright, I’ve pushed a commit.

Xottab-DUTY:
Thanks!

PR #2044, opened 2026-02-06
Utils cleanup

1therealcloud:
mp_balancer — Not present in the solution
mp_gpprof_server — Not present in the solution
NvMender2003 — Required for editors and compilers (utilities)
xrLoadSurface.cpp — Required for editors; it uses the FreeImage library, which is not included in the engine

I’m also not sure whether these utilities (mp_*) work at all. Overall, I made this PR because they don’t do anything — they’re just sitting there. In any case, they can always be brought back later.

Also, Common.vcxproj now doesn’t contain any .cpp files. So, it seems like it’s not needed?

Also removed the old Borland projects `(.dsk, .bpr, .bpf)`.


Xottab-DUTY (approved):
@yohjimane, what do you think?

I didn't removed any multiplayer tools because we still support multiplayer and I didn't qualified/investigated if we need them.
As for other files related to compilers and editors, there was a plan to return them or recreate them, so I left these files as is until we decide what to do with the compilers.
I think that it's good to remove these, because I just was cautious and didn't had time to clean it up.

Common.vcxproj is used as a visual representation for files in the Common folder. Otherwise, these files will be shown in "External Dependencies" list which is so long that it is inconvenient to look at it. So, it is needed until we finish the transition to CMake.

Xottab-DUTY:
Ah, please fix the style-check workflow script, it tries to access removed `mp_*` folders and fails

1therealcloud:
Fixed it. Probably.

1therealcloud:
And what about `mp_configs_verifier`?

Xottab-DUTY:
You just removed the spacing check, no it's incorrect. The check should be preserved, you just need to remove missing paths from the arguments.

Xottab-DUTY:
> And what about `mp_configs_verifier`?

It's an anticheater tools for server admins. It should be working I think, but we can fix it later if it's broken.
But yeah, it's needed.

1therealcloud:
Okay, I see. Now looks like I fixed it.

Xottab-DUTY:
Still didn't fixed it. There are more missing paths in file :)

1therealcloud:
I see. This time it seems like I've done everything.

Xottab-DUTY:
Please, remove the language change and open a new PR for it! Language change is not related to utils cleanup. Keep logically different things in different PRs.

1therealcloud:
I'm just too lazy to make a new one.

1therealcloud:
The file face_smoth_flags.h is also needed by editors and is not used anywhere.

PR #2065, opened 2026-03-06
fix: populating npc spawn data asserting when npc has no class set

AMS21:
If the npc has no class set we fail with a fatal error when reading the clsid.

Note that this happened when using mods not on the vanilla games.


Xottab-DUTY:
Thanks as always!

PR #2068, opened 2026-03-09
fix: variable name for current Git commit SHA

AMS21:
I've noticed that the log message when starting never includes the actual git commit SHA.

```
OpenXRay Debug build 9892, Mar  9 2026 (Linux 64-bit, static)
Custom build from commit[] branch[dev]
```

This is because the actual variable defined [here](https://github.com/OpenXRay/xray-16/blob/ef47a17b7d444c92fb216b7565c6d4f47e6dbca2/cmake/XRay.Build.cmake#L53) is missing the trailing 1.


Xottab-DUTY:
Might be an accidental typo during refactoring :)

PR #2074, opened 2026-03-16
A few minor fixes

ctrl256:
Ive done a few minor fixes while browsing the code: potential oob write duplicated flag and typo

### DIFF
--- a/src/xrCore/FileSystem.cpp
+++ b/src/xrCore/FileSystem.cpp
@@ -99,7 +99,8 @@ void MakeFilter(string1024& dest, pcstr info, pcstr ext)
     }
     xr_strcpy(dest, res.c_str());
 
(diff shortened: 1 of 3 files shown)


src/xrCore/FileSystem.cpp
```diff
@@ -99,7 +99,8 @@ void MakeFilter(string1024& dest, pcstr info, pcstr ext)
     }
     xr_strcpy(dest, res.c_str());
 
-    for (size_t i = 0; i < res.size(); ++i)
+    const auto bound = res.size() > sizeof(dest) ? sizeof(dest) : res.size();
```
Xottab-DUTY:
```suggestion
    const auto bound = std::min(res.size(), sizeof(dest));
```

PR #2076, opened 2026-03-20
ALife graph registry robustness (#2058)

ChugunovRoman:
ALife graph registry (alife_graph_registry.cpp)
CoC ends up calling teleport_object from Lua in situations where the NPC is no longer in the graph cell that matches m_tGraphID. The engine then hit CSafeMapIterator::remove with “object not in registry” and died. I made remove look for the id on all graph vertices, and if it’s nowhere, skip the graph remove instead of asserting. Same situation could leave you with a strict level().remove on an empty level registry, so when the graph remove didn’t actually remove anything, the level remove is done with no_assert.

On disconnect I also saw the opposite: add blowing up with “already in registry” when se_monster’s on_unregister (or whatever runs during teardown) tried to register again while the object was still listed. add is now safe in that case: if the same pointer is already in the target vertex we no-op; otherwise we strip that id from every vertex once and then insert.

Refs #2058 (same class of CoC + ALife weirdness).

Key bindings for Lua (key_binding_registrator_script.cpp + key_binding_dik_lua.inl)
AXR’s axr_keybind does key_bindings["DIK_…"] for values coming from config. In OpenXRay those scan codes only lived on DIK_keys, so Lua died with no static 'DIK_NUMPADENTER' in class 'key_bindings'. I mirrored the same value(...) list onto key_bindings via a shared include so we don’t maintain two divergent copies. Also fixed the stray ) in DIK_KBDILLUMTOGGLE’s string while at it.

### DIFF
--- a/src/xrGame/alife_graph_registry.cpp
+++ b/src/xrGame/alife_graph_registry.cpp
@@ -176,7 +176,27 @@ void CALifeGraphRegistry::add(CSE_ALifeDynamicObject* object, GameGraph::_GRAPH_
     if (!object->m_bOnline && object->used_ai_locations() /**&& object->interactive()**/)
     {
         VERIFY(ai().game_graph().valid_vertex_id(game_vertex_id));
-        m_objects[game_vertex_id].objects().add(object->ID, object);
+        OBJECT_REGISTRY& target = m_objects[game_vertex_id].objects();
+        const auto& target_map = target.objects();
+        const auto existing_at_target = target_map.find(object->ID);
+        const bool already_registered_here =
+            (existing_at_target != target_map.end() && existing_at_target->second == object);
+
+        if (!already_registered_here)
+        {
+            const GameGraph::_GRAPH_ID vertex_count = (GameGraph::_GRAPH_ID)m_objects.size();
+            for (GameGraph::_GRAPH_ID i = 0; i < vertex_count; ++i)
+            {
+                if (!ai().game_graph().valid_vertex_id(i))
+                    continue;
+                OBJECT_REGISTRY& reg = m_objects[i].objects();
+                const auto& om = reg.objects();
+                if (om.find(object->ID) == om.end())
+                    continue;
+                reg.remove(object->ID, true);
+            }
+            target.add(object->ID, object);
+        }
         object->m_tGraphID = game_vertex_id;
     }
     else if (!m_level && update)
@@ -191,6 +211,7 @@ void CALifeGraphRegistry::add(CSE_ALifeDynamicObject* object, GameGraph::_GRAPH_
 
 void CALifeGraphRegistry::remove(CSE_ALifeDynamicObject* object, GameGraph::_GRAPH_ID game_vertex_id, bool update)
 {
+    bool removed_from_graph = false;
     if (object->used_ai_locations() /**&& object->interactive()**/)
     {
 #ifdef DEBUG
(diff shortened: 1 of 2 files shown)


Xottab-DUTY (commented):
> ALife graph registry (alife_graph_registry.cpp)
CoC ends up calling teleport_object from Lua in situations where the NPC is no longer in the graph cell that matches m_tGraphID. The engine then hit CSafeMapIterator::remove with “object not in registry” and died. I made remove look for the id on all graph vertices, and if it’s nowhere, skip the graph remove instead of asserting. Same situation could leave you with a strict level().remove on an empty level registry, so when the graph remove didn’t actually remove anything, the level remove is done with no_assert.
> 
> On disconnect I also saw the opposite: add blowing up with “already in registry” when se_monster’s on_unregister (or whatever runs during teardown) tried to register again while the object was still listed. add is now safe in that case: if the same pointer is already in the target vertex we no-op; otherwise we strip that id from every vertex once and then insert.

Could you describe the problem in Russian? :)

> Key bindings for Lua (key_binding_registrator_script.cpp + key_binding_dik_lua.inl)
AXR’s axr_keybind does key_bindings["DIK_…"] for values coming from config. In OpenXRay those scan codes only lived on DIK_keys, so Lua died with no static 'DIK_NUMPADENTER' in class 'key_bindings'. I mirrored the same value(...) list onto key_bindings via a shared include so we don’t maintain two divergent copies. Also fixed the stray ) in DIK_KBDILLUMTOGGLE’s string while at it.

The problem with `DIK_` crashes in Lua is due to wrong luabind behaviour and it's bad idea to make workarounds for it in the game code, we should fix luabind directly. So, I don't approve the proposed workaround.

ChugunovRoman:
> > ALife graph registry (alife_graph_registry.cpp)
> > CoC ends up calling teleport_object from Lua in situations where the NPC is no longer in the graph cell that matches m_tGraphID. The engine then hit CSafeMapIterator::remove with “object not in registry” and died. I made remove look for the id on all graph vertices, and if it’s nowhere, skip the graph remove instead of asserting. Same situation could leave you with a strict level().remove on an empty level registry, so when the graph remove didn’t actually remove anything, the level remove is done with no_assert.
> > On disconnect I also saw the opposite: add blowing up with “already in registry” when se_monster’s on_unregister (or whatever runs during teardown) tried to register again while the object was still listed. add is now safe in that case: if the same pointer is already in the target vertex we no-op; otherwise we strip that id from every vertex once and then insert.
> 
> Could you describe the problem in Russian? :)

При Disconnect / выгрузке сначала уничтожается CALifeObjectRegistry, у объектов вызывается Lua on_unregister (se_monster.script). Скрипт (или цепочка из него) снова дергает путь, который делает graph().add() для уже зарегистрированного в этой вершине офлайн-объекта.

> > Key bindings for Lua (key_binding_registrator_script.cpp + key_binding_dik_lua.inl)
> > AXR’s axr_keybind does key_bindings["DIK_…"] for values coming from config. In OpenXRay those scan codes only lived on DIK_keys, so Lua died with no static 'DIK_NUMPADENTER' in class 'key_bindings'. I mirrored the same value(...) list onto key_bindings via a shared include so we don’t maintain two divergent copies. Also fixed the stray ) in DIK_KBDILLUMTOGGLE’s string while at it.
> 
> The problem with `DIK_` crashes in Lua is due to wrong luabind behaviour and it's bad idea to make workarounds for it in the game code, we should fix luabind directly. So, I don't approve the proposed workaround.

Давай тогда я убиру правки по DIK_  из этой ветки и может позже их сделаю. Тогда получаеться надо править сам luabind?

ChugunovRoman:
@Xottab-DUTY Убрал правки по DIK_

Xottab-DUTY:
> Давай тогда я убиру правки по DIK_ из этой ветки и может позже их сделаю. Тогда получаеться надо править сам luabind?

Прости, не заметил (уведомление только прочитал, оно обрезанное было).
На Release и Master Gold у luabind отключена проверка ошибок (через дефайн LUABIND_NO_ERROR_CHECKING) и он всё пропускает. В остальных конфигурациях проверка на ошибки включена и запрос несуществующего поля у enum, экспортированного из движка, считается ошибкой. Как я понимаю, в оригинальном луабинде такого поведения нет даже с включенной проверкой на ошибки.

PR #2087, opened 2026-04-15
Fix PDA task icons: enable icons for additional tasks and restore texture rendering

### LINKED ISSUE #2086: Problems with task icons in PDA: icons for secondary tasks are missing, and the main task icon is not displayed after switching tasks.
**Describe the bug**
There are two related issues with task icons in the PDA user interface:

1. Secondary (additional) tasks do not display icons in map tooltips.
2. Task icons may disappear after completing or switching tasks.

### Issue 1: Missing icons for secondary tasks

In the CUIMapLocationHint::SetInfoTask() function, secondary tasks explicitly hide the icon:

else if (task->GetTaskType() == eTaskTypeAdditional)
{
m_info["t_icon"]->Show(false); <---- Here!

}

###Issue 2: Icon not rendering after task switch

After completing a task or switching the active task:

texture path is valid
InitTexture() is called

However, the icon is not rendered because UIStatic remains in a disabled texture state (m_bTextureEnable = false).

**To Reproduce**
1. Open PDA
2. Hover a secondary task → no icon shown
3. Complete a task or wait for auto-switch
4. New active task appears without icon

**Expected behavior**
Both storyline and additional tasks should display icons
Icons should always render when texture is set.

**Actual behavior**
Additional task icons are hidden
Icons may not render after task switch
Root cause
UI logic explicitly hides icons for additional tasks
UIStatic::TextureOff() disables rendering and InitTexture() does not restore it

**Screenshots, videos**
Image

**Proposed solution**
Enable icon rendering for additional tasks and align layout with storyline tasks
Ensure texture rendering is re-enabled before assigning a new texture
This results in no icon being displayed for secondary tasks, even when a valid texture is provided.
Fix PDA task icons for additional tasks and after task switch

Image

 - OS: Windows 10 22H2 19045.6466
 - OpenXRay night build version 7368d4b0

yaroslavrib:
Fix PDA task icons for additional tasks and after task switch

Closes #2086 

### Summary
- Enable icons for additional tasks
- Fix UIStatic texture not rendering after TextureOff()

### Result
- Icons always render correctly
- Layout consistent

### DIFF
--- a/src/xrGame/ui/UITaskWnd.cpp
+++ b/src/xrGame/ui/UITaskWnd.cpp
@@ -441,6 +441,7 @@ void CUITaskItem::InitTask(CGameTask* task)
     {
         if (task)
         {
+            S->TextureOn();
             S->InitTexture(task->m_icon_texture_name.c_str());
             S->SetStretchTexture(true);
             m_info["t_icon_over"]->Show(true);
--- a/src/xrGame/ui/map_hint.cpp
+++ b/src/xrGame/ui/map_hint.cpp
@@ -173,20 +173,31 @@ void CUIMapLocationHint::SetInfoTask(CGameTask* task)
     }
     else if (task->GetTaskType() == eTaskTypeAdditional)
     {
-        m_info["t_icon"]->Show(false);
-        float w = m_info["t_hint_text"]->GetWidth();
+        // Иконка побочного задания
+        m_info["t_icon"]->Show(true);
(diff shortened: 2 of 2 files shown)


Xottab-DUTY:
I figured out we can fix it with a simpler and better approach.
Thanks for the clue on how to fix vanilla bug and for the effort overall — it helped!

PR #1980, opened 2025-11-07
fix: compressing invalid direction vectors

AMS21:
Similar to https://github.com/OpenXRay/xray-16/pull/1974 but apparently this time the direction to compress included NaN.

<details>
  <summary>UBSAN report</summary>
  
  ```
/mnt/data/dev/xray-16/src/xrCore/_compressed_normal.cpp:103:20: runtime error: left shift of negative value -2147483648
    #0 0x7f67e2874eaa in pvCompress(_vector3<float> const&) /mnt/data/dev/xray-16/src/xrCore/_compressed_normal.cpp:103
    #1 0x7f67e2846884 in NET_Packet::w_dir(_vector3<float> const&) /mnt/data/dev/xray-16/src/xrCore/NET_utils.cpp:37
    #2 0x7f67f3aed2f6 in SHit::Write_Packet_Cont(NET_Packet&) /mnt/data/dev/xray-16/src/xrGame/Hit.cpp:108
    #3 0x7f67f3aed74a in SHit::Write_Packet(NET_Packet&) /mnt/data/dev/xray-16/src/xrGame/Hit.cpp:133
    #4 0x7f67f3443b01 in CExplosive::ExplodeWaveProcessObject(collide::rq_results&, CPhysicsShellHolder*) /mnt/data/dev/xray-16/src/xrGame/Explosive.cpp:715
    #5 0x7f67f344a087 in CExplosive::ExplodeWaveProcess() /mnt/data/dev/xray-16/src/xrGame/Explosive.cpp:737
    #6 0x7f67f344a3e7 in CExplosive::UpdateCL() /mnt/data/dev/xray-16/src/xrGame/Explosive.cpp:503
    #7 0x7f67f3a1b898 in CGrenade::UpdateCL() /mnt/data/dev/xray-16/src/xrGame/Grenade.cpp:272
    #8 0x7f67e3a5fdd0 in CObjectList::SingleUpdate(IGameObject*) /mnt/data/dev/xray-16/src/xrEngine/xr_object_list.cpp:144
    #9 0x7f67e3a69e22 in CObjectList::Update(bool) /mnt/data/dev/xray-16/src/xrEngine/xr_object_list.cpp:287
    #10 0x7f67e39eb9c3 in IGame_Level::OnFrame() /mnt/data/dev/xray-16/src/xrEngine/IGame_Level.cpp:196
[...]

PR #2119, opened 2026-07-13
[FIX] Normalize explosive hit direction

Neloreck:
### Notes

- Cause: explosion code applied its upward-bias formula to an unnormalized explosion-to-target vector, despite assuming a unit vector. A target sufficiently below an RGD-5/F1 explosion could produce a negative square-root input, resulting in a `NaN` direction and a debug `pvCompress` assertion.
- The upward impulse component is now consistent across target distances. This can make distant targets receive a more noticeable upward push.

### Changes

- Normalizes the radial explosion-to-target direction before applying `up_throw_factor`.
- Normalizes the final hit direction before serializing `SHit`.
- Uses an upward direction fallback for invalid, coincident, or cancelling vectors.

### Additional context

- Reproduced when had a lot of grenades thrown near Skadovsk (height variance here allows to reproduce such cases).
- Tested proposed normalization fix in a same place with 50+- grenades thrown with ~10 squads.

### DIFF
--- a/src/xrGame/Explosive.cpp
+++ b/src/xrGame/Explosive.cpp
@@ -696,11 +696,18 @@ void CExplosive::ExplodeWaveProcessObject(collide::rq_results& storage, CPhysics
         Fvector l_dir;
         l_dir.sub(l_goPos, m_vExplodePos);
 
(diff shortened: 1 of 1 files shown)


AMS21:
Seems like a more proper fix to #1980 which just addresses the UB in the compress function

PR #2122, opened 2026-07-22
xrRender: apply gamma correction via static_globals cbuffer

L1SP:
Apply gamma/brightness/contrast correction in the tonemap pass by piggybacking on the existing `static_globals` cbuffer (`register b2`), avoiding Slang reflection issues with new PS cbuffers.


src/xrGame/CharacterPhysicsSupport.cpp
```diff
@@ -573,9 +573,8 @@ void CCharacterPhysicsSupport::in_UpdateCL()
 #endif
     update_animation_collision();
     m_character_shell_control.CalculateTimeDelta();
-    if (m_pPhysicsShell)
+    if (m_pPhysicsShell && m_pPhysicsShell->isFullActive())
```
Xottab-DUTY: This is surely not related to the renderer, it's better to split it into separate commit at least.

L1SP: True, I changed this file to make it launch without an assertion failure on debug build. I'll remove it and maybe will commit these changes to a separate PR.

PR #2126, opened 2026-07-27
Fix/inventory f key closes window

mikhailRo:


Xottab-DUTY:
> Fix: F key in inventory/container windows equips/unequips instead of closing
> 
> F is bound to kUSE (world interact) but the engine's default keybinding
> table also maps it as a secondary keyboard bind for kUI_ACCEPT. Inside
> the inventory grid, CUICellItem::OnKeyboardAction treats kUI_ACCEPT on a
> hovered cell as a double-click, so pressing F while hovering an item
> equipped/unequipped it instead of closing the screen.
> 
> CUIActorMenu::OnKeyboardAction now intercepts kUSE first and closes the
> window (same as Escape/I), before the event can reach the focused cell
> item.

Good catch!
This change was made for gamepads or keyboard only UI navigation.
The better fix would be disable equip/unequip when the mouse is active.
The mouse activation/deactivation itself needs improvements.

PR #2131, opened 2026-08-02
[FIX] Fix false-positive constant buffer unregister errors and noisy log.

Neloreck:
### Notes

- Since the parallel render contexts refactor, `dx11ConstantBuffer` is registered in exactly one  per-context registry, but its destructor attempted to unregister from all `R__NUM_CONTEXTS`  registries. Every destroyed buffer produced guaranteed misses in the remaining registries, flooding the log with thousands of `! ERROR: Failed to find compiled constant buffer` lines on level unload and renderer teardown — all false positives. With the owning context tracked, the error can only fire on genuine registry corruption.
- Removes 6,680 false positive logs from log file that are appended on game shutdown.

### Changes

- Adds `m_context_id` to `dx11ConstantBuffer`, stamped by `CResourceManager::_CreateConstantBuffer` at registration time.
- Replaces the destructor's all-contexts unregister loop with a single targeted `_DeleteConstantBuffer(m_context_id, this)` call, guarded by `RF_REGISTERED` and a `VERIFY` on  the context id.
- Extends the failed-unregister error message with the buffer name and context id; adds a `GetBufferName()` accessor.

### Links

Introduced as side effect with https://github.com/OpenXRay/xray-16/pull/1331

### DIFF
--- a/src/Layers/xrRenderDX11/dx11ConstantBuffer.cpp
+++ b/src/Layers/xrRenderDX11/dx11ConstantBuffer.cpp
@@ -7,10 +7,12 @@ namespace xray::render::RENDER_NAMESPACE
 {
 dx11ConstantBuffer::~dx11ConstantBuffer()
 {
-    for (int id = 0; id < R__NUM_CONTEXTS; ++id)
+    if (dwFlags & xr_resource_flagged::RF_REGISTERED)
     {
-        RImplementation.Resources->_DeleteConstantBuffer(id, this);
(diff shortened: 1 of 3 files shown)


Xottab-DUTY:
@Neloreck, hi! Can you redirect this to `opengl-max` branch?

Xottab-DUTY (approved):

Xottab-DUTY:
Thank you!

